PR #44, opened 2018-06-26
Improve the documentation

pvizeli:


thomasvs:
This pull request combined with https://github.com/home-assistant/hassos/pull/86 results in two wired network configurations on the documentation page, without a clear explanation on why they are different (they look the same apart from id and uuid) or what they're supposed to do.

PR #48, opened 2018-06-28
Fix hassos-conf label

pvizeli:
Fix: #49


cogneato:
The label `hassos-conf` will work for fat32 volumes

PR #149, opened 2018-08-07
Update config.txt

kwdavidson:
Enable SPI and I2C by default.


pvizeli:
That have no effect, we build this dynamic. Please explain your change request a bit more.

kwdavidson:
There are Hass-supported devices that use the SPI and I2C ports on a Raspberry Pi. Since these ports are disabled by default, it forces the user to figure out how to get into the boot partition and modify the config.txt file. If they were on by default, things would just work.

pvizeli:
Is that the default on rasbian?

buildroot-external is the layer where we can change things.

pvizeli:
Put it in post build script of rpi

PR #162, opened 2018-08-17
Adding Support for Odroid C2

rbray89:
Tested and mostly working.

The only thing that seems to be partially broken is USB, in that it will not initialize unless a device was plugged in at boot. Seems to be a Kernel issue. Aside from that, works well.


pvizeli:
Good work. Would be nice if we can finish that together. I'm fine to use MBR as backup if we can't use the first 1024 bytes for GPT like xu3/4.

First, can we use the linux kernel LT (4.14)? 4.17 lost the support on next linux kernel release and that will be soon. It is hard to maintain none LT kernel version, because we need also update stable release.

pvizeli:
If you want help on it, I can add you as member and we can create a branch and work on it until we can merge it to dev. That make it easier to working on it. (but we need make PRs to this branch).

I become access to repository and a developer OTA key, if you promise to help and maintain this cu2 because i have no hardware for tests.

pvizeli:
Wait with eMMC.

We should first make it compatible to HassOS and our env and as next we can make a MMC concept for all devices

pvizeli:
I backup your last commit

rbray89:
So the eMMC commit really only does two things:
1. Saves/loads env to the bootstate partition for the booted device (saveenv uses hard-coded device)
2. Switches the loaded partitions based on boot device (eMMC vs. SD)
All of this just happens in the u-boot script so I'm not sure there is much to create a universal concept for. 
I pushed it to https://github.com/rbray89/hassos/tree/odroid_emmc_test

pvizeli:
Yeah, that's fine. We need hold the focus on wide range of device support. Actual we should look that the device is inside upstream and after we are done with XU/C2 we could look forward 👍  It make not a lot sense to work on new feature before we are not done with base.

Do you have discord? We can make a short chat to split up the work.

rbray89:
Ah, I see. 
Yeah, you can find me @ /u/waka324

pvizeli:
I'm was not able to find ou :D can you send me a friend request at pvizeli#6293

don't understand me wrong. The eMMC is what we want have, but first we should fix the base.

rbray89:
Yeah, I agree.

PR #165, opened 2018-08-19
WIP: Odroid

pvizeli:
Please no one commit directly into this PR. Any change need to be PRs.

ToDo:
- [x] linux LT kernel (4.14)
- [x] Use file structure like raspberry for xu/c2
- [x] Use HassOS bootstate partition
- [ ] Use HassOS fileenv function for bootargs
- [x] Use U-Boot defconfig
- [ ] Create SPL for MBR like for GPT


rbray89:
For SPL on the odroid-C2, it requires that we write some data to block 0 (442 bytes), so the current mbr-spl.img won't work unless we overlay a non-gpt mbr-spl.img (containing the MBR partition entries) overtop. We also have to write to sector 1 as well, rather than 2. Given that some of these boards will require very different boot processes, perhaps it makes sense to have 'make_bootable' functions that contain the logic themselves similar to how I did it for the c2?

pvizeli:
I create the spl overlay for MBR very dynamic. Currently only odroid have this problem. So I create a hash that can fill into the function and they generate the image. But that is my problem :) would be nice if you port it to LT kernel.

rbray89:
Will do.

PR #176, opened 2018-08-26
Odroid-C2: updating kernel to 4.14.67

rbray89:


pvizeli:
Do you not need this patches for 4.14 anymore?

rbray89:
Yeah, the removed patches applied to other chipsets and/or already applied.

PR #188, opened 2018-09-02
Do not assume index 0 for boot medium

jtru:
When booting the HassOS VMDK appliance disk image on QEMU with OVMF UEFI
firmware, barebox numbers devices differently from other environments.

This patch makes the barebox `sh` implementation expand a pathname
wildcard at runtime, so that assigned disk indices != 0 will also work.

This fix introduces a potential (but most likely irrelevant)
bug/problem: If there is more than one disk connected to the system with
a matching name assigned, the wildcard will expand to two or more
pathnames, breaking the resulting `mount` command.


homeassistant:
Hi @jtru, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;hassos#188).

Thanks!

jtru:
This should fix #151 under most circumstances.

pvizeli:
Thanks

PR #199, opened 2018-09-18
Update documentation for a static IPV4 address via nmcli

borpin:
In looking for a way to set a static IPV4 address on a Esxi hosted Hassio, I discovered this method using the NetworkManager CLI (nmcli).  I felt it might be good to document it.

Happy for the formatting to be modified to suit local convention.

I cannot be more specific about how to get into the base OS as I don't know!

### DIFF
--- a/Documentation/network.md
+++ b/Documentation/network.md
@@ -97,6 +97,36 @@ If you have trouble with powersave you can do following:
 # Values are 0 (use default), 1 (ignore/don't touch), 2 (disable) or 3 (enable).
 powersave=0
 ```
+## Using nmcli to set a static IPV4 address
+
+this method has been tested with a Hass.io image on an Esxi server. Log into the HASSOS base system via a console – note this is not the same as an SSH login via the add-on.
+```
+Welcome to HassOS
+Hassio login:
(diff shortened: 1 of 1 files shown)


homeassistant:
Hi @borpin, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;hassos#199).

Thanks!

borpin:
Done.  I could see I needed to from the checks, but there was no link to that location :smile:

pvizeli:
Thanks

mattnewham:
I think this should also be updated to reference setting DNS and gateway, as with the current nmcli config you will only get local LAN access with no DNS from the docker containers etc

klrock:
I am trying to change the ipv4 ip address of my supervised Home Assistant installed on a win 10 VirtualBox.  Using cli in the console, when I enter "`# nmcli con edit “HassOS default”` " I get not recognized and am therefore unable to edit or change any settings.  Is this method only for Hassio?  Is there an alternate method for my installation?

PR #209, opened 2018-10-08
Implement basic ACPI shutdown support via busybox acpid

jtru:
On systems where ACPI support is present as inidcated by the presence of
/proc/acpi (e.g. on OVA compatible hypervisors), we want to properly
shut down the system when the power button is pressed (or the hypervisor
simulates this kind of event to the guest machine that executes hassos).

This changeset provides the following basic infrastructure for this
feature to work as expected:

 * a systemd service to start acpid, if ACPI support can be assumed
 * an acpid configuration directory
 * a trivial shutdown script to invoke when a PWR event is registered

Tested locally with the OVA image/target under qemu. I've taken care not to intefere with platforms without ACPI support, but I did not test those.


pvizeli:
Thanks

Issue #182, opened 2018-08-27
Configuration without a USB Stick


urish:
I installed Hass.io OS on my Pi Zero W, and didn't have an extra USB stick, so I found the following workaround for configuring the network through my windows machine:

1. Rename the boot partition from `hassos-boot` to `CONFIG`
2. Put the network configuration file inside `network/my-network` as outlined in the [documentation](https://github.com/home-assistant/hassos/blob/dev/Documentation/network.md).

I believe this workaround can be useful for other users who want to set up WiFi or a static IP address but don't have an available USB stick. If you wish, I can send a PR for the documentation suggesting this workaround.

pvizeli:
I'm open for PR's they allow to fallback into boot partition folder.

urish:
What do you think about the following procedure?

1. If /boot/config exists, mount it over /mnt/config (using bind)
2. If we have a config partition, mount it over /mnt/config as overlay mount
3. Then load all configuration files from /mnt/config as we do today

thesoftwarejedi:
@urish Thank you!  After failing to get the Pi Zero to read several USB sticks, this method worked great.

pvizeli:
@urish 

1. Check if a USB stick exists -> process
2. Otherwise look if a config partition exists -> process -> remove config

borpin:
Can I add a couple of cents worth, I am trying to change an instance of Hassio running on a VM to a static IP.  I cannot find the boot partition. I have tried putting the configuration in a 'my-networks' file in `\etc\networks\` but each time it gets deleted and nothing has changed.  I realise this is a slightly different issue, but if something was being implemented this use case might be considered.

urish:
@pvizeli thank you for the feedback! What is the recommended procedure for testing my changes? Is there some kind of QEMU setup where I can check it without building and writing an image to the Pi every single time?

PR #212, opened 2018-10-11
OS: Config importer support now also boot partition

pvizeli:
fix: #182 

Device like RPi zero have no USB stick. It allow also to simple set a config with boot partition (Fat32) after write a the SD card and need no extra device.

With support config per boot partition and with a USB stick, we could support any case to setup a device unattended.

### DIFF
--- a/Documentation/configuration.md
+++ b/Documentation/configuration.md
@@ -3,7 +3,7 @@
 ## Automatic
 
 You can use an USB drive with HassOS to configure network options, SSH access to the host and to install updates.
-Format a USB stick with FAT32/EXT4/NTFS and name it `CONFIG`. Use the following directory structure within the USB drive:
+Format a USB stick with FAT32/EXT4/NTFS and name it `CONFIG`. Alternative you can create a `CONFIG` folder inside boot partition. Use the following directory structure within the USB drive:
 
 ```text
 network/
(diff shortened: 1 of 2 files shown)


Documentation/configuration.md
```diff
@@ -3,7 +3,7 @@
 ## Automatic
 
 You can use an USB drive with HassOS to configure network options, SSH access to the host and to install updates.
-Format a USB stick with FAT32/EXT4/NTFS and name it `CONFIG`. Use the following directory structure within the USB drive:
+Format a USB stick with FAT32/EXT4/NTFS and name it `CONFIG`. Alternative you can create a `CONFIG` folder inside boot partition. Use the following directory structure within the USB drive:
```
jihao: I am not sure what boot partition is. 

> 1. Put the SD card in your SD card reader.
> 2. Open Etcher, select the Hass.io image and flash it to the SD card.
> 3. WiFi and Static IP setup only: Format a USB-Stick with name CONFIG, create a folder named network and within that folder a file named my-network. Copy one of the examples to the my-network file.

For step3, is it correct to create a folder in my SD card "/boot/CONFIG" before I install hassos to the Pi?

pabelanger: You don't seem to indicate which partition the is the boot, a default install only has:
- disk
- hassos-data
- hassos-kernel
- hassos-overlay

Are we to assume the full path is:
  /hassos-overlay/boot/CONFIG

Or another location? I've tried to read the code here, but don't fully understand who /mnt/boot/CONFIG will be created after the rasberry pi is full booted

mbo18: If I understand correctly, the 'CONFIG' folder should be create at the root of the disk partition.
It will then be mounted in /mnt/config/CONFIG
If someone try, please open a PR to clarify the documentation. Thank you.

PR #213, opened 2018-10-11
Enable vlan 8021q support

### LINKED ISSUE #211: 802.1q VLAN Support
Hi,

I recently hooked up HassOS with KVM on my hypervisor. Since I've different VLAN's that Hass should work with, I set up two new network manager profiles for these, using the single virtio link as parent.

However, it seems that the 8021q kernel module is missing. Would it be possible to include that in future builds? For what it's worth, that's something really useful.

Personally, I can just add another virtio link to the VM and couple network manager profiles to their MACs, but devices with only a single physical NIC don't have that option.

Best regards
Robert

jtru:
This fixes #211 and enables IEEE 802.1q VLAN support in the shared part of the kernel configuration.


jtru:
I'm afraid I fubar'd something with my fork's commit history (I'm not exactly familiar with the Github workflow, and that clearly shows). If this (rather trivial) patch can't be accepted because of this, please let me know - no hard feeling from my side ;)

pvizeli:
I'm fine with it. I need only first test that it is compatible with docker kernel options with https://github.com/moby/moby/blob/master/contrib/check-config.sh

I do that tomorrow

pvizeli:
In flavor https://github.com/home-assistant/hassos/pull/216

Thanks anyway

Issue #214, opened 2018-10-11
Roadmap Release REL-3


pvizeli:
I will create a new major release in 2018. Follow work need to be done for it:

Need:

- [x] Add support for config in boot partition (also into REL-1)
- [x] Update Buildroot to 2018.11
- [x] Change uboot script environment handling like odroid C2 (CC @rbray89)
- [x] Update docker to 18.09
- [ ] Split global kernel file into a second one with HassOS specific options
- [ ] Add odroid xu3/4
- [ ] Add Intel-nuc (need only a usb installer, all other will work like ova)

Maybe:

- [ ] Script that allow to export data partition to a external harddisk

PR #215, opened 2018-10-12
Uboot bootstate

rbray89:
Updating bootstate env support for uboot
Adding potential eMMC tinker board support
#214


buildroot-external/board/asus/tinker/uboot-boot.sh
```diff
@@ -1,3 +1,17 @@
+
+part start mmc ${devnum} 6 mmc_env
+mmc dev ${devnum}
+setenv loadbootstate "\
+echo 'loading env...';\
+mmc read ${ramdisk_addr_r} ${mmc_env} 0x40;\
+env import -c ${ramdisk_addr_r} 0x8000;"
```
pvizeli: for help to read it better, can we use follow style?
```
setenv loadbootstate " \
    echo ' '; \
    ....
```

rbray89:
Updated to address comments.

PR #216, opened 2018-10-12
Added 8021q kernel module for VLAN support

### LINKED ISSUE #211: 802.1q VLAN Support

efficiosoft:
Closes #211


buildroot-external/kernel/hassos-4_14.config
```diff
@@ -41,6 +41,10 @@ CONFIG_MACVLAN=y
 CONFIG_IPVLAN=y
 CONFIG_VXLAN=y
 
+CONFIG_VLAN_8021Q=m
```
pvizeli: Will that autoload my NetworkManager?

efficiosoft:
If some VLAN should be used or not has to be configured in the networkmanager profiles, as if you want to configure a WiFi connection. But yes, ``ip link add link eth0 type vlan id 123`` is enough to have the ``8021q`` module loaded automatically.

scottini:
Hello, how can I update my Hass OS installation to enable VLAN support? Then, how can I configure the functionality and, above all, make the config permanent?

efficiosoft:
@scottini Either build it yourself with the changes I made (not recommended if you are completely unfamiliar with those sort of things) or wait until it gets released.

Anyway, you'll find how to create VLAN connections in the NetworkManager docs. After all, it's just plain NM connection profiles. The installation tutorial mentions how to provide external connection profiles to hassio.

PR #220, opened 2018-10-16
Add informal "getting started with development" document

jtru:
I've written some English prose that I imagine could help others get started contributing to hassos. If you agree, I'd very much like to contribute it to the project's documentation.


pvizeli:
Thanks

PR #232, opened 2018-10-29
Fixed keeping name of udev rules after usb import

### LINKED ISSUE #231: udev rule import from usb stick fails
It copies its udev files to `*`. I guess this is not the expected behavior:

```
~ # ls /etc/udev/rules.d/
*
~ #
```

sholzmayer:
Closes #231


pvizeli:
Make a bugfix release this week 👍

PR #244, opened 2018-11-11
OdroidC2: Kernel Upgrade, USB update

rbray89:
- Upgraded kernel to 4.18.18 to fix some IRQ flodding issues.
 - Added additional USB drivers for APCUPS support.


rbray89:
I know it isn't a longterm support kernel version, but I wasn't able to find the patch that fixes the IRQ issue. Without this change, once HassOS kicks off HASS, the USB interrupts go crazy (assuming due to the z-wave dongle I have) and the softIRQ process uses 1/4 of the processing power.

pvizeli:
Ping, you can start the work :)

PR #267, opened 2018-11-28
Update timesyncd.conf

magnusoverli:
Set the default NTP servers to ntp.pool.org and Google NTP as fallback, avoiding defaulting to a less open and transparent service.


pvizeli:
We don't do that. Because of the anycast technology, the device receives the best and shortest way to next time server. NTP Pool is an excellent project, and we use this as a fallback, but if you don't can give a region like `0.de.pool.ntp.org`, they have a bad load and long time to adjust the correct date time.

PR #272, opened 2018-12-01
Bootloader: Allow boot from a USB device

pvizeli:


pvizeli:
@rbray89 what do you think?

rbray89:
Hmmm... I think there is much more work to get USB boot working... I think RPI-3 and OVA would be the only boards that could support it. Odroids need to load the boot code directly from flash, there are no USB drivers that can do this. We *might* be able to have a bootloader (uboot) on SD/eMMC that loads a USB driver, but I'm not sure how workable that would be.

For RPI3, there is a built-in USB storage driver that might work, but it requires some additional setup: https://www.raspberrypi.org/documentation/hardware/raspberrypi/bootmodes/msd.md

PR #299, opened 2018-12-30
OS: Make use of hardware rng devices

jvanderneutstulen:
Add rng-tools package to all boards. This will increase the entropy when a hardware rng device is present. This will speed up boot & setup quite a bit. (first time boot/setup on an ova image from 5 minutes down to 2 minutes). When no hardware rng device is present the rng-tools daemon will exit cleanly.


pvizeli:
Looks good, thanks for evaluation

Issue #289, opened 2018-12-23
Odroid XU4 Support


rbray89:
Documenting Odroid XU4 support progress here.

- [ ] uboot boots
- [ ] Kernel boots
- [ ] Kernel boots off eMMC
- [ ] Issues with using apparmor. Can be compiled in, but setting kernel security to apparmor causes kernel to not boot. Investigating.

rbray89:
Looks like there may be some tie-in to the big-little architecture hacks made for the xenyos processor. This might be addressed in 4.20.x where those changes were supposedly mainlined.

PR #303, opened 2019-01-04
Odroid-XU4: Initial commit

rbray89:
#289
There still seem to be issues with apparmor, such that it can't yet be enabled on the kernel command line, but can be compiled in. CGROUP_PIDS must also be disabled, but I'm not sure to the extent that this will impact security.


pvizeli:
You are my Hero!

Issue #290, opened 2018-12-26
Buildroot: Use git submodule instead of copying the distro in


deece:
This would allow us to more easily experiment with alternate versions of buildroot, see the revision history within buildroot and contribute patches upstream.

pvizeli:
`contribute patches upstream` it is, but Buildroot have a lot of work an a upstream patches take a year until it is in ;)

PR #304, opened 2019-01-04
Buildroot submodule: Bring in buildroot as a git submodule dependency

rbray89:
#290


jvanderneutstulen:
What about the local patches in `/buildroot-patches`? They are not applied now.
And the submodule should probably track a release branch (2018.11.x)

pvizeli:
Yes, we are on 2018.11.1, and we need the patches. I don't see an issue to have the buildroot folder inside... because the patches also only work for this version.

pvizeli:
I think the only possibility is that the new make script check out the needed version and apply the patches but I see no benefit to hold it inside the repository

Issue #250, opened 2018-11-17
find command not available


cogneato:
It looks like it is set here?
https://github.com/home-assistant/hassos/blob/3c21a8c486870c8da14e09fa331fdae7679df4eb/buildroot/package/busybox/busybox.config#L432

But it is not available when connected to the host.

jvanderneutstulen:
There are two issues here:

1. When you login via hassos-cli, it does not run ash as login shell, so `/etc/profile` will not be read and `/usr/sbin` is added to the $PATH.
2. The actual busybox config is set from `buildroot-external/busybox.config` and does not enable the find command.

pvizeli:
@jvanderneutstulen we should fix the path on the hassos-cli script before we map the host shell

PR #305, opened 2019-01-04
OS: Run ash as login shell, so $PATH is set correctly.

jvanderneutstulen:
Run ash as login shell will source /etc/profile where the `/sbin` and `/usr/sbin` are included in the path. Side effect is that `PS1` will be different from the current situation. Part of solution for #250.


pvizeli:
Cool, thanks

PR #306, opened 2019-01-04
BUILD: Build with make

jvanderneutstulen:
Use a Makefile for buiding, easier to build just a few targets and expandable in future (e.g. creating the defconfigs from fragments)


pvizeli:
Good work :+1:

PR #316, opened 2019-01-10
Addition of wireguard VPN

gregod:
Hi,
this adds the minimal VPN [Wireguard](https://www.wireguard.com/) to the HassOS Host.
Primary use cases are securely connecting to remote sensors and remote administration of the system.

This is implemented for the host instead of a docker container since wireguard lives as kernel module anyway and the userspace is a small wrapper around the busybox ip tools. Additionally stable & direct remote access to the host for debugging is worth the small image size and complexity increase.

There is [effort](https://lwn.net/ml/linux-kernel/20180731191102.2434-1-Jason@zx2c4.com/) to integrate wireguard into mainline linux and network manager is considering [native wireguard support](https://gitlab.freedesktop.org/NetworkManager/NetworkManager/commit/e3bd482329344b1d7f31b29fea2814ba5ce97d79). So ideally no customizations would be needed in the future.

This is tested on a raspberry 3+, but there shouldn't be any issues with other platforms.

Let me know what you think.


homeassistant:
Hi @gregod, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;hassos#316).

Thanks!

pvizeli:
Yeah, that looks fine. 👍 

Would be an option to use https://github.com/max-moser/network-manager-wireguard instead the manual configuration? I can also help to make a buildroot package for it.

gregod:
The network-manager-wireguard is a wrapper around the "wg-quick" script of wireguard, which has a dependency on bash. If pulling that in is something you are comfortable with i could have a look. I might need some help with creating the buildroot package though. One other option would be to use just wg-quick which would still require the seperate config file but allows much deeper configuration than the manual configuration and would keep all dependencies upstream.

gregod:
I started work on the network-manager-wireguard here: https://github.com/gregod/hassos/tree/feature-wireguardNetworkManager.  

So far it builds and i can connect the vpn via the nmcli. I can't get it to auto start on boot yet but i guess this is a solvable network manager config issue.

pvizeli:
@gregod this branch looks good, and I'm fine to merge this network manager plugin.

@frenk is working on an Add-on (Docker) they can create/manage VPN connections with wireguard over the UI as an add-on solution. They need the kernel module (userland only solution works too without the module) and you PR also add this module.

PR #321, opened 2019-01-16
Odroid-C2: Updating to 4.19 kernel

rbray89:


buildroot-external/configs/odroid_c2_defconfig
```diff
@@ -2,9 +2,9 @@ BR2_aarch64=y
 BR2_DL_DIR="/cache/dl"
 BR2_CCACHE=y
 BR2_CCACHE_DIR="/cache/cc"
-BR2_GLOBAL_PATCH_DIR="$(BR2_EXTERNAL_HASSOS_PATH)/patches $(BR2_EXTERNAL_HASSOS_PATH)/board/hardkernel/odroid-c2/patches"
+BR2_GLOBAL_PATCH_DIR="$(BR2_EXTERNAL_HASSOS_PATH)/patches"
 BR2_TOOLCHAIN_BUILDROOT_GLIBC=y
-BR2_KERNEL_HEADERS_4_18=y
+BR2_PACKAGE_HOST_LINUX_HEADERS_CUSTOM_4_19=y
```
jvanderneutstulen: Why not use `BR2_KERNEL_HEADERS_4_19=y`?

rbray89: For some reason, I started to see this error at build:
> package/linux-headers/linux-headers.mk:149: *** LINUX_HEADERS_SITE cannot be empty when LINUX_HEADERS_SOURCE is not.  Stop.
I assumed something changed in the buildroot update.

jvanderneutstulen: Ah, we need the next buildroot release to fix that. 

But can you remove the `*_HEADERS_*` configs, buildroot should then default to same version as the kernel.

PR #382, opened 2019-03-20
OS: Enable systemd-coredump

pvizeli:


sholzmayer:
Is there a reason why you added systemd-coredump?

PR #384, opened 2019-03-24
Add remark for linuxers to find partition

gluap:
For me it was confusing to find the instrurction "on the SD" without specifying which  partition of the SD. Both on MacOS with fuse drivers and Linux as well as on Windows with proper drivers multiple partitions are visible on the SD. So I think it makes sense to specify which partition to put the CONFIG folder below.

### DIFF
--- a/Documentation/network.md
+++ b/Documentation/network.md
@@ -4,7 +4,7 @@ HassOS uses NetworkManager to control the host network. In future releases, you 
 
 ## Configuration Examples
 
-You can read the [Official Manual][keyfile] or find many configuration examples across the internet. The system is read-only, if you don't want the IP address to change on every boot, you should set the UUID property with a generic [UUID4][uuid]. Inside `\CONFIG\network\` on the USB or SD, create a file called `my-network` and add the appropriate contents below:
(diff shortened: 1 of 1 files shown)


homeassistant:
Hi @gluap, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;hassos#384).

Thanks!

pvizeli:
That is technically not correct. But thanks anyway for your contribution. Maybe you will provide a full example? You can also create a PR to Home Assistant website. This documentation is more for integrators.

gluap:
@pvizeli How is this not technically correct? I took the time to create a new hassio image from scratch and on the resulting SD there are indeed multiple partitions. To successfully write the configuration I have to use the vfat partition of these. For windows and mac users whether the partitions show in explorer/finder depends on whether they have the proper fuse drivers installed. Most linux distros will auto-mount multiple partitions (but some will not auto-mount the vfat-partition in question).

Example, using debian testing:
- create hassio sd from reference image using ``dd if=<image file> of=/dev/sdX`` (substituting your SD device node for ``/dev/sdX``
- mount the ``vfat`` partition (first partition of the written image) (mount ``/dev/sdX1 /mnt``)
- add network config in said partition as ``/mnt/CONFIG/network``
- boot hassio
- network config is used 

Advantage of using the vfat partition of the disk hassio is installed on: Can configure WIFI from the start and without needing to have an USB-stick or extra sd-card at hand.

PR #397, opened 2019-05-01
Adding /etc/modprobe.d folder for module configuration

maelk:
This enables users to create specific module configurations in the /etc/modprobe.d folder, similarly to /etc/modules-load.d
The folder is mounted and hence retains changes. It can be populated using a config USB stick or the CONFIG folder on the boot partition.


pvizeli:
That is tested?

maelk:
Hi,
I have tested it locally, but did not find any testing guidelines in the repo, or any automated testing for that matter. If you have any, please let me know.
BR,

PR #400, opened 2019-05-10
ID parameter in file have to be the same as name of file

piowit:
id parameter in file have to be the same as name of file


homeassistant:
Hi @piowit, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;hassos#400).

Thanks!

PR #408, opened 2019-05-15
Add cifs support

Fantu:
This make possible mount remote smb share on hassos and with cifs support in
kernel is also possible have addon that can mount cifs (if they have
cifs-utils).


homeassistant:
Hi @Fantu, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;hassos#408).

Thanks!

buildroot-external/kernel/hassos.config
```diff
@@ -34,3 +34,11 @@ CONFIG_TUN=y
 CONFIG_VLAN_8021Q=m
 CONFIG_VLAN_8021Q_GVRP=y
 CONFIG_VLAN_8021Q_MVRP=y
+
+CONFIG_CIFS=y
+CONFIG_CIFS_STATS2=y
+CONFIG_CIFS_XATTR=y
+CONFIG_CIFS_POSIX=y
+CONFIG_CIFS_ACL=y
+CONFIG_CIFS_DEBUG=y
```
pvizeli: We don't need debug support

buildroot-external/kernel/hassos.config
```diff
@@ -34,3 +34,11 @@ CONFIG_TUN=y
 CONFIG_VLAN_8021Q=m
 CONFIG_VLAN_8021Q_GVRP=y
 CONFIG_VLAN_8021Q_MVRP=y
+
+CONFIG_CIFS=y
```
pvizeli: We should handle that as module, modules can also be loaded from add-ons

buildroot-external/configs/tinker_defconfig
```diff
@@ -106,3 +106,4 @@ BR2_PACKAGE_HASSOS_CLI_PROFILE="docker-default"
 BR2_PACKAGE_HASSOS_APPARMOR_DIR="supervisor/apparmor"
 BR2_PACKAGE_APPARMOR=y
 BR2_PACKAGE_BLUETOOTH_RTL8723=y
+BR2_PACKAGE_CIFS_UTILS=y
```
pvizeli: Like you say, the add-on needs cifs-utils. The Host is an embedded Linux and any functionality comes over docker/add-ons. On my point of view, we don't need that on host.

pvizeli: Please explain the use case on a working example and the needed size on rootfs if you want it on host.

Fantu: Thanks for reply, I changed the commit following your advice, this will make possible mount cifs from addon that support it.
About cifs I open a topic here:
https://community.home-assistant.io/t/mount-remote-smb-share-on-hassio/116734
For now I use it with custom hassos build but for testing but for production use on customers official support will be better.

pvizeli: Well, I'm open to support /mnt/data on a cifs network share. The question is, how do you want to manage that on an embedded system without hack?

Fantu: For now we'll use them for connect to nas and use it with video surveillance and media center but we have to fork the addons that need to use it adding cifs-utils (and mount -t cifs...) for now until I'll not found a solution or workaround to make a cifs addon working sharing mounted smb share on all other containers.
As I wrote here (https://community.home-assistant.io/t/mount-remote-smb-share-on-hassio/116734) I not found a way to set bind-propagation to shared on addon with config.json (is not possible actually or I'm wrong?) and I not found other solution without using hack that will be broken on reboot or hassos update. Any advice for solve this without fork any addon will need to use smb share is appreciated.
About testing cifs for other possible use I'll probably not have time in these days because I'm busy in other project more urgent to be ready for production.
Thanks for any reply and sorry for my bad english.

pvizeli: You can create an add-on they can load modules. You can create a folder on the share that is mounted by an add-on to the remote. Maybe NFS could be faster at that point.

The other way is offloading /mnt/data to external storage instead of the USB drive. The question is still how we want to manage that on OS to be persistent.

Fantu: Without cifs-utils in hassos (removed in this PR) will remain only possibility to use it in addon (with cifs-utils) and about what will use is ok because with addon can be done, started at boot, changed parameters easy from hassio gui adding only with few line of code in run.sh, and included also in backup are things possible, tested and ok.
The annoyance/trouble is that I not found a way (supported by hassio addon "out of the box" for what I found) to do mount smb share in 1 dedicated addon and share it in any other (instead fork any addon that need use smb share, that waste big amount of time on long term); the only thing I found was setting of "bind-propagation" setting (from default "rprivate" to "shared") of the addon's containers bind where will be mounted the smb share in one subfolder because map[] option of addon's config.json don't support it, or at least seems to me (I not checked the source code about addon's containers creation).
Can you please tell me if possible do bind-propagation changing in a permanent way (in the addon) or if they is another way to share things mounted inside an addon with other containers? it's probably obvious for you but I not have good experience with hassio/hassos and docker, other colleagues are working on the project using hassos and they only asked me to solve this important lack that they were unable to solve with existing addon and hassio settings.
Any advice is appreciated, thanks for any reply and sorry for my bad english.

PR #425, opened 2019-07-12
Odroid xu4 updates

rbray89:


buildroot-external/board/hardkernel/odroid-xu4/meta
```diff
@@ -4,4 +4,4 @@ CHASSIS=embedded
 BOOTLOADER=uboot
 KERNEL_FILE=zImage
 BOOT_SYS=mbr
-BOOT_ENV_SIZE=0x2000
+BOOT_ENV_SIZE=0x4000
```
pvizeli: Do that not break OTA updates from 0x2000 to 0x4000 ?

rbray89: This fixes the ability to read the env from Linux for rauc. Turns out the CRC is calculated using the actual env size for the export function. You are correct though that if a user is running on boot A, if they try to perform an upgrade, B may not boot. I think to fix this, we can try to import both 0x2000 and 0x4000. The CRC will fail on one, but the other should succeed and allow for backwards compatibility.

PR #430, opened 2019-07-15
Allow attaching external drive as a backup directory

darashenka:
if you label a partition "BACKUP", it will be mounted as a backup folder for hassio


pvizeli:
You need to define a full workflow, started from OS over to Supervisor/dbus that they cover a lot of user cases and can be used over the UI.

Please open an issue https://github.com/home-assistant/architecture and describe how a user-friendly backup system should look like on Hass.io | If they will be approved, you can start to implement your concept.

Thanks for your contribution.

PR #440, opened 2019-07-22
rpi: Add USB boot support

rbray89:
Tested/verified on RPI3 B1.2 with Lexar USB drive.


pvizeli:
Hmm, I'm don't like this.
The way to go is a script (can be also dbus driven) they allow moving the data partition to an external device. That will also work for all other boards.

Anyway, the uboot script will be not mess up with this and so I'm fine to merge this. Do that not work for 64bit?

rbray89:
It should work on the 64bit arch, I'd need to test to verify. 
You are correct in that it should not affect SD/eMMC boot. 

USB boot is similar to eMMC boot functionality. eMMC is not on all boards, but we still support it as a boot option for Odroid/tinkerboard.

pvizeli:
@rbray89 nice. Please update the 64bit uboot script to and we can merge it 👍

Documentation/boards/raspberrypi.md
```diff
@@ -21,14 +21,27 @@ For access to terminal over serial console, add `console=ttyAMA0,115200` to `cmd
...
 
 rpi-i2c.conf:
 ```
 i2c-dev
 i2c-bcm2708
 ```
 
+## USB Boot
+
+USB mass storage boot is available on Raspberry Pi 3B, 3B+, 3A+, and 2B v1.2.
```
cogneato: also pi 4?

rbray89: Not at the moment it seems: 
https://www.raspberrypi.org/forums/viewtopic.php?t=243995
https://www.raspberrypi.org/documentation/hardware/raspberrypi/booteeprom.md

taanczos:
It works well with element14 Pi desktop included sata usb converter and Integral 120GB mSATA SSD.
Thank you!

mattclar:
> Tested/verified on RPI3 B1.2 with Lexar USB drive.

Where do I find a rpi3 image that includes this feature?

cogneato:
@truglodite do you have the list of devices somewhere?

pvizeli:
In forum is a list

PR #442, opened 2019-07-25
make clear the usb drive needs to be connected only while configuration

capstan1:
A question I asked myself and that got asked frequently in the forums is weather the usb stick used for configuration needs to stay within the device. So I just added some line to make this more precise.


homeassistant:
Hi @capstan1, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;hassos#442).

Thanks!

PR #444, opened 2019-07-25
Update configuration.md

capstan1:
A question I asked myself and that got asked frequently in the forums is weather the usb stick used for configuration needs to stay within the device. So I just added some line to make this more precise.


pvizeli:
Doublicate

PR #459, opened 2019-08-27
Fix boot

pvizeli:


iannecj:
Should this be resolved with today's Hassio update?  my sysmon sensor still shows 1GB on a 4GB model.

pvizeli:
There is a Home Assistant / Hass.io / HassOS. Which kind of the update do you means?

PR #467, opened 2019-09-06
Support USB WIFI Dongles on rpi/rpi2

GeoffWilliams:
* Include the firmware for most WIFI USB Dongles

~~~
The old Raspberry PI 1/2 boards need USB WIFI dongles if their not going to be used in wired mode. It seems easiest to just include the firmware in the squashfs so that you can plug-in-and-go. The PR grows the images by ~50MB.

Someone on the forums encountered this error as well as me:
https://community.home-assistant.io/t/usb-wi-fi-not-recognised-when-installing-hassos-rpi2-2-12-on-rspi-2-b/132508


homeassistant:
Hi @GeoffWilliams, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;hassos#467).

Thanks!

GeoffWilliams:
Tested OK on Raspberry PI 1B and 2

pvizeli:
What have this for an affect for image size?

GeoffWilliams:
@pvizeli approx +50MB. New sizes:
189M hassos_rpi2-3.4.img.gz  
188M hassos_rpi-3.4.img.gz

tari:
I just happened to encounter the same missing firmware problem on my Pi2 and hacked around it by saving a copy of the requisite firmware file in a persistent location and setting firmware_class.path on the kernel command line to point to it.

Packaging the firmware would be much easier, since most users won't be able to figure out what I did.

ocalvo:
Could we extend this PR to support also USB WWAN Dongles?
I believe we need to add: BR2_PACKAGE_USB_MODESWITCH=y to the images.
This will install usb_modeswitch which is required to switch the dongle from storage to modem.
More info [here](http://www.draisberghof.de/usb_modeswitch/#intro)

GeoffWilliams:
@ocalvo I couldn't see that switch enabled on any of the other builds:

```
fgrep BR2_PACKAGE_USB_MODESWITCH buildroot-external/configs/* -Rl
$
```

... So I think its more of project wide decision that should have its own ticket/PR?

ocalvo:
> @ocalvo I couldn't see that switch enabled on any of the other builds:
> 
> ```
> fgrep BR2_PACKAGE_USB_MODESWITCH buildroot-external/configs/* -Rl
> $
> ```
> 
> ... So I think its more of project wide decision that should have its own ticket/PR?

It has it own [issue](https://github.com/home-assistant/hassos/issues/484)
I am currently testing the change in ova. Does anyone know long does this thing takes to build?

I will be submitting a separate PR.

pvizeli:
How do you generate the list of those devices?

The question will be, how do we decide which wifi stick will be around. If we can solve that question, we can do it. Maybe also an option to allow binary firmware on boot partition?

GeoffWilliams:
I found the list inside buildroot in section `menu "WiFi firmware":

https://github.com/home-assistant/hassos/blob/dev/buildroot/package/linux-firmware/Config.in#L75

I'm not sure what dongle I have it just says "The Pi Hut" on the outside. I figured I'd just enable everything and then there's a pretty good chance that whatever gets plugged in works.

GeoffWilliams:
is there any interest in this? I really like being able to upcycle my old PI 1. We could I guess just include the firmware for people affected by this ticket but seems easier and more practical to just include all WIFI drivers for 50MB.

bruecke:
I would love to see those firmware files included. I tried to build the image for rpi1 myself, but it failed at some point building the docker container. So I'm currently using your (older) pre-build image.
Hass itself runs fine on the old PI 1.

camjac251:
@pvizeli Can this be reopened? I am using a Pi 2b and am having to compile it myself using Geoff's additions because it's not available in the official image.

flavioarcega:
While reading this [Configuration](https://github.com/home-assistant/operating-system/blob/dev/Documentation/configuration.md) instructions I was thinking...

Is there a chance to produce a .raubc file to inject those packages through ha os command?

sairon:
@flavioarcega Please don't gravedig this old PR for discussion :sweat_smile: In the meantime we added support for a bunch of drivers and firmwares across all support boards, if there's something you're still missing, it'll be best to create a ticket, we can consider it.

About your question, it's possible, just not using `ha os` commands - to install a custom `raucb` update file, you must use the `rauc` command directly in the root shell. But again, this is not the correct place to ask, you can use [discussions](https://github.com/home-assistant/operating-system/discussions).

PR #470, opened 2019-09-06
BACKUP fs support for HassOS

pssc:
These changes are the most non invasive way I could see to add the ability to do local backups onto a different device, that didn't adversely affect boot time.

At boot config time this change mounts a filesystem labeled BACKUP to /mnt/data/supervisor/backup

This has been tested using the ova image in a virtual box both with and without BACKUP fs present.

### DIFF
--- a/Documentation/configuration.md
+++ b/Documentation/configuration.md
@@ -1,5 +1,8 @@
 # Configuration
 
+## Alternate Backup/Snapshot File system
+Format a USB stick/Drive with a FAT32/EXT4(Best without a journal on flash)/Ext-FAT(not yet supported) file system and label it `BACKUP` case sensitive. This is mounted at configuration only if the supervisor container is not running which is usually only at boot. **It is not safe to remove this unless HassOS is powered down.**
+
 ## Automatic
 
 You can use an USB drive with HassOS to configure network options, SSH access to the host and to install updates.
--- a/buildroot-external/rootfs-overlay/usr/lib/systemd/system/mnt-data-supervisor-backup.mount
+++ b/buildroot-external/rootfs-overlay/usr/lib/systemd/system/mnt-data-supervisor-backup.mount
@@ -1,0 +1,10 @@
+[Unit]
+Description=HassOS data backup partition
+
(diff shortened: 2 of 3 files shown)


pvizeli:
You need to explain a bit more:
- Workflow for user
- UI support for this feature
- Possible issues
- Pro/Contra

pssc:
You need to explain a bit more:

Hi @pvizeli sure.

- Workflow for user.
Insertion of usb stick with correct fs label (Documented)

- UI support for this feature
None needed, though and indicator that this is active might be useful, other than checking the logs.

- Possible issues
Removal of device during running (Documented)
Supervisor starting before fs mounted (restarts supervisor as a result)

Pro
- It's simple
- uses existing methods already used for config usb stick
- Provides backup potentially to a different block device to that which hassio is running on (my backups have been enten more than once on the same sd card)
- Potential reduction of IO on hassio block device for backups (increasing speed & reducing wear on hassio sdcard)

Cons
- Must power off device for removal
- restart of supervisor if running (seems to only happen on first boot)

boesing:
Hm. I'd like to have this feature aswell. Since there is support to move the data filesystem to a different storage, I'd also like to move the backup filesystem to a different one.

Having the backup partition on the physically a different filesystem increases reliability. 
Is there any chance to get this PR opened up again @pvizeli?

pvizeli:
Is not part of the OS: https://github.com/home-assistant/supervisor/issues/1866

boesing:
I am not sure if I dont see what the supervisor has to do with the location where the backup is stored or you get us wrong.

I think, there are two ways to approach this feature:
1. supervisor lets us configure the backup location (which is by default the `/backup` folder which is symlinked in the SSH addon). we could just provide the external hard drive here. 
2. HassOS gives us the ability to configure the `/backup` location and thus let us choose the external hard drive here 

Was that more clear or am I the one who dont get what you want to say?

pssc:
Ok So the supervisor is doing the backups? so the filesystem has to be in the right place now in the filesystem is put in place by the underlying OS?

PR #485, opened 2019-09-26
Add support for SMS integrations with USB/GSM modem dongles

ocalvo:
Fix for issue [484](https://github.com/home-assistant/hassos/issues/484)
This is important to allow hassio installation to be able to send notifications while offline using SMS.
More info [here](https://community.home-assistant.io/t/send-sms-with-usb-gsm-modem-when-alarm-triggered/28942/33)

This change does 2 things:
1. Install the drivers for USB modems.
2. Install usb_modeswitch to deal with dual mode USB devices. More info [here](http://www.draisberghof.de/usb_modeswitch/)

Size impact: (Measured in VDI image)
70.26 Kb which is increase of 0.040%

Test:
Tested on a VDI virtual appliance with [this device](https://www.amazon.com/gp/product/B01N6P3HI2/ref=ppx_yo_dt_b_search_asin_title?ie=UTF8&psc=1)
After this change the modem is visible to hassio.

Annotation 2019-09-26 141636


pvizeli:
How do you now use that modem connection?

ocalvo:
Usb_modeswitch does not interact with network or modem manager, it interacts with udev to setup new USB serial devices under /dev/

I will post details on how we are using this from home assistant

balloob:
Using something like this would be easier for everyone: https://www.home-assistant.io/components/netgear_lte/

ocalvo:
I actually using one, and there are multiple reasons why this will not work for everyone:
1. Cost, this device is hundreds of dollars  VS 30 bucks.
2. Power consumption, ideally you want to have power backup, now you need to feed power to 2 devices instead of just one. 
3. This is a terrible switch, I have a 1gigabit/s service and it can only route about 200mbit/s
4. GSM service: I use my netgear with google fi as backup but Google fi does not support SMS and their data Sims.
5. Complexity, more points of failure, if the internet is down of if the netgear is down, then you won't actually be able to get notifications.

ocalvo:
I am also planning on writing a proper SMS integration for home assistant to make it super simple for everyone. But I need to ensure that such integration will work in All flavors of home-assistant.

Hassos is the only flavor will that this won't work today.

pvizeli:
You need this to send SMS only?

ocalvo:
This is the complete use case in with this fix on homeassistant:

send_sms.sh script:
```
#!/bin/bash

DEST=$1
# Modify the MD variable to match your hardware configuration
MD="/dev/ttyUSB0"

if [ -z "$2" ]
then
    read message
else
    message="$2"
fi

echo "Sending $message to $DEST"

echo "AT+CMGF=1" > $MD
sleep 1
echo "AT+CMGS=\"$DEST\"" "OK" > $MD
sleep 1
echo "$message^Z" > $MD
```
configuration.yaml:
```
notify:
  - platform: command_line
    name: sms_person1
    command: !secret notify_sms_person1
  - platform: command_line
    name: sms_person2
    command: !secret notify_sms_person2
```
secret.yaml:
```
notify_sms_person1: "/config/send_sms.sh +1NNNNNNNNNN"
notify_sms_person2: "/config/send_sms.sh +1NNNNNNNNNN"
```

Discussion about this integration [here](https://community.home-assistant.io/t/send-sms-with-usb-gsm-modem-when-alarm-triggered/28942/19)

ocalvo:
> I will ask someone to test that with an RFlink adapter. Since we did not know what the bug affected with USB dongles, I'm a bit carefully.

This change is a "pay for play" change, and even then, this is only affecting if the new serial port is referenced by home-assistant configuration.
I will try to get a hold of a RF stick, do you know where I can buy one?

ocalvo:
Here is nmcli con show:
nmcli-con-show

Here is nmcli:
nmcli-1

ocalvo:
I have read the warning about ModemManager, I also happen to use a dual dongle for Z-Wave and ZigBee, and given that this change does not interact with the modemmanager or the networkmanager, I wont interfere with RFLink also.

buildroot-external/kernel/device-support.config
```diff
@@ -1,9 +1,12 @@
 CONFIG_USB_SERIAL=y
 CONFIG_USB_SERIAL_GENERIC=y
+CONFIG_USB_SERIAL_OPTION=y
+CONFIG_USB_SERIAL_WWAN=y
```
pvizeli: Let load it as a module with `m`

pvizeli:
You can approve that:
- Zwave/Zigbee is working (on HA)
- The use case is solved and working on HA

ocalvo:
Everything working as runtime module in the kernel:
lsmod

PR #486, opened 2019-10-01
Update azure-pipelines-ci.yml

reharmsen:


pvizeli:
Hi @reharmsen 
Would be nice if you are able to offload the docker validation and maybe also the script validation as template to https://github.com/home-assistant/ci-azure/tree/master/templates. The hassio-addons version of dockerfile validation support to scan for dockerfiles, maybe you should use that code for the template.

PR #502, opened 2019-12-09
Fix RPI bluetooth issues

pvizeli:


scop:
Hi @pvizeli, I'm running r3 build 5 on rpi3 off a usb ssd, works fine. Except when I turn on bluetooth device tracker, bt doesn't seem to work and HA almost drops from the network, I start getting like 20 second pings, a lot of packet loss etc. Using wifi here. I think bt used to work sometime earlier on the same rpi3 when I ran hassos 2.x off the sd card.

What kinds of bt issues does this PR fix? Is there a build of this that I could easily try out?

pvizeli:
@scop try 3.6 - They reverted the bluetooth/wifi binary driver for 20 days. I include that update on version 3.6

rangerzhou:
> @scop try 3.6 - They reverted the bluetooth/wifi binary driver for 20 days. I include that update on version 3.6

Hi pvizeli,
I cannot find the hassos_rpi3-3.6.img.gz, is the Raspberry3B+ image released?

scop:
@pvizeli thanks, but unfortunately same problem with 3.6. I will try to narrow it down if it's the 3.x series or USB disk or something else.

pvizeli:
@scop strange because we use the same kernel, firmware, driver blobs as raspbian

scop:
Yeah, I haven't been able to pinpoint the problem -- I can reproduce it with setups booting from usb or sdcard, hassos 2.12, 3.5, and 3.6.

So I guess it must be one of (in no particular order):
- this was broken also in the 2.x releases
- my rpi3 is broken
- the problem was caused by enabling usb bootability (`program_usb_boot_mode=1`)

I *think* I tried with another rpi3 already, but that one also has been usb boot enabled. Will continue trying to hunt this down.

Issue #482, opened 2019-09-16
add support for qemu-guest-agent


JoeyG1973:
Please add support in hassio for the qemu-guest-agent

https://wiki.qemu.org/Features/GuestAgent

buginside:
I use hassos in Proxmox VE, qemu-guest-agent could be used for:

1-Properly shutdown the guest, instead of relying on ACPI commands
2-Freeze the guest file system when making a backup

pvizeli:
I don't block this, make sure that the guest agent is only started on the correct virtualisation like vmware tools.

JoeyG1973:
@pvizeli I am not sure what you are trying to say.

There doesn't seem to be a straight forward way to install qemu-guest-agent or anything to hassos. This can't be installed as a container.

pvizeli:
It's buildroot on that repository, not containers. You need learn how buildroot work and you are able to create an PR with this feature. You can look at it as hacktoberfest chellange

mihalski:
I searched to find how it is possible to determine the virtualisation technology being used and found:

https://unix.stackexchange.com/questions/89714/easy-way-to-determine-virtualization-technology

It seems dmidecode would be a really useful command to determine the virtualisation technology being used:

```
dmidecode -s system-product-name
VMware Virtual Platform

dmidecode -s system-product-name
VirtualBox

root@router:~# dmidecode -s system-product-name
KVM
```
Would it be reasonable to include it in Hassos for this purpose?

Regards,
Michal

pvizeli:
Because we use systemd, you can easy say which service should start on which virtualization. We do that for vmware tools.

However, I do not add or work on this feature. If someone implement it and make a PR, I'm fine to merge this.

mihalski:
That's way over my head for now.

I was not able to figure out how VMware was detected and then I couldn't figure out how to build (or if it's possible) the qemu-guest-agent without also building the entirety of Qemu.

Maybe someone else is a bit more capable? I should be able to test patches if that helps.

pvizeli:
https://buildroot.org/downloads/manual/manual.html

We simplify some steps with make script and docker but to contribute to this project, you need understand how buildroot work. For small things like detect the virtualisation, I can help you on the PR.

jens-maus:
[...]

PR #505, opened 2019-12-11
Add qemu guest agent

pvizeli:
Fix: #482


moeffju:
I'm not sure whether this is supposed to be working in the Release 3 Build 9 because it's not explicitly listed in the changelog for that version, only a previous one, but I'm using HassIO/HassOS on Proxmox and the Qemu Guest Agent is not starting because

```
critical: failed to create guest agent channel
critical: failed to initialize guest agent channel
```

The unit is then marked as failed for restarting too quickly.
Do I need to configure anything else in the guest for this to work?

moeffju:
> Do I need to configure anything else in the guest for this to work?

The answer (Proxmox 6) was to enable the Qemu agent support for the node, then shutdown and start the node. That made the character device show up and let the qemu-ga start fine guest-side.

PR #512, opened 2019-12-14
Fix a bug where GSM modem is not visible

ocalvo:
In order for usb_modeswitch to be able to find the devices, the kernel must have already load the drivers.
Even if we managed to run usb_modeswitch after udev rules are evaluated, it will still be too late as the dockers containers will still not see the GSM device.


ocalvo:
@pvizeli could you review please?

pvizeli:
Can we not use udev to load this?

ocalvo:
We already added udev rules as part of the original change, but these won't load because it's execution order is too early (40) even when configuring kernel load modules via USB stick.

In order to workaround we would have to duplicate the existing udev rule into the USB stick to execute with a lower priority.

But if feels incorrect that we have to duplicate an existing udev rule into the usb stick.

ocalvo:
It seems that there is something going with the actual processing of udev rules:
Even after adding the rule to call usb_modeswitch the rule stop processing with this error:
bug 01

I am running in a RPi 4
I will test this on a VM image to see if I observed the same behavior.

ocalvo:
There is something wierd going on, the bug does not repro on a brand new VM, but it repros when in my production RPI4.
I am continue to investigate.

ocalvo:
So, in RPI4, it had to create a new file named udev/10-gsm-modem.rules in the USB config.
With this content:
```
ACTION=="add" \
, ATTRS{idVendor}=="12d1" \
, RUN+="/sbin/usb_modeswitch -X -v 12d1 -p 14fe -W"
```
Notice that 14fe is my specific product id and 12d1 is the vendor id, these can be read from the output of lsusb.

However, none of this is necessary in the test environment using a VM.
For now, I am going to close this PR.

ocalvo:
I have created a new [issue](https://github.com/home-assistant/hassos/issues/513)

PR #515, opened 2019-12-17
BACKUP fs support for HassOS

pssc:
These changes are the most not invasive way I could see to add the ability to do local backups onto a different device, that didn't adversely affect boot time.

At boot config time this change mounts a filesystem labeled BACKUP to /mnt/data/supervisor/backup

This has been tested using the ova image in a virtual box both with and without BACKUP fs present.

**Workflow for user.**
Insertion of device (pre boot) with correct fs label (Documented)

**UI support for this feature**
None needed, though an indicator that this is active might be useful, other than checking the logs.

**Possible issues**
Removal of device during running (Documented)
Supervisor starting before fs mounted (restarts supervisor as a result)

**Pros**

- It's simple
- uses existing methods already used for config usb stick
- Provides backup potentially to a different block device to that which hassio is running on (my backups have been enten more than once on the same sd card)
- Potential reduction of IO on hassio block device for backups (increasing speed & reducing wear on hassio sdcard)

**Cons**

- Only activated on boot
- Must power off device for removal
- restart of supervisor if running (seems to only happen on first boot)

### DIFF
--- a/Documentation/configuration.md
+++ b/Documentation/configuration.md
@@ -1,5 +1,8 @@
 # Configuration
 
+## Alternate Backup/Snapshot File system
+Format a USB stick/Drive with a FAT32/EXT4(Best without a journal on flash)/Ext-FAT(not yet supported) file system and label it `BACKUP` case sensitive. This is mounted at configuration only if the supervisor container is not running which is usually only at boot. **It is not safe to remove this unless HassOS is powered down.**
+
 ## Automatic
 
 You can use an USB drive with HassOS to configure network options, SSH access to the host and to install updates.
--- a/buildroot-external/rootfs-overlay/usr/lib/systemd/system/mnt-data-supervisor-backup.mount
+++ b/buildroot-external/rootfs-overlay/usr/lib/systemd/system/mnt-data-supervisor-backup.mount
@@ -1,0 +1,10 @@
+[Unit]
+Description=HassOS data backup partition
+
+[Mount]
+What=/dev/disk/by-label/BACKUP
+Where=/mnt/data/supervisor/backup
+Type=auto
+Options=nosuid,relatime,noexec
+SloppyOptions=yes
+
--- a/buildroot-external/rootfs-overlay/usr/sbin/hassos-config
+++ b/buildroot-external/rootfs-overlay/usr/sbin/hassos-config
@@ -5,6 +5,26 @@ USB_CONFIG="/mnt/config"
 CONFIG_DIR=""
 USE_USB=0
 
+
+# Check and mount usb BACKUP to folder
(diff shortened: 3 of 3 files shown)


pssc:
@pvizeli what would you like to see to add this sort of functionality

pvizeli:
Why you can't solve this use case with an add-on?

pssc:
> Why you can't solve this use case with an add-on?

Hassos does the backups the back filesystem therefore needs to be mounted there not it an addon not make the vfs consistent.

Example from ssh addon

```
➜  ~ mount -t nfs -o nolock blackpi3.lan:/srv/backup/exports/hassio/Test /share/remote/backup 
➜  ~ ls -la /share/remote/backup 
total 210848
drwxr-xr-x    2 root     root          4096 Feb  7 03:00 .
drwxr-xr-x    3 root     root          4096 Jan 19 01:25 ..
-rw-r--r--    1 root     root             0 Jan 16 21:37 .remote_backup_dir
-rw-r--r--    1 root     root      34723840 Feb  5 03:00 04072444.tar
-rw-r--r--    1 root     root       6051840 Feb  1 03:00 92e14531.tar
-rw-r--r--    1 root     root      35174400 Feb  3 03:00 b07e8052.tar
-rw-r--r--    1 root     root      35041280 Feb  7 03:00 c0ee28b6.tar
-rw-r--r--    1 root     root      34979840 Feb  4 03:00 c79eecd2.tar
-rw-r--r--    1 hassio   hassio         805 Feb  6 12:19 database.png
-rw-r--r--    1 root     root      35153920 Feb  2 03:00 e1a34a0c.tar
-rw-r--r--    1 root     root      34754560 Feb  6 03:00 f28a578c.tar
-rw-r--r--    1 hassio   hassio         561 Feb  6 12:21 nas.png
-rw-r--r--    1 hassio   hassio         459 Feb  6 13:14 transfer-down.png
➜  ~ docker exec -it hassio_supervisor bash     
bash-5.0# ls -la /data/share/remote/backup/
total 8
drwxr-xr-x    2 root     root          4096 Jan 19 01:25 .
drwxr-xr-x    3 root     root          4096 Jan 19 01:25 ..
bash-5.0# 
```

pvizeli:
Look at the exists backup add-ons. They map it inside the add-on not over the hole platform and copy the backup to the remote system

pssc:
@pvizeli I am well aware of this and I have before I started and they don't satisfy this need.  The only current viable solution is to take this off a hassio install,  using one of these addons that means you backing up onto your SD card increasing IO wear on the card, and worse possibly writing to an already failing card and toasting your system (my experience multiple times). This change would allow backup to an independent usb medium directly.  I am trying to engage in a constructive dialogue to make this better.

Issue #529, opened 2020-01-04
HassOS on Proxmox - USB TI 3410/5052


darekhome:
**Home Assistant release with the issue:**
 Home Assistant 0.103.5

**Operating environment (HassOS/Generic):**
 System HassOS 3.7

**Supervisor logs:**

**Description of problem:**

Newest 'hass.io' on newest 'proxmox'.
Hardware 'Intel-NUC'.
Device connected to USB port, but not seen in 'hass.io'.
The same tested on 'plain PC' (DELL OptiPlex 780) - problem the same.
Only seen 4 'ttySx' ports (x = 0...3).

<img width="220" alt="hardware" src="https://user-images.githubusercontent.com/55238173/71769145-77e0ce00-2f1d-11ea-86a4-a23b649ec530.png">

On 'proxmox' the device is enabled and seen:

<img width="379" alt="USB" src="https://user-images.githubusercontent.com/55238173/71769197-3866b180-2f1e-11ea-9647-51e53cf2bb01.png">

darekhome:
It is Proxmox 6.1 with kernel 5.3...
And I use the highest version of the kernel with hass.io...

pvizeli:
I means what kernel option/module are needed for it?

darekhome:
With hass.io installed on ex. IntelNUC no any special options are required. The device is seen something like ttyUSB0.
But when (has.io) installed with proxmox nothing is seen...

So simply answering your question - no special kernel option is needed.

szurr:
Hi 
In my case workflow looks
1.Proxmox add devices (USB port/vendor)
2.Proxmox restart ! Whole vm/lxc
3. Hassio I see USB device and make changes to config
4. Restart and working (most of the time)

darekhome:
Well, this is what I expected...
But in my case it is not working.
I hass.io I always see (ONLY) ttyS0,...,ttyS3. Under hardware button.

pvizeli:
Look like our kernel need a module for this. The USB redirection on vmware/hyperV work. The question is only, which module would be needed on the kernel side...

pvizeli:
I put this to next release, maybe it work. I can't test it

pvizeli:
If someone will test it `hassio os update --version 3.8`

Iam-lcl:
Tested with a clean install of 3.8
no USB showed up.

pvizeli:
[...]

PR #542, opened 2020-01-13
Add support for TI USB 3410 or 5052 serial devices

pvizeli:
Fix: #529


scstraus:
Hi @pvizeli 

After migrating to Debian from Ubuntu in order to stay in line with ADR-0014, I now have the same issue where Debian will not load the firmware for the HGI-80 (it was plug and play on Ubuntu). Is there any way I could ask you for a hint how you added this support so I can try to recreate it in Debian?

pvizeli:
Please read: https://github.com/home-assistant/architecture/blob/master/adr/0014-home-assistant-supervised.md#required-expertise

PR #552, opened 2020-01-26
Include oath-toolkit package

nkgilley:
Needed for `alexa_remote_control.sh` from https://github.com/thorsten-gehrig/alexa-remote-control


pvizeli:
Thanks for your contribution.

HassOS is a docker based system. That means everthing run in a docker container and not on Host. The Host or HassOS as embedded Linux is only the minimal to run docker/hass.io

You can create an add-on to run your script in a container.

nkgilley:
@pvizeli I installed the tool I need in an addon, but I'm not sure how I can execute it from the host?  I don't see docker installed.

I need this tool installed on the host as it as a dependency for a shell command script and the shell command is stored in /config.

PR #587, opened 2020-02-16
Rename Hass.io

pvizeli:
`hassio` or `hass.io` is now depricated to be called for the network layer and tagging system.


buildroot-external/busybox.config
```diff
@@ -320,7 +320,7 @@ CONFIG_TRUE=y
 CONFIG_TRUNCATE=y
 CONFIG_TTY=y
 CONFIG_UNAME=y
-CONFIG_UNAME_OSNAME="Hassio/OS"
+CONFIG_UNAME_OSNAME="HassOS"
```
frenck: Sure you want to keep that term around?

pvizeli: The OS is called on that layer HassOS and will be forever because it will break everthing if I rename this. For busybox, I can use `HomeAssistant/OS` but it's weird if the host information is different to the from uname?

pvizeli: It's like the network layer, they will be forever hass.io otherwise we brick the system.

frenck: Yeah, get it. Let's keep it 👍 (Some awesome nostalgic references deep in the code base reminds us of the awesome history, nothing wrong with that 😉 )

Issue #491, opened 2019-10-19
REQUEST: USB camera / v4l-utils package


pergolafabio:
Hi,  is it possible to add this package to HassOS , this enables /dev/video* to use as ffmpeg for USB cameras attached to HassOS : v4l-utils

this package is also needed for MotionEYE to support USB cameras
i  have as example bought a USB converor like below:

https://www.amazon.com/Easyday-EasyCAP-S-Video-Grabber-Capture/dp/B01C2M85PA

accoridin to wiki : https://linuxtv.org/wiki/index.php/Easycap
those drivers are present already in kernel : 
```
Making it work
STK1160 based devices are supported (video and audio) under Linux by the easycap kernel module (kernel 2.6.38 and later) and further on by the stk1160 kernel module (kernel 3.7). More information on this wiki: Stk1160
```

not sure if this is the right place to ask? but thnx in advance

PR #609, opened 2020-03-21
Add USB video kernel support

DaAwesomeP:
This looks excessive but this seems to be required given the completely different requirements for the different webcams I tested with (I was able to get them all to appear in `/dev` and have their drivers loaded). Since support is not quite baked into everything else yet, I was not able to test beyond ensuring they show up and the right modules are loaded and that `dmesg` doesn't complain.

I saw some of the other boards have some subset of these parameters, but I did not modify them since I have no way to test them and am unsure of any other platform specifics.

This includes webcams and most simple USB capture cards, so it should solve #491 at least for `intel_nuc`.


DaAwesomeP:
@pvizeli Looking at past pulls it looks like maybe this should go in ` operating-system/buildroot-external/kernel/device-support.config `? Please let me know if I should make that change.

Assuming there are no caveats with any of these configs in the other platforms, I suppose the duplicate configs lines could be removed from other board-specific configs if this is added to `device-support.config`?

DaAwesomeP:
OK, I just installed the latest (unreleased) version of the motion eye addon that adds the video option to the addon config, and I can confirm that this works! Cameras can be viewed and controlled properly.

pvizeli:
What will be the +size of all the modules? Can you group those options and explain a bit more about why that module/group is required and why you listen in there.

> device-support.config
Yes, devices they are plug&play like USB stuff and can be used also by other hardware, need to listen on this file. Just hardware they are only recommended for this kind of image, are in the board kernel config.

DaAwesomeP:
> What will be the +size of all the modules?
My build went up by about two MB.

> Can you group those options and explain a bit more about why that module/group is required and why you listen in there.

Yes, commit imminent!

> Yes, devices they are plug&play like USB stuff and can be used also by other hardware, need to listen on this file. Just hardware they are only recommended for this kind of image, are in the board kernel config.

Cool, I will move it over and sort through any duplicates in other files. Yes, that makes sense (i.e. the RPI video stuff should only be in those configs)

DaAwesomeP:
OK, without the video stuff intel_nuc is 219M. With the video configs it is 221M.

I have loaded this updated commit on my device and I can confirm it works.

buildroot-external/kernel/device-support.config
```diff
@@ -24,3 +24,171 @@ CONFIG_BT_RTL=m
...
+CONFIG_VIDEOBUF2_DMA_CONTIG=m
+CONFIG_VIDEOBUF2_VMALLOC=m
+CONFIG_VIDEOBUF2_DMA_SG=m
+CONFIG_VIDEOBUF2_DVB=m
+CONFIG_DVB_B2C2_FLEXCOP=m
+CONFIG_VIDEO_SAA7146=m
+CONFIG_VIDEO_SAA7146_VV=m
+CONFIG_SMS_SIANO_MDTV=m
+CONFIG_SMS_SIANO_RC=y
+CONFIG_VIDEO_V4L2_TPG=m
```
pvizeli: Move this to the common devices

buildroot-external/kernel/device-support.config
```diff
@@ -24,3 +24,171 @@ CONFIG_BT_RTL=m
...
+CONFIG_MEDIA_DIGITAL_TV_SUPPORT=y
+CONFIG_MEDIA_USB_SUPPORT=y
+CONFIG_MEDIA_PCI_SUPPORT=y
+CONFIG_VIDEO_DEV=m
+CONFIG_VIDEO_V4L2_SUBDEV_API=y
+CONFIG_VIDEO_V4L2=m
+CONFIG_VIDEOBUF_GEN=m
+CONFIG_VIDEOBUF_DMA_SG=m
+CONFIG_VIDEOBUF_VMALLOC=m
+CONFIG_V4L_PLATFORM_DRIVERS=y
```
pvizeli: This part also to the common devices

DaAwesomeP:
> Move this section to the common devices
> This section also to the common devices

OK, I have grouped the first and last sections together

> Can we remove that # Analog/digital TV/input USB devices ? We have no use case for this and also no add-on which brings tv support or something else. In the end, I need to maintain that until the end of my life and is not funny to fix the issue or add a kernel patch for things where no one is used.

Yes, good point—there is no need for tuners. This list originally came from the Odroid config that I trimmed down a little.

I removed all of the DVB devices (TV tuners) and went through the rest of the non-webcam device configs to ensure they are all used for direct capture card modules (e.g. composite/component/S-video input only). These were originally labeled as "TV" in the Linux kernel comment since the previous most common use was to plug them into the output of a cable box and they are often found inside of devices that also have a demodulator.

This is what is being used in #491. These analog non-TV, tunerless capture devices are now most often used to connect to an analog security camera or the multi-view output of an NVR for remote monitoring/streaming.

I have not tested this commit and won't have a chance to until the end of the week, but I am 90% sure it should work fine since I only removed devices.

pvizeli:
ok. Thanks for contribution

DaAwesomeP:
Thank you for maintaining such an awesome project!

PR #614, opened 2020-03-25
Attempt to fix bluetooth communication problem on rpi3b

lucagiove:
Attempt to fix bluetooth communication problem on rpi3b.
I couldn't test this pull request because of #613 
  
This problem is referenced in many different issues, some of them:
Bluetooth: hci0: Frame reassembly failed (-84)
https://github.com/raspberrypi/firmware/issues/1150
    
Error communicating with local bluetooth adapter (RPi3)
https://github.com/home-assistant/core/issues/30147
    
Xiaomi BLE Temperature and Humidity sensor no longer working after update from 0.95.4 to .096.5
https://github.com/home-assistant/core/issues/25704
    
Xiaomi BLE Temperature and Humidity sensor
https://github.com/home-assistant/core/issues/24313#issuecomment-578554315
    
Solution seemed to reeduce the boud rate of the serial connection to BT
on rpi3b where flow control is missing.
    
The patch to the original btuart file is to reduce the baudrate only for rpi3b models where flow control is missing:
    
    -               $HCIATTACH /dev/serial1 bcm43xx 921600 noflow - $BDADDR
    +               $HCIATTACH /dev/serial1 bcm43xx 460800 noflow - $BDADDR
    
I'm not a buildroot expert but the best way seemed to move the upstream btuart download in a PRE_PATCH_HOOK so that standard buildroot way of patching can be applied.

PR #616, opened 2020-04-01
Fixed bluetooth-bcm43xx build

lucagiove:
Fix for #614 I've been able to build the package and test my code (didn't work :)) this is a fixed version.
I'm still not able to build the whole image but I saw the patch applied in the target file system.

Long story short I brutally added patch command after download, more details below.

> I'm not a buildroot expert but the best way seemed to move the
> upstream btuart download in a PRE_PATCH_HOOK so that standard
> buildroot way of patching can be applied.

I've been finally able to test it and the PRE_PATCH_HOOK does not work
seems it's never triggered.
I tried also th POST_RSYNC_HOOK that works but still patch is not
applied.
I think that patch is not detected (odd, patch naming seems complient)
or since source code is not downloaded buildroot thinks that patch is
not needed, couldn't sort it out.

### DIFF
--- a/buildroot-external/package/bluetooth-bcm43xx/bluetooth-bcm43xx.mk
+++ b/buildroot-external/package/bluetooth-bcm43xx/bluetooth-bcm43xx.mk
@@ -10,15 +10,15 @@ BLUETOOTH_BCM43XX_LICENSE_FILES = $(BR2_EXTERNAL_HASSOS_PATH)/../LICENSE
 BLUETOOTH_BCM43XX_SITE = $(BR2_EXTERNAL_HASSOS_PATH)/package/bluetooth-bcm43xx
 BLUETOOTH_BCM43XX_SITE_METHOD = local
 
-define BLUETOOTH_BCM43XX_FIRMWARE_AND_HELPERS_DOWNLOAD
+define BLUETOOTH_BCM43XX_BUILD_CMDS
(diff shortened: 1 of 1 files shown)


pvizeli:
I will get the firmware of BT anyway out of that repository into our own repository for blobs

pvizeli:
If you are able to build the OS and have free time resource, I'm all time happy for help to make faster releases

lucagiove:
I think about patching because a discussion with the raspberry maintainer is still in progress maybe they will change that script upstream.
Anyway I think is a better option to patch in order to easily support further upstream changes.

I've been finally able to build the OS and it's up and running (and fixed) since more than 24h I built the dev branch plus my patch.

pvizeli:
I speak more about to help maintain the downstream RPI stuff like update kernel & firmware ect

lucagiove:
Sure I can help, I'm able to build, I've not much free time but I can try.
I mean right now during the quarantine I've some time :)

lukaszo:
Hi, big thank you for the patch. I can confirm that it fixed my issue with hanging bt on rpi.

lukaszo:
@lucagiove  unfortunately after a week it stopped to work again. The patch for sure helps. Without the path it would stop working after few hours.

In dmesg I see: `Bluetooth: hci0: command 0x200b tx timeout`
I also see `brcmfmac: brcmf_cfg80211_set_power_mgmt: power save enabled`  Could this be related?

PR #625, opened 2020-04-18
Migrate uefi to uboot

pvizeli:


pvizeli:
Not possible with u-boot 2020.04 because they miss IO block device support for application

PR #664, opened 2020-05-11
ova: Use a serial console

cfergeau:
For VM images, users are more likely to run them headless, and to try to get VM output from a serial console.
I've tested this by directly editing cmdline.txt in a VM, I did regenerate a new image from this repository.


pvizeli:
That is fine that you edit this file and it is persistent over updates. That is the idea of how you can customize the setup. The issue is, it will not boot if you have not that console around, so we set the minimal configuration which should work for 98% of people, and the other 2% need customize itself.

PR #693, opened 2020-05-27
Fix issue with Odroid-N2 on USB

pvizeli:


pepeEL:
It is mean that we can run from USB HassOS on Odroid N2 ?

bramkragten:
This is just a bug fix where USB devices could cause a crash.

pepeEL:
OK.
But is any chance to add to operatin system option for Odroid N2 to boot from USB memory ?

bramkragten:
Don't use (closed) PR's for questions/requests.

pepeEL:
But where i can ask ?

frenck:
<https://www.home-assistant.io/help>

Issue #660, opened 2020-05-09
Cannot Deploy OVA onto ESXi 6.7


dancwilliams:
**HassOS release with the issue:**

hassos_ova-4.6.ova

**Supervisor logs:**

N/A

**Journal logs:**

N/A

**Kernel logs:**

N/A

**Description of problem:**

I am attempting to import the beta OVA image onto a VMWare ESXi 6.7 host I have in my lab.  I am receiving the following error when attempting to deploy:

`Issues detected with selected template. Details: - 23:5:PARSE_ERROR: Parse error: Undeclared namespace prefix "vmw" (for attribute "osType") at [row,col,system-id]: [23,93,"descriptor.ovf"].`

I am using the following VMware Version: `VMware ESXi, 6.7.0, 14320388`

If there are any further details needed just let me know.  Thanks!

fabaff:
Ok, just to be sure, the hashes in the ova file are a match.

```bash
$ tar -xvf hassos_ova-4.6.ova
home-assistant.ovf
home-assistant.vmdk
home-assistant.mf
$ cat home-assistant.mf 
SHA256 (home-assistant.ovf) = aa59cd56f4712c871b507e5d148c273bb071734934d9747dbe44e90cd412123b
SHA256 (home-assistant.vmdk) = 6d8895600afcf44144100c56a9d61c5451dfd9673dcb91c74b331eaf307fc17f
$ sha256sum home-assistant.ovf home-assistant.vmdk
aa59cd56f4712c871b507e5d148c273bb071734934d9747dbe44e90cd412123b  home-assistant.ovf
6d8895600afcf44144100c56a9d61c5451dfd9673dcb91c74b331eaf307fc17f  home-assistant.vmdk
```

Looking at the `home-assistant.ovf` file. It seems that there is a mismatch between the `id` and the `osType`. I don't know if ESX 6.7 has a strong validation or if those details are right/still correct: 

```text
    <OperatingSystemSection ovf:id="102" vmw:osType="otherLinux64Guest" ovf:required="false">
```

The id 102 is "otherGuest64" according to a pretty old document I found. The id 101 is "otherLinux64Guest"

```text
    <OperatingSystemSection ovf:id="101" vmw:osType="otherLinux64Guest" ovf:required="false">
```

MaartenMees:
Hi,

I having also a problem when deploying the ova to esxi 6.7. with ovftools.

acederlund:
I'm having the same issue as well.

[...]

PR #696, opened 2020-05-29
Fix error importing OVA file to VMware

avdeevsv91:
Fixed the order of the network adapter description.


pvizeli:
And this solves the issue?

avdeevsv91:
> And this solves the issue? Edit, approved by @frenck

Yes. Another user also confirmed this https://github.com/home-assistant/operating-system/issues/660#issuecomment-629805965

PR #723, opened 2020-06-05
RaspberryPi move back to 4.19 Kernel

pvizeli:
Since the cooperation with other projects since Raspberry is going to freeze, the new kernel did not work stable with other projects like the bootloader u-boot. We stay away and look at what is going on with the RaspberryPi foundation...


jens-maus:
@pvizeli Could you please elaborate a bit more as to why you think the 5.4.x kernel doesn't work good yet with U-boot bootloader? In fact, in one of my projects (https://github.com/jens-maus/RaspberryMatic) which also uses buildroot + RaspberryPi + U-Boot I just switched to kernel 5.4.x and U-boot 2020.01 and couldn't identify any severe issues yet. So could you please explain why you think one should still stay away from kernel 5.4.x for the RaspberryPi hardware?

pvizeli:
The UART mappings fail dramatically. Also, many RPi specific libraries stop to detect the correct Pi because the device-tree is to different. So if you don't need UART and other RPi hardware and library, the Linux 5.4 work. We wait until all the library have updated the code to work correct on the 5.4 or the RPi foundation adjust the device-tree and procfs to work like with 4.19

jens-maus:
@pvizeli Thanks. Can you please be more specific? Which kind of "RPi specific libraries stop to detect correct Pi"? In my tests I couldn't see any problematic libraries not being able to identify the RPi model. In addition, RaspberryMatic of course uses the UART of the RaspberryPi4, but not via the standard linux kernel serial driver (pl011) but rather through an own raw uart kernel driver explicitly written for the usage purposes of RaspberryMatic (A dedicated serial module is used on the GPIO to communicate with HomeMatic/eQ3 devices). And at least that `generic_raw_uart` didn't show any issues so far. And what exactly should be different in procfs in their 5.4 versions compared to 4.19? Also in this regard I couldn't identify any problem so far.

LoermansA:
This merge causes a non functioning wifi module on Pi4, correct? Any word on that or on a fix?

PR #724, opened 2020-06-05
Make odroid upstream again

pvizeli:
The downstream kernel from hardkernel feels hacky. We had no issue with the upstream kernel before, just no audio. This PR migrates it back to upstream but we use 5.7 because they including all needed patch to have audio and work stable. The downstream kernel had also better network support, let's see if they lands also on 5.7


frenck:
I think that this might be a good idea in this case.

PR #726, opened 2020-06-05
Improve self healing capabilities of Supervisor service

frenck:
This extends the Supervisor start script to handle more critical situations and self-heal from them.

- Checks if the Supervisor exits too quickly. This can indicate a problem and it will clean up both the Supervisor container and images so new ones can be fetched.
- If for whatever reason the Supervisor image is missing, it will try to get the one from the updater information, if that fails, it will pull the latest.

This allows for:

- Recovering from corrupted layers in the Docker overlay.
- Recovering from broken releases.
- Manual/external interference caused by the user on the Supervisor containers by accidentally pulling incorrect images.

Basically, every reason the Supervisor wasn't able to start, it will just clean up everything, and start over fresh.

### DIFF
--- a/buildroot-external/rootfs-overlay/usr/sbin/hassos-supervisor
+++ b/buildroot-external/rootfs-overlay/usr/sbin/hassos-supervisor
@@ -25,7 +25,7 @@ if [ -n "${SUPERVISOR_CONTAINER_ID}" ]; then
     if [ -n "${SUPERVISOR_START}" ] && [ -n "${SUPERVISOR_STOP}" ]; then
         START=$(date --date="${SUPERVISOR_START}" +%s)
         STOP=$(date --date="${SUPERVISOR_STOP}" +%s)
-        SUPERVISOR_RUNTIME=$((${STOP}-${START}))
+        SUPERVISOR_RUNTIME=$((STOP-START))
         # Minimial runtime not met, remove container and image
         if [ "${SUPERVISOR_RUNTIME}" -lt "${SUPERVISOR_MIN_RUNTIME}" ]; then
             echo "[WARNING] Supervisor container exited too quickly, forcing a fresh image and container..."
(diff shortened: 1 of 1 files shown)


buildroot-external/rootfs-overlay/usr/sbin/hassos-supervisor
```diff
@@ -9,16 +9,72 @@ set -e
...
+
+# If Supervisor image is missing, pull it
+if [ -z "${SUPERVISOR_IMAGE_ID}" ]; then
+    # Get the latest from update information
+    # Using updater information instead of config. If the config version is
+    # broken, this creates a way (e.g., bad release).
+    SUPERVISOR_VERSION_UPDATER=$(jq -r '.version // empty' "${SUPERVISOR_DATA}/updater.json" || echo "")
+    SUPERVISOR_VERSION="${SUPERVISOR_VERSION_UPDATER:-latest}"
+
+    echo "[WARNING] Supervisor image missing, downloading a fresh one: ${SUPERVISOR_VERSION}"
```
balloob, line 48: Why is this a warning? Didn't we just delete it on purpose?

frenck: It doesn't have to be deleted on purpose by the system. This block stands for its own and has no relation to the if block above.

e.g., as a user, you can prune or delete your Supervisor Docker image

frenck:
Will address the shell checks. Reminds me of install spellcheck in that dev environment... 😓

pvizeli:
okay, be careful with the output of shellcheck because we use ash from busybox.

frenck:
Yeah, I realize that. Fixed the shellcheck warnings and building an ova right now to make sure the changes still work as expected.

frenck:
Adjusted shellcheck warnings and tested the affected lines.

PR #727, opened 2020-06-05
Fix Supervisor container name reference in hassos-supervisor service

frenck:
The systemd service for the hassos-supervisor service used incorrect references to the Supervisor container. This PR addresses that.

### DIFF
--- a/buildroot-external/rootfs-overlay/usr/lib/systemd/system/hassos-supervisor.service
+++ b/buildroot-external/rootfs-overlay/usr/lib/systemd/system/hassos-supervisor.service
@@ -11,9 +11,9 @@ StartLimitBurst=5
 Type=simple
 Restart=always
 RestartSec=5s
-ExecStartPre=-/usr/bin/docker stop hassos_supervisor
+ExecStartPre=-/usr/bin/docker stop hassio_supervisor
 ExecStart=/usr/sbin/hassos-supervisor
-ExecStop=-/usr/bin/docker stop hassos_supervisor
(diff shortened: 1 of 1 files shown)


buildroot-external/rootfs-overlay/usr/lib/systemd/system/hassos-supervisor.service
```diff
@@ -11,9 +11,9 @@ StartLimitBurst=5
 Type=simple
 Restart=always
 RestartSec=5s
-ExecStartPre=-/usr/bin/docker stop hassos_supervisor
+ExecStartPre=-/usr/bin/docker stop hassio_supervisor
 ExecStart=/usr/sbin/hassos-supervisor
```
balloob, line 15: I can see how this error has sneaked in, probably because of the name here. 

For a future PR, should the docker logic exists in the systemd file or should we add a `/usr/sbin/hassos-supervisor --stop` to keep all logic in 1 file?

frenck: I think having the docker stop right here in `ExecStop` is a valid case.

The `ExecStartPre` could be handled by the `hassos-supervisor` as it does a whole lot of check already.

pvizeli: Yeah, it would make sense. It would be nice to have some libs like bashio for OS to share some functions.

PR #741, opened 2020-06-14
Fix version json name if image is missing

pvizeli:


buildroot-external/rootfs-overlay/usr/sbin/hassos-supervisor
```diff
@@ -42,7 +42,7 @@ if [ -z "${SUPERVISOR_IMAGE_ID}" ]; then
     # Get the latest from update information
     # Using updater information instead of config. If the config version is
     # broken, this creates a way (e.g., bad release).
-    SUPERVISOR_VERSION_UPDATER=$(jq -r '.version // empty' "${SUPERVISOR_DATA}/updater.json" || echo "")
+    SUPERVISOR_VERSION_UPDATER=$(jq -r '.version // empty' "${SUPERVISOR_DATA}/config.json" || echo "")
```
frenck, line 45: I have played and tested with that.
However, if a release goes bad (whatever the reason), it will keep downloading the "current" broken one. That is why I later switched to the "updater" version (although not completely watertight either).

(this is where my original "curl" availability question came from)

pvizeli: The issue is, that we don't update required information, that is just a pretty simple and the risk that this bitten us in future is higher as the risk that a stable mark update going wrong (I hope so)

frenck: Yeah agreed, Ideally we should not trust local data in such situations. However, this is what we have now. So 👍

PR #742, opened 2020-06-14
Fix jitters during first update/boot

pvizeli:
Fix: https://github.com/home-assistant/supervisor/issues/1789

### DIFF
--- a/buildroot-external/rootfs-overlay/usr/sbin/hassos-supervisor
+++ b/buildroot-external/rootfs-overlay/usr/sbin/hassos-supervisor
@@ -9,7 +9,7 @@ set -e
 . /etc/os-release
 
 # Init supervisor
-SUPERVISOR_MIN_RUNTIME=30
+SUPERVISOR_MIN_RUNTIME=15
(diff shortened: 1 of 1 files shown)


buildroot-external/rootfs-overlay/usr/sbin/hassos-supervisor
```diff
@@ -27,7 +27,10 @@ if [ -n "${SUPERVISOR_CONTAINER_ID}" ]; then
         STOP=$(date --date="${SUPERVISOR_STOP}" +%s)
         SUPERVISOR_RUNTIME=$((STOP-START))
         # Minimial runtime not met, remove container and image
-        if [ "${SUPERVISOR_RUNTIME}" -lt "${SUPERVISOR_MIN_RUNTIME}" ]; then
+        # if there is no update process ongoing
+        if [ "${SUPERVISOR_IMAGE_ID}" != "${SUPERVISOR_CONTAINER_ID}" ]; then
```
frenck, line 31: This check does not belong here, it should be part of the if check that is done at line 21

PR #749, opened 2020-06-20
Bump dropbear to 2020.79

chrisaljoudi:
Most notable addition is support for ed25519 keys. [Full release notes here](https://github.com/mkj/dropbear/releases/tag/DROPBEAR_2020.79).


theseal:
A modern dropbear with support for ed25519 keys is included in the master branch of buildroot but seems not to be released yet. I guess that you rather wait for a proper release of buildroot then adding a patch for a more modern version of dropbear?

PR #765, opened 2020-07-06
add out-of-tree intel e1000e ethernet module

### LINKED ISSUE #697: intel NUC i219-v Not supported
HassOS release with the issue:
0.1010.3

Operating environment (HassOS/Generic):
HassOS, Intel NUC10FNH with NVME SSD

Supervisor logs:
N/A

Journal logs:
N/A

Kernel logs:
N/A

Description of problem:
Tried running Hassio 4.5 and 4.8 NUC install image and the network adapter isn’t being recognized.

Ubuntu VirtualBox install it works fine

sopelt:
Fixes #697 (I219-V for NUC 10)


buildroot-external/package/intel-e1000e/intel-e1000e.hash
```diff
@@ -0,0 +1,5 @@
+# from https://downloadcenter.intel.com/download/15817/Intel-Network-Adapter-Driver-for-PCIe-Intel-Gigabit-Ethernet-Network-Connections-Under-Linux-
```
pvizeli, line 1: This URL do not match with the url they are in: https://downloads.sourceforge.net/project/e1000/e1000e%20stable/$(INTEL_E1000E_VERSION)

I think we can remove this part and just hold the local one

sopelt: The URL here is where the official MD5 was located ... someone mirrors the tgz release files on sourceforge. I was not sure what the preferred source would be. We could also switch the URL in the .mk to https://downloadmirror.intel.com/15817/eng/e1000e-3.8.4.tar.gz but I was unsure what you prefer and if intel allows that.
But yeah, I'll go ahead and remove the redundant MD5 as it doesn't provide actual benefit.

pvizeli: check

PR #781, opened 2020-07-21
Allow to execute commands using enter.sh directly

agners:
This is useful e.g. if one just wants to run a build.


pvizeli:
I need fix the linter :)

PR #787, opened 2020-07-25
disable autofs

pvizeli:
Disable autofs because some kernel configs have it active and it leads just in issues


agners:
Do you have a reference or log what specific issues this leads to?

PR #793, opened 2020-07-27
Add ODROID-N2+ support

agners:
Add ODROID-N2+ support with the new SoC revision c. Extend the U-Boot
script: Assume ODROID-N2 if the SoC revision is "a" (there are only "a"
revision SoCs on ODROID N2) and assume N2+ otherwise.

Currently using overclock mode as proposed in the upstream kernel patches.


pepeEL:
Hi
On Odroid N2+ will be availeble boot directly from SSD on USB ? Becuase on Odroid N2 developer tell that will be not available.

PR #797, opened 2020-07-31
Enable systemd-logind

Konpon96:
Enabling systemd-logind allows the system to handle power button presses and shut down gracefully. It eliminates the necessity of handling power button manually in HA - in fact, it works even if the Home Assistant Docker containers are not running.

In case of Raspberry Pi, it's possible to utilize [gpio-shutdown device tree overlay](https://github.com/raspberrypi/firmware/blob/8174b21b65f26e6b71aca63b6c4686ff583cccca/boot/overlays/README#L860) in order to turn the arbitrary GPIO pin into a shutdown power button - when it's connected with the GND pin of course. What's more, if it's the GPIO3 one (the default one for the gpio-shutdown overlay), it will be also able to power up RPi from the halt state, effectively turning it into a fully-functional power button. However, the generic nature of systemd-logind should make the power buttons work on the other devices too, as long as their power buttons register themselves properly. It can be also done manually with the udev rules by tagging the input with the "power-switch" tag - such a custom rule can be easily deployed to the system with the [USB stick configuration import](https://github.com/home-assistant/operating-system/blob/dev/Documentation/configuration.md).

I've built it already and tested on my Raspberry Pi 4, as I [described on the forum](https://community.home-assistant.io/t/add-systemd-logind-to-the-home-assistant-os-to-handle-power-button-shutdown/215346/2?u=konpon96). It seems like adding systemd-logind didn't seem to affect the typical system behavior at all - except of the fact the power button is now handled and pressing it shuts down the system gracefully. :grinning: 

Here is how it looks after SSH-ing to the RPi - entering the root shell, executing `uname -a` and checking the service status with `systemctl status systemd-logind`:

Screenshot from 2020-07-31 03-00-36 (copy)


agners:
Hm, on one side I am wondering if systemd-logind really make sense on a OS which is not really meant to be used interactively. We pull in a lot of functionality not needed, just to get the power button handling.

I was also thinking this might be better handled by HA itself: E.g. an integration which can parse Linux input events from /dev/input directly, and then automation could be written to act accordingly.

On the other side, I must say it is a nice capability to shutdown gracefully without having to rely on any HA or any container... Also it seems that memory usage of systemd-logind is really rather minimal (1.0M according to your screenshot). I guess more often than not the user actually just wants to be able to gracefully reboot (which can be configured by setting `HandlePowerKey=reboot`, see https://www.freedesktop.org/software/systemd/man/logind.conf.html).

No strong opinion.

Konpon96:
That's what I've been wondering as well - if the inclusion of systemd-logind isn't bit too much for getting such a seemingly simple feature as handling the power buttons. However, it's probably the best way to go while following the convention of how systemd-based systems behave, without reinventing the wheel which would be writing something to handle the power button presses. As you said, my screenshot shows that logind doesn't seem to be particularly heavy during the runtime and size-wise it occupies 4.5% of the systemd library catalogue (library itself is 222K, the whole `/usr/lib/systemd` directory is 4.8M):

Screenshot from 2020-08-04 22-20-43

Even few excess kilobytes for its non-used functionalities wouldn't matter that much - HassOS is not supported nor recommended to be runned on the washing machine computers. :stuck_out_tongue: What's more, part of the things provided by the logind might not be active even if they happen to be compiled into the binary - for example, [according to the manpages](https://www.freedesktop.org/software/systemd/man/systemd-logind.service.html) it "provides polkit-based access for users for operations such as system shutdown or sleep" but HassOS doesn't even have polkit bundled, hence no risk of something like surprise password prompt on shutdown/reboot attempt by the HA (which I can confirm to be working):

Screenshot from 2020-08-04 21-46-29

As far as logind configuration is concerned, IMO it'd be better to leave the defaults - I can imagine change there being source of quite major bamboozle for people, with somebody wondering why is their freshly-installed device rebooting instead of powering off when the power button is pressed. :stuck_out_tongue: Once the logind is part of the system, it's a matter of changing the `HandlePowerKey` setting to suit one's needs.

pvizeli:
Let's test it

PR #802, opened 2020-08-03
Improves the Virtual Appliance section

aogriffiths:


Documentation/boards/README.md
```diff
@@ -25,14 +25,14 @@ The following boards/devices are supported:
 - Asus
   - Tinker Board
 - Virtual appliance (x86_64/UEFI):
-  - VMDK
-  - OVA ?
-  - VHDX ?
-  - VDI ?
-  - QCOW2 ?
+  - OVA - Open Virtual Appliance (open format, supports multiple virtual machines)
+  - VMDK - Virtual Machine Disk (open format, supports multiple virtual machines)
```
agners, line 29: Since those are just file formats, and the VM hypervisor needs to support it, I would write it passive, e.g. `(open format, supported by multiple virtual machine hypervisors)`

PR #808, opened 2020-08-05
Align all partitions to 1MiB boundary

agners:
Aligning partitions (and hence file system structures) to higher level
then 512 byte sectors is common practise and highly recommended for flash
backed block devices. It makes sure that the underlaying flash translation
layer (FTL) does not amplify writes due to missalignment of its erase
blcok size. Use a 1MiB boundary which is what a modern fdisk is doing.

Before this change:

```
# fdisk /dev/mmcblk0

Welcome to fdisk (util-linux 2.35.1).
Changes will remain in memory only, until you decide to write them.
Be careful before using the write command.

Command (m for help): p
Disk /dev/mmcblk0: 14.57 GiB, 15634268160 bytes, 30535680 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes
Disklabel type: dos
Disk identifier: 0x48617373

Device         Boot   Start      End  Sectors  Size Id Type
/dev/mmcblk0p1 *      16384    65537    49154   24M  c W95 FAT32 (LBA)
/dev/mmcblk0p2        65539  1228814  1163276  568M  5 Extended
/dev/mmcblk0p3      1228816  1425425   196610   96M 83 Linux
/dev/mmcblk0p4      1425427 30535679 29110253 13.9G 83 Linux
/dev/mmcblk0p5        65540   114693    49154   24M 83 Linux
/dev/mmcblk0p6       114695   638984   524290  256M 83 Linux
/dev/mmcblk0p7       638986   688139    49154   24M 83 Linux
/dev/mmcblk0p8       688141  1212430   524290  256M 83 Linux
/dev/mmcblk0p9      1212432  1228814    16383    8M 83 Linux
```

After this change:

```
# fdisk /dev/mmcblk0

Welcome to fdisk (util-linux 2.35.1).
Changes will remain in memory only, until you decide to write them.
Be careful before using the write command.

Command (m for help): p
Disk /dev/mmcblk0: 14.57 GiB, 15634268160 bytes, 30535680 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes
Disklabel type: dos
Disk identifier: 0x48617373

Device         Boot   Start      End  Sectors  Size Id Type
/dev/mmcblk0p1 *      16384    65535    49152   24M  c W95 FAT32 (LBA)
/dev/mmcblk0p2        65536  1239039  1173504  573M  5 Extended
/dev/mmcblk0p3      1241088  1437695   196608   96M 83 Linux
/dev/mmcblk0p4      1439744 30535679 29095936 13.9G 83 Linux
/dev/mmcblk0p5        67584   116735    49152   24M 83 Linux
/dev/mmcblk0p6       118784   643071   524288  256M 83 Linux
/dev/mmcblk0p7       645120   694271    49152   24M 83 Linux
/dev/mmcblk0p8       696320  1220607   524288  256M 83 Linux
/dev/mmcblk0p9      1222656  1239039    16384    8M 83 Linux
```

See also:
https://unix.stackexchange.com/questions/248939/how-to-achieve-optimal-alignment-for-emmc-partition
http://3gfp.com/wp/2014/07/formatting-sd-cards-for-speed-and-lifetime/

### DIFF
--- a/buildroot-external/scripts/hdd-image.sh
+++ b/buildroot-external/scripts/hdd-image.sh
@@ -201,27 +201,27 @@ function _create_disk_mbr() {
     local hdd_img="$(hassos_image_name img)"
     local hdd_count=${DISK_SIZE:-2}
     local disk_layout="${BINARIES_DIR}/disk.layout"
-    local boot_start=16384
+    local boot_start=$(size2sectors "8M")
 
-    local boot_size=$(($(size2sectors "$(get_boot_size)")+2))
-    local kernel0_size=$(($(size2sectors "$KERNEL_SIZE")+2))
-    local system0_size=$(($(size2sectors "$SYSTEM_SIZE")+2))
-    local kernel1_size=$(($(size2sectors "$KERNEL_SIZE")+2))
-    local system1_size=$(($(size2sectors "$SYSTEM_SIZE")+2))
-    local bootstate_size=$(($(size2sectors "$BOOTSTATE_SIZE")+2))
-    local overlay_size=$(($(size2sectors "$OVERLAY_SIZE")+2))
-    local data_size=$(($(size2sectors "$DATA_SIZE")+2))
-    local extended_size=$((kernel0_size+system0_size+kernel1_size+system1_size+bootstate_size+2))
+    local boot_size=$(size2sectors "$(get_boot_size)")
+    local kernel0_size=$(size2sectors "$KERNEL_SIZE")
+    local system0_size=$(size2sectors "$SYSTEM_SIZE")
+    local kernel1_size=$(size2sectors "$KERNEL_SIZE")
+    local system1_size=$(size2sectors "$SYSTEM_SIZE")
+    local bootstate_size=$(size2sectors "$BOOTSTATE_SIZE")
+    local overlay_size=$(size2sectors "$OVERLAY_SIZE")
+    local data_size=$(size2sectors "$DATA_SIZE")
+    local extended_size=$((kernel0_size+system0_size+kernel1_size+system1_size+bootstate_size+5*$(size2sectors "1M")))
 
     # we add one here for the extended header.
-    local extended_start=$((boot_start+boot_size+1))
-    local kernel0_start=$((extended_start+1))
-    local system0_start=$((kernel0_start+kernel0_size+1))
-    local kernel1_start=$((system0_start+system0_size+1))
-    local system1_start=$((kernel1_start+kernel1_size+1))
-    local bootstate_start=$((system1_start+system1_size+1))
-    local overlay_start=$((extended_start+extended_size+1))
-    local data_start=$((overlay_start+overlay_size+1))
+    local extended_start=$((boot_start+boot_size))
+    local kernel0_start=$((extended_start+$(size2sectors "1M")))
+    local system0_start=$((kernel0_start+kernel0_size+$(size2sectors "1M")))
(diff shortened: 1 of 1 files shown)


pvizeli:
What's about the GPT? Does that work with the OTA updates? If the SPL image size and handling now the same for the OTA update?

agners:
Yeah GPT would definitely be nicer, as it supports partition labels and does not require a extended boot records (EBR) between the partitions (and hence does not require those 1MiB gaps...). However, at least ODROID N2 boots exactly from the second block (LBA), which is where the primary header is stored... That will require us to write it in a non-standard location (`-j`, https://github.com/home-assistant/operating-system/blob/dev/buildroot-external/scripts/hdd-image.sh#L141). This however will make operating systems not find the partition table by default...

agners:
Did some testing with rauc: Created an image with the old format (5.2 dev build), and an image with the new format (5.3 dev build). I then upgraded from a 5.2 installation to 5.3, and vice versa, both cases worked and rauc.service did not show an issue.

The image sizes in the rauc bundle do stay the same (which make sense, since the images are based on `BOOT_SIZE`, `SYSTEM_SIZE` and `KERNEL_SIZE` from `buildroot-external/scripts/hdd-image.sh`, which are left untouched):

```                                   
/tmp/5.2:
total 143290
drwxr-xr-x  2 root root      118 27. Aug 23:51 .
drwxrwxrwt 22 root root      700 30. Aug 21:21 ..
-rw-r--r--  1 root root 25165824 27. Aug 23:51 boot.vfat
-rwxr-xr-x  1 root root     1192 27. Aug 23:51 hook
-rw-r--r--  1 root root 25165824 27. Aug 23:51 kernel.ext4
-rw-r--r--  1 root root      581 27. Aug 23:51 manifest.raucm
-rw-r--r--  1 root root 91021312 27. Aug 23:51 rootfs.img
-rw-r--r--  1 root root  8387584 27. Aug 23:51 spl.img

/tmp/5.3:
total 143290
drwxr-xr-x  2 root root      118 29. Aug 16:53 .
drwxrwxrwt 22 root root      700 30. Aug 21:21 ..
-rw-r--r--  1 root root 25165824 29. Aug 16:53 boot.vfat
-rwxr-xr-x  1 root root     1192 29. Aug 16:53 hook
-rw-r--r--  1 root root 25165824 29. Aug 16:53 kernel.ext4
-rw-r--r--  1 root root      581 29. Aug 16:53 manifest.raucm
-rw-r--r--  1 root root 91021312 29. Aug 16:53 rootfs.img
-rw-r--r--  1 root root  8387584 29. Aug 16:53 spl.img
```

The `kernel.ext4` and `boot.vfat` are exactly 24MBytes (25165824 bytes, 49152 blocks) with what they fit into the new layout. The `rootfs.img` image is dynamically sized and fits as well. The `spl.img` is stored between MBR and the first partition via hook/dd, and since the start of the first partition is still the same (16384 sectors/blocks), this shouldn't be a problem either.

So I think we can safely apply this patch to use the new layout on new installations. Of course this won't update the layout of old installations, but the raucb images do remain backwards/forward compatible.

pvizeli (approved):

schmurtzm:
Is it a good alignment recommended for vmdk too ?

PR #824, opened 2020-08-17
Load loop module before starting the container

agners:
This makes sure that the kernel module loop is loaded, the loop devices
under /dev have been created before the container starts. Docker uses
the current /dev as template for the container /dev. If the loop entries
are missing, loop devices can't be used inside the container. Use
losetup which does not make assumption weather loop support is built-in.

This fixes issues seen on my machine when entering the build environment
the first time after build:
  mount: /mnt/data: failed to setup loop device for /export/data.ext4.
  make[2]: *** [package/pkg-generic.mk:364: /build/buildroot/output_rpi4/build/hassio-1.0.0/.stamp_target_installed] Error 32


DaAwesomeP:
Aha! I have been having this mysterious issue for so long. Thank you!

Issue #733, opened 2020-06-08
Ver 4.10 RPI4 WiFi don't work anymore.


dianlight:
**HassOS release with the issue:**
- Operating System: 4.10

**Kernel logs:**
.. alot of log like:
```
[  556.305711] brcmfmac: brcmf_cfg80211_set_power_mgmt: power save enabled
[  558.917741] brcmfmac: brcmf_set_key_mgmt: invalid cipher group (1027074)
[  558.917766] brcmfmac: brcmf_cfg80211_connect: wl_set_key_mgmt failed (-22)
....
```

**Description of problem:**
After uprgading my RPi4 installation to ver 4.10 the wifi don't work anymore.
I also try with a new installation and the problem is the same. No Wifi connection. If I revert to a 4.9 backup with the same config all work.

This is my-network config:

```
[connection]
id=my-network
uuid=72111c67-4a5d-4d5c-925e-f8ee26efb3c3
type=802-11-wireless

[802-11-wireless]
mode=infrastructure
ssid=LTOfficenet
# Uncomment below if your SSID is not broadcasted
#hidden=true

[802-11-wireless-security]
auth-alg=open
key-mgmt=wpa-psk
psk=<private>

[ipv4]
method=auto

[ipv6]
addr-gen-mode=stable-privacy
method=auto
```

cogneato:
What is the hardware for this install?

dianlight:
RasperryPi 4 Model B 4GB.

da-anda:
missing kernel driver due to the kernel downgrade to 4.19 in #723 ?

LoermansA:
Can I revert back to 4.9 without a reinstall? I have this same issue. Plugging in an ethernet cable doesn't seem to work. I have no monitor connected to the pi so I cannot see the bootlog. So; no network, no display but I do hava access to the sdcard ofcourse. Can I downgrade?

LoermansA:
I managed to get the pi running on eth0 instead of wifi by altering the NetworkManager config in one of the overlay partitions (i think). So it is running again. Now to get wifi back online...

cogneato:
The network config file @dianlight is using as formatted does not work for me. Creating a file with the following format does work on my pi4 when imported from usb CONFIG stick (the name of the file does not matter, as long as it is in CONFIG/network/ and has no extension).

I had no issue connecting to wifi using nmcli commands.

```
[connection]
id=wovenomena
[...]

PR #830, opened 2020-08-22
RPi4 fix WiFi with 802.11r fast roaming enabled (#733)

agners:
Add brcmfmac patch which adds 802.11r fast roaming support.


corvy:
Just wondering, will this merge be part of the next release? Thanks for all the hard work! 👍

agners:
@corvy yes, the rel-4 tag means it will get back-ported to rel-4 and hence part of the next 4 release.

corvy:
Thank you for taking the time to clarify this @agners, much appreciated 👍

PR #831, opened 2020-08-24
Optimize overlay driver

pvizeli:
https://www.youtube.com/watch?v=fSyr_IXM21Y&feature=youtu.be

Not sure about the INDEX, but the other should be safe.


agners:
This refers to the overlayfs talk at Plumbers: https://linuxplumbersconf.org/event/7/contributions/639/attachments/501/969/Overlayfs-containers-lpc-2020.pdf

The speaker distinguish three use cases with different recommendations (Local instance, Migratable instance and Portable image). I guess during building we are in the "Migratable instance" use case, we want the overlayfs image mounted/created by Docker on the build host to be mountable by the device later on. On the device we are in the "Local instance" case, e.g. we typically do not want to move those containers (this is/would be the case when moving the whole file system from one device to the other, as suggested in !791. This case shows that just backing up data and moving to the new device gives more flexibility in terms of optimizing/changing the underlying system architecture).

So from what I can tell it should be safe to enable all three, but he recommends only enabling redirect_dir and metacopy, the index option can have performance impact and is only really required when NFS is used, so I think I would not enable that one.

buildroot-external/kernel/docker.config
```diff
@@ -71,6 +71,9 @@ CONFIG_EXT4_FS=y
 CONFIG_EXT4_FS_POSIX_ACL=y
 CONFIG_EXT4_FS_SECURITY=y
 CONFIG_OVERLAY_FS=y
+CONFIG_OVERLAY_FS_REDIRECT_DIR=y
+CONFIG_OVERLAY_FS_METACOPY=y
+CONFIG_OVERLAY_FS_INDEX=y
```
agners, line 76:
```suggestion
```

agners (changes requested):

agners (approved):

PR #834, opened 2020-08-26
Another OS rename catch

mikebaz:


pvizeli (commented):
Let's call it HassOS on this place because that's the internal project name which was never renamed, just rebrand.

Issue #650, opened 2020-05-01
Tinker Board S does not boot from eMMC with HassOS 4.5 release/update


mbo18:
**HassOS release with the issue:**

HassOS 4.5

**Supervisor logs:**

None, does not boot from eMMC. Nothing on screen (HDMI). Does boot from SD card

**Journal logs:**

None

**Kernel logs:**

None

**Description of problem:**
Just tried to upgrade from HassOS 3.13 to 4.5 from 'import from USB' and my Tinker Board S did not boot anymore. Tried to flash 4.5 directly to eMMC and it does not boot. Flash to SD card and it boot fine.
It seems that boot from eMMC is not working anymore with HassOS 4.5

venditdevs:
That would make sense  why my Odroid N2 isn't booting anymore from SD. Because the Odroid-N2 must select MMC to get it to boot Home Assistant from SD. I know: that doesn't make sense but it really needs  it ;).
So basically we can say: EMMC booting is broken in HassOS4.5.

Multipage:
That would make sense why my Odroid N2 isn't booting anymore from SD. Because the Odroid-N2 must select MMC to get it to boot Home Assistant from SD. I know: that doesn't make sense but it really needs it ;).
So basically we can say: EMMC booting is broken in HassOS4.5.

alex7614:
Same problem on Tinker S on eMMC. I flahed back to 4.4 on eMMC until new version.

mbo18:
It seems that it will not be solved as Pvizeli has no tinker board S to make tests and rbray89 who worked on it has leaved the project

mbo18:
@pvizeli Maybe you should block OS update for this specific board? because people will click on update and it won't boot anymore...
What is the best way now? SD card + offloading data to eMMC?

pvizeli:
@Multipage the N2 was just a bug in the official kernel from the manufacture and is fix with 4.6. The Tinker was just designed for usage on Tinker (without S) because for that is needs a new image. It's my wrong that I did not look exactly what Ryan did. But also on the official website, we just list the Tinker board without S (https://www.home-assistant.io/hassio/installation). However, Look like there are some people, maybe someone of you will join the project and spend some days per year into maintenance and responsible for issues.

cybergrimes:
What is this? I've had a Tinker Board S running Home Assistant since December 2018. We have had other non-booting OS image issues in the past for the S model that were fixed.

cybergrimes:
[...]

PR #845, opened 2020-09-03
Fix Tinker Board S (eMMC) boot (#650)

agners:


agners:
One think which is a bit odd with the board detection in the U-Boot script is that HAOS will boot with the Tinker Board device tree on a Tinker Board S when booting from SD. But it turns out to actually be an advantage: Linux won't see the eMMC, and hence it is guaranteed to mount the partitions from the SD card... :+1:

pvizeli (approved):

PR #847, opened 2020-09-04
Use Microsoft Basic Data instead of Microsoft Reserved Partition

agners:
Currently the Microsoft Reserved Partition GUID is used for this FAT32
formatted partition. This GUID is a rather Microsoft Windows specific
GUID and not commonly used on Linux.

On Linux systems partitions of this type do not get automatically
mounted (see /usr/lib/udev/rules.d/80-udisks2.rules). However, since
this partition contains some files user commonly need to adjust
(config.txt, cmdline.txt) it would be good if the partition does get
mounted.

Use Microsoft Basic Data instead, which is used by default for FAT32
partition (even by Linux partitioning tools such as gparted). Tested
on ASUS Tinker Board and RPi4.

### DIFF
--- a/buildroot-external/scripts/hdd-image.sh
+++ b/buildroot-external/scripts/hdd-image.sh
@@ -268,7 +268,7 @@ function _create_disk_mbr() {
 function _fix_disk_hyprid() {
     local hdd_img="$(hassos_image_name img)"
 
-    sgdisk -t 1:"E3C9E316-0B5C-4DB8-817D-F92DF00215AE" "${hdd_img}"
+    sgdisk -t 1:"EBD0A0A2-B9E5-4433-87C0-68B6B72699C7" "${hdd_img}"
     dd if="${BR2_EXTERNAL_HASSOS_PATH}/bootloader/mbr.img" of="${hdd_img}" conv=notrunc bs=512 count=1
 }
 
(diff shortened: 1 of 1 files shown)


pvizeli:
Did you test it also what is happening if you plug it into a Microsoft PC?

s00500:
Hey, just wondering, why was this actually reverted ?

agners:
@s00500 there are some information in the revert pull request #852. From what I can tell using Microsoft Reserved Partition is the wrong type for a embedded boards boot partition (since MRR really are meant for Windows use, see https://en.wikipedia.org/wiki/Microsoft_Reserved_Partition). But it seems to behave better on Windows machines. I plan to come up with a solution which works for Windows as well.

agners:
At least at one point it seems that RPi 4 booted only from EFI System Partiation or Basic Data Partition:
```
* Resolve: USB boot fails if the GPT contains no basic data or EFI partitions #130
```

 https://github.com/raspberrypi/rpi-eeprom/issues/130

s00500:
Ok I see, interesting, I could make a Pi4 boot without any issues with the changed ids, and also be be able to mount the boot partition on macOS and linux automatically (which is the whole point I guess ?)

And since also raspbian does behave like this I was curious why this image does not / did revert the change that made it behave like other raspberry images...

So that is why I was trying to learn what caused the revert here... can we maybe get another clarifying remark from @pvizeli ? I am really mainly trying to learn why... And as always: Thanks for all the awesome work on this system, its really, really nice built!!! =D

phiresky:
I just want to mention that I just tried booting HAOS from the current image (`haos_rpi4-64-9.5.img.xz`) in a Raspberry Pi 4 on a USB3 USB-Stick (with no microsd inserted), but it failed with the error "GPT contains no bootable partitions". HDMI screenshot:

image


The boot partition in the current image has type "Microsoft Reserved".

I changed the partition type manually to "Basic Data" and booting immediately started working.

PR #852, opened 2020-09-07
Revert "Use Microsoft Basic Data instead of Microsoft Reserved Partition"

pvizeli:
Reverts home-assistant/operating-system#847


pvizeli:
image

pvizeli:
Is now not marked as hyprid and use the hidden partition code which make it not visible anymore

PR #898, opened 2020-10-15
Bump ODROID boards to Linux 5.9

agners:
This makes quite some patches obsolete which since have been upstreamed.


pvizeli (dismissed):

pvizeli (commented):
I think we need to update the buildroot-patches too?

https://github.com/home-assistant/operating-system/blob/dev/buildroot-patches/0008-Linux-5.7.patch

agners:
> I think we need to update the buildroot-patches too?

Hm, I think those config symbols are just used to control packages which require a certain kernel version. But since there is no package in buildroot which requires a newer kernel than the one which was current at the point buildroot has been released, there is not much value in adding newer kernel config symbols.

I think we can use the `_LATEST` mechanism: https://patchwork.ozlabs.org/project/buildroot/patch/20200113180639.7344-1-yann.morin.1998@free.fr/

I'll give this a try.

agners:
Actually, it seems that our configuration use the headers from the kernel provided (via `BR2_KERNEL_HEADERS_AS_KERNEL=y`, which is getting set implicitly). When using a custom kernel, buildroot needs to know what version those headers are at least, so that it can hide packages which need something newer than that (e.g. `package/libgpiod` requires at least 4.8). However, introducing newer symbols without packages does not bring value really. Buildroot suggests to select the latest version if a newer kenel is used, see [package/linux-headers/Config.in.host](https://github.com/home-assistant/operating-system/blob/dev/buildroot/package/linux-headers/Config.in.host#L116)). So we can safely remove that patch and just use the latest version which was there at the time buildroot has been released, which is `BR2_PACKAGE_HOST_LINUX_HEADERS_CUSTOM_5_4` (the name of the symbol is confusing, I which they had used something like `...5_4_OR_NEWER`.

agners:
FWIW, when we update buildroot we should then use the closes matching config symbol there. But I think its better to change our machine configs then rather then backporting the symbols in buildroot today.

pvizeli (approved):

PR #899, opened 2020-10-15
Bump rpi-firmware to latest stable release (sep 2020)

tmm1:
https://github.com/raspberrypi/firmware/releases/tag/1.20200902


hdjweb:
So before update to 5.3 on rpi4 I must update firmware for boot? or I can only update with terminal hassio with the command ha os update --version 5.3 without update rpi4 firmware!?

tmm1:
I'm not quite sure what you mean, but this is not related to boot or kernel. The firmware manages things like the VideoCore and power-saving.

hdjweb:
I've updated hassios on 5.3 and now hassio doesn't go

hdjweb:
I want to say that before I must update boot on September 2020 and after I must update hassos on 5.3??

pvizeli (commented):
This work with our current 4.19 kernel & DTB?

tmm1:
Yes I tested with 4.19 and it works.

Jpsy:
Just for my understanding (and to keep my RPi4 community guide up-to-date):
Does this mean that HA OS now contains the Pi4 boot EEPROM firmware and will update it automatically at boot time just like Raspberry OS does?

tmm1:
No. This is the videocore firmware which is separate from the eeprom bootloader.

tmm1:
>Does this mean that HA OS now contains the Pi4 boot EEPROM firmware and will update it automatically at boot time just like Raspberry OS does?

If this is something you think the community will find useful, I can send a PR to add it.

It will still require an initial flash to put the new eeprom with self update in place. Also self-update only works from USB as it was not reliable via SD card.

Jpsy:
> If this is something you think the community will find useful, I can send a PR to add it.

I think this would be very helpful in the current situation where many Pi 4 in the field still come with the old firmware. From the long part of the guide that handles this update you can tell that it would help people a lot and make HA the easy to use system on Pi 4 that it was on Pi 3.
But on the other hand it adds another fragment of code to the OS that must be maintained and kept current.
I can't tell the amount of long-term work this new code would generate and I don't want to add unnecessarily to the load of the devs.
 
BTW: Thanks for your related comment on the guide. I will add your information into the guide's text soon.

Jpsy:
@tmm1 is the videocore firmware transient or is it flashed permanently into the Pi?
I ask because an increased number of community members are reporting strange problems after updating to OS 5.4 (see the guide link above). It looks like once they have 5.4 installed they cannot boot from SSD anymore - even if they revert to 5.3 or 5.2. I am running 5.4 without any problems, so I cannot reproduce their problems.

tmm1:
Firmware is loaded at boot from the boot partition

PR #901, opened 2020-10-15
Remove not needed partition magic

pvizeli:


Documentation/partition.md
```diff
@@ -51,15 +51,7 @@ Log in as `root` to get to the Home Assistant CLI and then enter `login` to cont
...
-It is recommended to use fdisk to remove the existing partition(s) before proceeding.
-
-- Type `fdisk /dev/XXX` (replacing XXX with your drive)
-- Type `d` to delete a partition.
-- Continue if needed, then write the changes.
-
-Creating a new partition is not necessary.
-
-With the drive now prepared, use the below command (again, replacing XXX with your drive)
+With the drive, use the below command (again, replacing XXX with your drive)
```
agners, line 54: Deleting partition is not necessary, unless there is a partition named `hassos-data`. Since I got into trouble after not being patient enough as well, i think its worth mentioning:

```suggestion
Make sure the drive has no partition named `hassos-data` (or no partition at all). With the drive, use the below command (again, replacing XXX with your drive)
```

Issue #712, opened 2020-06-03
Odroid N2 USB crash


jrhodk:
**HassOS release with the issue:**

Release 4.9 running on Odroid N2 on eMMC
It freezes after some time with a screen full of this message: 
[  216.799302]   fe.dai-link-0:    ASoC:    No backend DAI enabled for fe.dai-link-0
where the number in the start counts up..

**Supervisor logs:**

I cant run anything because the system is frozen

**Journal logs:**

**Kernel logs:**

**Description of problem:**
It freezes after some time with a screen full of this message: 
[  216.799302]   fe.dai-link-0:    ASoC:    No backend DAI enabled for fe.dai-link-0

I have had this problem since v. 4.5. This time i did start with fresh install with only the samba plugin and the deConz zigbee driver installed.

I have 2 Ordoid N2 with eMMC, they both crash with the fe.dai-link-0 message,

balloob:
1. How did you flash the eMMC?
2. Do you have any USB sticks inserted?
3. What kind of network do you have attached?

jrhodk:
1. I did the flashing with the Hardkernel eMMC flasher: https://www.hardkernel.com/shop/emmc-module-reader-board-for-os-upgrade/. 
2. There are only the deConz zigbee USB stick connected.
3. The Odroid N2 are connected directly to my networkwith ethernet cable.

jrhodk:
I get the ui, and it works fine until the system decides to freeze. Which eMMC drive do you use and which flasher?

balloob:
I use https://www.hardkernel.com/shop/usb3-0-emmc-module-writer/ with [Etcher](https://www.balena.io/etcher/)

jrhodk:
And your eMMC module?

balloob:
16GB from ODroid website.

balloob:
I checked with some people. 

```
 [ 216.799302] fe.dai-link-0: ASoC: No backend DAI enabled for fe.dai-link-0
```

This message is about the DAC, which is not supported. That should not be the cause for freezing.

jrhodk:
![Uploading 83820902...]()

jrhodk:
I caught some kernelpanick on the screen after a crash! - Hope this gives a hint anout whats wrong!

jrhodk:
What could have corrupted the kernel stack? Is it a buffer overflow error, or a memory hardware error og something else?

balloob:
I don't know 🤷

[...]

PR #905, opened 2020-10-17
Fix CDC ACM error recovery path (#712)

agners:
Revert CDC ACM cool-down patch. This should fix the error recovery paths
in the CDC ACM driver and allow CDC ACM devices to continue working even
in the event of USB issues.


agners:
@pvizeli the patch applies to 5.9 as well as 5.7, so we can merge this before or after #898.

pvizeli (approved):

PR #935, opened 2020-10-27
make it easier to edit and transfer files in the docker builder

tmm1:
This makes my life easier when I sometimes need to `scripts/enter.sh` and edit or transfer files.

Feel free to close if you don't think this is useful.


agners:
`vim` I definitely welcome as well, ssh I wouldn't have missed so far, usually use the bind mounted directory for that :-)

agners (approved):

PR #936, opened 2020-10-27
use GPT partition table for rpi4

tmm1:
The RPI4 eeprom bootloader can read directly from GPT, so hybrid MBR is not necessary.


pvizeli (approved):

pvizeli:
Can you add this also to rpi4-64?

agners:
Actually, I tested that on 64-bit and it seems to break boot (I get the rainbow screen, no U-Boot on serial port).

tmm1:
Hmm I have been using this with rpi64 and it was working okay.

Maybe it's related to the eeprom version? Or maybe it only works for USB drives.

agners:
> Maybe it's related to the eeprom version? Or maybe it only works for USB drives.

EEPROM is the latest 2020-09-03. I tested using SD card, so that could be the difference...

pvizeli:
> Or maybe it only works for USB drives.

Look like. However, the SD card is the focus for us, since booting from a hard disk with an SoC is maybe working but hacky at all.

tmm1:
Hm I tried with SD and it also works. I must have done something wrong in this PR so I will check it again.

Can you share the `sgdisk -Op` from your generated image?

agners:
> Can you share the sgdisk -Op from your generated image?

Cleared that SD already again, would have to recreate.

FWIW, its a RPi 8GB I tested it here. I just added `BOOT_SYS=efi`, and made sure boot partition and all really get rebuilt by doing a clean build.

pvizeli:
> Actually, I tested that on 64-bit and it seems to break boot (I get the rainbow screen, no U-Boot on serial port).

So this doesn't work on the SD card, right?

agners:
Retested this with the new firmware, and it seems to work well on a RPi4 8GB.

PR #937, opened 2020-10-27
reboot kernels on panic so u-boot can fallback to previous update

tmm1:
This makes the system recover automatically by rebooting three times after bad kernel update until uboot switches back to the old kernel.


buildroot-external/board/raspberrypi/kernel.config
```diff
@@ -1,2 +1,3 @@
 # CONFIG_AUTOFS4_FS is not set
 # CONFIG_AUTOFS_FS is not set
+CONFIG_PANIC_TIMEOUT=5
```
pvizeli, line 3: That would be better at https://github.com/home-assistant/operating-system/blob/dev/buildroot-external/kernel/hassos.config

tmm1: Thanks. Moved to central config.

pvizeli:
LGTM / @agners ?

PR #938, opened 2020-10-27
automatically fsck to repair issues after an unclean shutdown

tmm1:
I have run into situations where ext4 fs is badly damaged and fails to mount, which essentially kills the entire OS and requires manual intervention to repair

With this PR, fsck is run automatically using `systemd-fsck`.


buildroot-external/board/raspberrypi/uboot-boot64.ush
```diff
@@ -18,7 +18,7 @@ test -n "${BOOT_A_LEFT}" || setenv BOOT_A_LEFT 3
 test -n "${BOOT_B_LEFT}" || setenv BOOT_B_LEFT 3
 
 # HassOS bootargs
-setenv bootargs_hassos "zram.enabled=1 zram.num_devices=3 apparmor=1 security=apparmor rootwait systemd.machine_id=${MACHINE_ID} cgroup_enable=memory"
+setenv bootargs_hassos "zram.enabled=1 zram.num_devices=3 apparmor=1 security=apparmor rootwait systemd.machine_id=${MACHINE_ID} cgroup_enable=memory fsck.repair=yes"
```
pvizeli, line 21: We should do that on all platforms. Default is `preen` which is safer to run in theories, but maybe it helps on some issues which I saw on Sentry.

agners:
Ideally we should understand why those corruption happened. ext4 is journalling, and should (in absense of hardware issue) recover always! Given that you are adding fsck for FAT, I assume it was the FAT partition which caused troubles? Maybe we can mount it read-only by default?

pvizeli:
read-only would be suboptimal. I never saw an issue with boot partition in the last 3 years. We can not restore boot data with snapshot (the same kind of device, otherwise we do not touch that)

pvizeli:
But I agree, ext4 fs is badly damaged for data partition is sometimes happening. And I see also the only issues on RPi platform on sentry with corrupt ext4. Like the docker overlay bug which is only happening on RPi kernel (too many symbolic links)

tmm1:
I saw corruption only on ext4. The fat partition was marked dirty but it did not cause any problems. With fsck.fat the dirty bit was cleared.

pvizeli:
I personal fine with that, we should just handle it all in the same way

buildroot-external/board/raspberrypi/uboot-boot64.ush
tmm1: Is there a central place to add bootargs for all platforms?

tmm1: Or we can start with the default of preen for now. I think the bigger change is calling systemd-fsck on each mount, and even preen should fix most issues.

pvizeli: There is no central place, you need but it on each uboot script, but that is not much place. `preen` is already running in the background after mounting. So it would make sense to go over all uboot scripts or search for `bootargs_hassos` and but it in. 30s with vscode :)

tmm1: >preen is already running in the background after mounting.

Are you sure about this? I thought fsck is not being run at all right now because we don't have any systemd-fsck dependencies setup. By default only systemd-fsck-root runs.

agners:
> But I agree, ext4 fs is badly damaged for data partition is sometimes happening. And I see also these issues only on the RPi platform on sentry. Like the docker overlay bug which is only happening on RPi kernel (too many symbolic links). However, I have also reports (not sentry) on OVA which damaged ext4.

Hm, do we know which VM solution is causing issues with OVA image? Would be interesting.

In any case, it shows that this is not a ext4 issue, but a hardware issue typically. And my fear is that by just force `fsck -y` we hide problems which down the road just bite us anyway. I think we should make at least sure that we have something in the logs which shows if such issues have been fixed.

pvizeli:
We should be able to see it on journald?

buildroot-external/board/raspberrypi/uboot-boot64.ush
pvizeli: >https://www.freedesktop.org/software/systemd/man/systemd-fsck@.service.html#Kernel%20Command%20Line

Like I understand is the mode auto with repair type preen the default value if you don't set one.

tmm1: Yes auto and preen are defaults, but they only kick in for entries inside fstab:

>These services are started at boot if passno in /etc/fstab for the file system is set to a value greater than zero.

Since our fstab is empty, systemd-fsck is never run so it never looks at the preen option or runs fsck: https://github.com/systemd/systemd/blob/master/src/fsck/fsck.c#L96-L97

Only with the systemd dependencies added in the rest of this PR will preen mode run. So that's why I'm wondering maybe we only keep it as preen for now and if there are more problems in the future we can switch it to forced yes.

buildroot-external/rootfs-overlay/usr/lib/systemd/system/mnt-data.mount
```diff
@@ -5,6 +5,8 @@ DefaultDependencies=no
 After=hassos-expand.service
 Before=umount.target local-fs.target
 Conflicts=umount.target
+After=systemd-fsck@dev-disk-by\x2dlabel-hassos\x2ddata.service
```
tmm1, line 8: Is it allowed to have multiple `After=` lines? There is already one above.

agners: Hm, description is not very clear, but I think so? https://www.freedesktop.org/software/systemd/man/systemd.syntax.html#

tmm1: Yea it seems it is fine to have multiple.

buildroot-external/board/raspberrypi/uboot-boot64.ush
tmm1: BTW, `buildroot/board/hardkernel/odroidxu4/boot.ini` also has `fsck.repair=yes`. Not sure why.

tmm1:
I updated this PR to:

- install dosfstools on all rpi targets
- remove fsck.repair=yes so preen is default

I will open a separate PR to add fsck.repair=yes and we can wait to merge that one separately.

buildroot-external/rootfs-overlay/usr/lib/systemd/system/mnt-data.mount
pvizeli: both work

pvizeli (approved):

buildroot-external/rootfs-overlay/usr/lib/systemd/system/mnt-boot.mount
```diff
@@ -3,6 +3,8 @@ Description=HassOS boot partition
 DefaultDependencies=no
 Before=umount.target local-fs.target
 Conflicts=umount.target
+After=systemd-fsck@dev-disk-by\x2dlabel-hassos\x2dboot.service
```
pvizeli, line 6:
```suggestion
Wants=systemd-fsck@dev-disk-by\x2dlabel-hassos\x2dboot.service
```

buildroot-external/rootfs-overlay/usr/lib/systemd/system/mnt-data.mount
pvizeli, line 8:
```suggestion
Wants=systemd-fsck@dev-disk-by\x2dlabel-hassos\x2ddata.service
```

buildroot-external/rootfs-overlay/usr/lib/systemd/system/mnt-overlay.mount
```diff
@@ -3,6 +3,8 @@ Description=HassOS overlay partition
 DefaultDependencies=no
 Before=umount.target local-fs.target
 Conflicts=umount.target
+After=systemd-fsck@dev-disk-by\x2dlabel-hassos\x2doverlay.service
```
pvizeli, line 6:
```suggestion
Wants=systemd-fsck@dev-disk-by\x2dlabel-hassos\x2doverlay.service
```

pvizeli (changes requested):
Let's use Wants to enforce it. Otherwise, it can fail pretty fast very hard on an issue. After that, we can merge it. It allows recovery on the issue but also doesn't stop to boot if fsck fails.

buildroot-external/rootfs-overlay/usr/lib/systemd/system/mnt-boot.mount
```diff
@@ -3,6 +3,8 @@ Description=HassOS boot partition
...
+Requires=systemd-fsck@dev-disk-by\x2dlabel-hassos\x2dboot.service
```
pvizeli, line 7:
```suggestion
Wants=systemd-fsck@dev-disk-by\x2dlabel-hassos\x2dboot.service
```

buildroot-external/rootfs-overlay/usr/lib/systemd/system/mnt-data.mount
```diff
@@ -5,6 +5,8 @@ DefaultDependencies=no
...
+Requires=systemd-fsck@dev-disk-by\x2dlabel-hassos\x2ddata.service
```
pvizeli, line 9:
```suggestion
Wants=systemd-fsck@dev-disk-by\x2dlabel-hassos\x2ddata.service
```

buildroot-external/rootfs-overlay/usr/lib/systemd/system/mnt-overlay.mount
```diff
@@ -3,6 +3,8 @@ Description=HassOS overlay partition
...
+Requires=systemd-fsck@dev-disk-by\x2dlabel-hassos\x2doverlay.service
```
pvizeli, line 7:
```suggestion
Wants=systemd-fsck@dev-disk-by\x2dlabel-hassos\x2doverlay.service
```

pvizeli (changes requested):

tmm1:
> Let's use Wants to enforce it. Otherwise, it can fail pretty fast very hard on an issue.

Good catch, that would have been bad if fsck fails. Updated and ready to merge.

pvizeli (changes requested):
Please add the DOSFSTOOLS to all our board defconfig, all images should look most as possible in the same.

pvizeli:
I found also some systemd services which need also that:

Data:
- https://github.com/home-assistant/operating-system/blob/dev/buildroot-external/rootfs-overlay/usr/lib/systemd/system/hassos-expand.service
- https://github.com/home-assistant/operating-system/blob/dev/buildroot-external/rootfs-overlay/usr/lib/systemd/system/hassos-data.service

Otherwise, it's possible to run an fsck while we want to manipulate the partition/disk.

tmm1:
> Please add the DOSFSTOOLS to all our board defconfig, all images should look most as possible in the same.

done!



> I found also some systemd services which need also that:

I added as `After=` to hassos-expand and hassos-data service.

pvizeli (approved):

PR #939, opened 2020-10-27
Add rpi-eeprom firmware

tmm1:
This will allow us to take advantage of the RPI4 self-update feature to push new stable eeprom bootloader firmwares as part of HassOS updates

cc https://github.com/home-assistant/operating-system/pull/899#issuecomment-714890960 @jpsy


pvizeli (approved):
Nice work

buildroot-external/configs/rpi4_64_defconfig
```diff
@@ -42,6 +42,7 @@ BR2_PACKAGE_SQUASHFS=y
 BR2_PACKAGE_RPI_FIRMWARE=y
 BR2_PACKAGE_RPI_FIRMWARE_VARIANT_PI4=y
 BR2_PACKAGE_RPI_WIFI_FIRMWARE=y
+BR2_PACKAGE_RPI_EEPROM=y
```
pvizeli, line 45: rpi4?

pvizeli: I think we can this add also to the rpi4 (32bit) defconfig

agners:
Seems to work here, with `enable_uart=1` I can see that the EEPROM is being updated:

```
Read config.txt bytes     1826 hnd 0x00000780 hash '1850978bb5bda2bb'
Read pieeprom.upd bytes   524288 hnd 0x00000136 hash '5159f01d9c15fe32'
Reading EEPROM: 524288
Writing EEPROM
+++++++++++++++++++++++++.......++++++++++++++++++++++++++++++++++++++++++++++++++..............................................
Verify BOOT EEPROM
Reading EEPROM: 524288

```

pvizeli:
On the next OTA update, the file is present again, do that flash it again?

agners:
The files are still there after the update, so I assume it detects if its newer? I don't have diagnostic output on UART anymore since the update also sets the config back...

pvizeli:
Look like, so we have still open the rpi4_devconfig before we can merge it

agners:
I tried to make it update again, but it just wouldn't... I am now on 2020-09-03, and it seems that this version does not automatically update from SD card. Quote from https://www.raspberrypi.org/documentation/hardware/raspberrypi/bcm2711_bootloader_config.md:

> Self-update is not enabled in SD boot; the ROM can already load recovery.bin from the SD card.

tmm1:
>The files are still there after the update, so I assume it detects if its newer?

Yes it checks and only applies the .upd file is the current firmware does not match. It will show this info over HDMI as well telling you if update was skipped.

>I don't have diagnostic output on UART anymore since the update also sets the config back...

This part is annoying for development. If you have a custom eeprom with NETCONSOLE or UART it gets overwritten. Maybe we can have a special dev file inside the boot partition and if it exists the eeprom file is removed or renamed during rauc update?

>Self-update is not enabled in SD boot; the ROM can already load recovery.bin from the SD card.

Correct, there was some big problem with self-update from SD card: https://github.com/raspberrypi/rpi-eeprom/pull/209

So this technique is mostly useful for USB boot scenarios. In this PR I customized USB_PWR_OFF_TIME for the eeprom to fix issues with some types of drives: https://github.com/raspberrypi/rpi-eeprom/issues/241

buildroot-external/package/rpi-eeprom/rpi-eeprom.mk
```diff
@@ -0,0 +1,25 @@
...
+RPI_EEPROM_VERSION = f24abcd305ab844dfba2ac92e52883056786ccab
+RPI_EEPROM_SITE = $(call github,raspberrypi,rpi-eeprom,$(RPI_EEPROM_VERSION))
+RPI_EEPROM_LICENSE = BSD-3-Clause
+RPI_EEPROM_LICENSE_FILES = LICENSE
+RPI_EEPROM_INSTALL_IMAGES = YES
+RPI_EEPROM_FIRMWARE_PATH = firmware/stable/pieeprom-2020-09-03.bin
+
+define RPI_EEPROM_BUILD_CMDS
+	$(@D)/rpi-eeprom-config $(@D)/$(RPI_EEPROM_FIRMWARE_PATH) --out $(@D)/default.conf
+	(cat $(@D)/default.conf | grep -v ^$$; echo USB_MSD_PWR_OFF_TIME=1500) > $(@D)/boot.conf
```
tmm1, line 15: Here I changed USB_MSD_PWR_OFF_TIME from 1000 to 1500 as workaround for some drives not booting after soft-reboot.

Soon there will be a new beta firmware with a better fix, so we can use this same technique to update to that as part of HassOS update.

agners:
> > The files are still there after the update, so I assume it detects if its newer?
> 
> Yes it checks and only applies the .upd file is the current firmware does not match. It will show this info over HDMI as well telling you if update was skipped.

Not in the SD card case, it seems.

my_photo-20


> > I don't have diagnostic output on UART anymore since the update also sets the config back...
> 
> This part is annoying for development. If you have a custom eeprom with NETCONSOLE or UART it gets overwritten. Maybe we can have a special dev file inside the boot partition and if it exists the eeprom file is removed or renamed during rauc update?
> 

I tried to just rebuild the upd file manually and flash it. I think its good enough, I hope we don't have to do this very often :smile: 

> > Self-update is not enabled in SD boot; the ROM can already load recovery.bin from the SD card.
> 
> Correct, there was some big problem with self-update from SD card: [raspberrypi/rpi-eeprom#209](https://github.com/raspberrypi/rpi-eeprom/pull/209)
> 
> So this technique is mostly useful for USB boot scenarios. In this PR I customized USB_PWR_OFF_TIME for the eeprom to fix issues with some types of drives: [raspberrypi/rpi-eeprom#241](https://github.com/raspberrypi/rpi-eeprom/issues/241)

Hm, yeah that is a bit annoying. I think most people still run SD-card boot. Can we update via recovery.bin?
 I tried that, but I am not sure where to get recovery.bin from... Is that just the same file with different name? That I tried, it did not pick it up.

pvizeli:
> https://github.com/home-assistant/operating-system/blob/dev/buildroot-external/ota/rauc-hook

This file allows customizing OTA updates with the boot folder

pvizeli:
> This part is annoying for development. If you have a custom eeprom with NETCONSOLE or UART it gets overwritten. Maybe we can have a special dev file inside the boot partition and if it exists the eeprom file is removed or renamed during rauc update?

We should not allow us to customize it. It's already hard to maintain, if every device has x variation to run, it's killing us or bite later in our a**.

tmm1:
>I think most people still run SD-card boot. Can we update via recovery.bin?

Yes I think so. The recovery.bin is a special firmware, but once it loads it will look for pieeprom.upd and flash it, then rename itself to recovery.000

You can find it here: https://github.com/raspberrypi/rpi-eeprom/blob/master/firmware/stable/recovery.bin

According to https://www.raspberrypi.org/documentation/hardware/raspberrypi/booteeprom.md, the recovery.bin can only be loaded from SD and is ignored on USB boot.

So I think we simply need to place the recovery.bin into boot and then it will pick up and flash the pieeprom.upd on the next boot.

agners:
> You can find it here: https://github.com/raspberrypi/rpi-eeprom/blob/master/firmware/stable/recovery.bin

This does not work for me either. Leads to a red screen :confused:

McGiverGim:
Maybe a silly idea: if we are going to update automatically the eeprom (I think is a great idea to let USB boot for example), maybe we need to add the command to let the user know the eeprom installed? Or show it in the hardware info under Home Assistant UI?
Maybe can be useful if something is going wrong.

buildroot-external/configs/rpi4_64_defconfig
tmm1: Added to 32-bit. Left as RPI_EEPROM since I'm sure eventually rpi5 will come out and also use this.

agners:
I am generally ok merging this, however currently it uses a beta firmware. I rather prefer not shipping a beta firmware. Let's wait until the firmware gets promoted to stable.

Have there been issues with 2020-09-03?

tmm1:
With the 2020-09 stable firmware, many drives get stuck inside the bootloader after a soft reboot.

https://github.com/raspberrypi/rpi-eeprom/issues/241

agners:
Interesting thread. Thinking a bit more about this I think we should add the update as is: Only people with USB boot will get the update applied. And for those this new beta firmware seems to really bring improvements. And since your tests were successful too, let's give it a go.

Jpsy:
Would this mean that starting with 5.6 there will be no need to update the RPI boot firmware through Raspberry Pi Imager before installing HA on an RPi 4 with SSD attached through USB?

Or will it still be necessary to update with the Imager at least once to get HA running for the first time. And from there on HA would install further boot EEPROM updates?

I would update [the guide](https://community.home-assistant.io/t/installing-home-assistant-on-a-rpi-4b-8gb-with-ssd-boot/230948)  to reflect your answer.

agners:
> Would this mean that starting with 5.6 there will be no more need to update the RPI boot firmware through Raspberry Pi Imager before installing HA on an RPi 4 with SSD attached through USB?

There is a chicken egg issue here: The update can only happen if the current firmware supports that feature/is capable to actually talk to that USB SSD drive attached. This is not the case with older firmwares. So depending on what is on the Pi now, a manual upgrade is required first still...

Issue #928, opened 2020-10-23
With HassOS 5.4 Raspberry 3 does not boot anymore


TheZoker:
**HassOS release with the issue:** 5.4

**Supervisor logs:** None

**Journal logs:** None

**Kernel logs:** None

**Description of problem:**
When I upgraded from HassOS 5.3 to 5.4 today, my Pi 3B+ does not boot anymore. When I connect a HDMI cable, the only output I get is the colored image. I reflashed the SD card with 5.4 again to see if it was a upgrade error, but still same issue. I then flashed version 5.3 and the Pi booted normal again. Unfortunately I can't attach any logs, since there where just no logs anywhere. So I assume that something in 5.4 broke something for the Pi3 (maybe other Pi versions, but I only have version 3 to test here)

For reference, this is what I mean with a "colored image":
![image](https://user-images.githubusercontent.com/1368405/97114405-9accb080-16f0-11eb-962c-43a0a94a82e5.png)

archerne:
Same issue with the 64bit Rasp Pi 4 8GB

archerne:
I did a fresh install of 5.4 and it started (still testing to make sure everything works) but looks like the upgrade from 5.3 to 5.4 breaks something

SmarterHomeLife:
Seems to break something with the SD card, I upgraded my system that runs on an SSD with no issues but upgrading or a fresh install on an SD card fails to boot for me.

SeanPM5:
Can confirm, my secondary instance on an RPi4 (64bit / 4GB) was bricked after updating from 5.3 to 5.4.  It can no longer boot off the SD card.

fabiocastagnino:
Same issue with the 64bit Rasp Pi 4 8GB, upgrading from 5.2 to 5.4 or clean install. It can no longer boot off the SD card.

Gadgit83:
Same for me - RPI 4 8GB. 64bit image. Rolled back to 4.15 and worked fine. 5.3 appears to be booting OK.

Thought it was me - first time I'm trying the Pi with HA. Tried everything, different USB cables, different power supply, reflashed card, used balenEtcher and the Pi Imager, no luck. Should have known better than to rely on a dev version :).

cogneato:
I wasn't sure what "colored image" was referring to, but now I see.

TheZoker:
Thanks @cogneato for the hint, I placed an example image in the issue description 🙂

selmaohneh:
[...]

PR #942, opened 2020-10-27
Revert "bump rpi-firmware to latest stable release (sep 2020) (#899)"…

agners:
… (#928)

This reverts commit dade3adf801e7ae304e449232165bda990909b0b.
The new firmware lead to black screens and rainbow screen on various
Raspberry Pis.


pvizeli:
Our kernel is on 4.19 and this is the latest stable Firmware for the 4.19 Kernel. We decide to update the firmware after we migrated to 5.4 if there is no real reason or somethings we need a fix.

pvizeli (approved):

PR #949, opened 2020-11-02
Add machine specific landingpage container at build time

agners:
The landingpage container is a minimal webserver with built-in zeroconf
annoucement. Preinstall the machine specific landingpage container to
make sure it will show up right after startup.


pvizeli (approved):

pvizeli:
I need at least the supervisor v2020.10.0

PR #952, opened 2020-11-03
Simplify self healing capabilities of Supervisor service

agners:
Instead of relying on time based information on how long the container
has been running use a startup marker file to infer if the last startup
has been successful.

### DIFF
--- a/buildroot-external/rootfs-overlay/usr/sbin/hassos-supervisor
+++ b/buildroot-external/rootfs-overlay/usr/sbin/hassos-supervisor
@@ -67,7 +67,7 @@ if [ -z "${SUPERVISOR_CONTAINER_ID}" ]; then
         --privileged --security-opt apparmor="hassio-supervisor" \
         -v /var/run/docker.sock:/var/run/docker.sock \
         -v /var/run/dbus:/var/run/dbus \
-        -v /run/supervisor:/run/supervisor \
+        -v /run/supervisor:/run/os:rw \
         -v /etc/machine-id:/etc/machine-id:ro \
         -v ${SUPERVISOR_DATA}:/data:rw \
         -v /mnt/overlay:/os/overlay:rw \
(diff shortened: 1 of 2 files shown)


agners:
Related to: https://github.com/home-assistant/supervisor/pull/2214

buildroot-external/rootfs-overlay/usr/sbin/hassos-supervisor
```diff
@@ -78,6 +67,7 @@ if [ -z "${SUPERVISOR_CONTAINER_ID}" ]; then
         --privileged --security-opt apparmor="hassio-supervisor" \
         -v /var/run/docker.sock:/var/run/docker.sock \
         -v /var/run/dbus:/var/run/dbus \
+        -v /run/supervisor:/run/supervisor \
```
pvizeli, line 70:
```suggestion
        -v /run/supervisor:/run/os:rw \
```

buildroot-external/rootfs-overlay/usr/sbin/hassos-supervisor
```diff
@@ -78,6 +67,7 @@ if [ -z "${SUPERVISOR_CONTAINER_ID}" ]; then
         --privileged --security-opt apparmor="hassio-supervisor" \
         -v /var/run/docker.sock:/var/run/docker.sock \
         -v /var/run/dbus:/var/run/dbus \
+        -v /run/supervisor:/run/os:rw \
```
agners, line 70: @pvizeli  should we maybe rename that to os to avoid confusion? I kinda liked supervisor since then its clear its reserved for the supervisor, but since it is used inside the supervisor container already we cannot use that path inside the container.

pvizeli (approved):
We should future not delete the image, maybe first try to recreate the container. We can write somethings into startup-maker to detect the right action on the next strip execution

PR #954, opened 2020-11-03
RaspberryPi: Update firmware 2ba11f2a07760588546821aed578010252c9ecb3

agners:
This bumps to the stable firmware released on 2020-10-22. I tested the firmware on a Raspberry Pi 4 8GB and Raspberry Pi 3 Model B+, it seems to boot and reboot fine. This firmware is required for #936.


tmm1 (approved):
Looks good to me. Diff matches exactly what I have been using myself for testing during the past week without any problems.

pvizeli (approved):

PR #956, opened 2020-11-04
Bump Raspberry Pis to 5.4 kernel

agners:


agners:
So far I tested the 32-bit image on a Raspberry Pi 3 Model B+ and the 64-bit image on a Raspberry Pi 4 Model B 8GB.

pvizeli (approved):
Looks fine to me. Do the uart mapping working to GPIO?

tmm1:
Tested on Pi4 2GB w/ USB boot and works fine. I will keep it on my dev unit to stress test more over the next few days.

Last time I was playing with 5.4 kernel, once after a reboot I had a problem with ethernet not coming up. I found a related issue which is worth tracking: https://github.com/raspberrypi/linux/issues/3108

pvizeli:
I have the same issue sometimes with XU4

agners:
> Looks fine to me. Do the uart mapping working to GPIO?

With the last update enabling UART2 through 5 works, with or without `enable_uart=1`.

bschatzow:
I am having issues with the 64 bit kernel causing Home Assistant to freeze.  I see anywhere from a couple of hours to a couple of days maximum before I need to power off the Pi4 . Running on SSD drive.  Is there some details I can add to try and see what the issue is?  The system is very stable on all OS releases prior to OS5.5 when this change was implemented,

Issue #940, opened 2020-10-27
Raspberry Pi Bluetooth Crashing


Veldkornet:
**HassOS release with the issue:**

RaspberryPi 4 (4GB) 64-bit:
| Item  | Value |
| ------------- | ------------- |
| arch | aarch64 |
| chassis | embedded |
| dev | false |
| docker | true |
| docker_version | 19.03.11 |
| hassio | true |
| host_os | HassOS 4.15 |
| installation_type | Home Assistant OS |
| os_name | Linux |
| os_version | 4.19.127-v8 |
| python_version | 3.8.5 |
| supervisor | 249 |
| timezone | Europe/Amsterdam |
| version | 0.116.4 |
| virtualen | false |

**Description of problem:**
It seems that after a period of time of usage, the Bluetooth becomes unstable and crashes. This is noticable when using the room-assistant add-on. 
```
10/25/2020, 11:23:39 AM - error - BluetoothClassicService: Command failed: hcitool -i hci0 cc "64:a2:f9:f1:b8:7b" && hcitool -i hci0 rssi "64:a2:f9:f1:b8:7b"
Invalid device: Network is down
10/25/2020, 11:23:45 AM - error - BluetoothClassicService: Command failed: hcitool -i hci0 cc "88:63:DF:8A:1B:DC" && hcitool -i hci0 rssi "88:63:DF:8A:1B:DC"
Invalid device: Network is down
```
Rebooting the host resolves the problem temporarily. 
See also https://github.com/mKeRix/room-assistant/issues/314

agners:
What environment is that exactly? can you fill out the sections please?

Veldkornet:
Updated version info

Veldkornet:
I initially thought that it was just the Bluetooth Crashing, but it seems that all my USB dongles also stop working. 

See below from the OoenZWave add-on

```
[20201027 12:42:10.788 CET] [ozw.mqtt.publisher] [debug]: Removing Stale Topic/Msg:  "OpenZWave/1/node/44/instance/1/commandclass/115/value/844425682796561/" 
[20201027 12:42:10.788 CET] [ozw.mqtt.publisher] [debug]: Topics:  "OpenZWave/1/node/44/instance/1/commandclass/115/value/1125900659507220/" 
[...]

PR #962, opened 2020-11-05
Fix CDC ACM error recovery in Raspberry Pi kernels as well (#940)

agners:
Backport fixes for CDC ACM error recovery path.


agners:
Since we are moving to 5.4 on dev those patches only make sense on rel-4. On dev the first patch won't be necessary and the second patch I will add with the 5.4 change.

pvizeli (approved):

PR #973, opened 2020-11-09
Use performance governor to avoid frequency scaling

agners:
After running HAOS on my ODROID N2+ several hours I see freezes and
sometimes stack traces which point to a problem in CPU frequency
scaling. This crash seems not to appear on Hardkernel's 18.04 Ubuntu
stable release. However, Hardkernel's Ubuntu uses the performance
governor. Use the performance governor as well to avoid crashes on N2+.


buildroot-external/board/hardkernel/odroid-n2/kernel.config
```diff
@@ -0,0 +1 @@
+CONFIG_CPU_FREQ_DEFAULT_GOV_PERFORMANCE=y
```
pvizeli, line 1: Do we want to switch in general and opt-out on a platform which causes an issue or high power consumption?

agners: Since HAOS is not meant to be running on a battery powered device, power consumption is likely not an issue in our case I agree. So the only reason to not use performance is just to safe energy in general or if there are thermal constraints.

I just checked what the default is in upstream Linux, it actually was performance for ARM devices just until recently, where they switched to `CPU_FREQ_DEFAULT_GOV_SCHEDUTIL` (see https://lore.kernel.org/linux-pm/20200722093732.14297-8-ionela.voinescu@arm.com/).

So the question ends up really, do we have a device which performance would cause issue?

I did a quick survey.

On my ODROID-N2+ test the difference between performance and ondemand was roughly ~200mW in power consumption (I measure mains, relative power difference is probably not too bad, but this is really a rough measurement!). Thermal effect was not really noticeable, which make sense given the huge heat sink.

On a ODROID-C4 I noticed about 0.5°C temperature difference but no change in current consumption (which probably means it is below my measurement sensitivity which is around 150mW).

On a RPi 3 Model B I noticed 2°C temperature difference (47.8°C to 49.9°C in idle), but current measurement did not change either. This is without any heatsink.

On a RPi4 I did not noticed a temperature difference (stayed around 57.5°C) and current measurement also did not change. Again without heatsink.

Since thermal effect is really minimal, and CPU frequency scaling actually introduced overhead and a slight delay when stepping up I think using performance governor really make sense (I did not measure/notice the delay and it probably is minimal in practise, but there sure is a slight delay when changing frequencies).

agners: After thinking a bit more about this, I am all in for performance governor by default, but I think we should do it in rel-5 only. So I am going ahead and merge this to dev and backport to rel-4 first, before then switching to performance governor in a second step.

PR #981, opened 2020-11-11
Support for exFAT drives

tmm1:
This allows users to format their CONFIG drive using exFAT.

Currently only [FAT/NTFS/ext4 are supported](https://developers.home-assistant.io/docs/operating-system/debugging/#home-assistant-operating-system)


agners:
Hm, yeah not sure if it is that valueable for HAOS as we don't really have support for external storage.

I also saw that Linux 5.4 already supports exFAT in staging, specific reasons you chose the external module over the one in tree?

tmm1:
The staging driver is deprecated. The new driver from 5.7 is the one that works well and replaces the old staging driver.

See:

- https://www.phoronix.com/scan.php?page=news_item&px=Staging-exFAT-Nuking-5.7
- https://www.phoronix.com/scan.php?page=news_item&px=Linux-5.7-New-exFAT-Lands

agners:
@tmm1 thanks for the info.

Yeah I think as long as there is not really request for this feature we there is not much value for us adding/maintaining this. Early next year we likely will jump onto the next LTS for most boards where we can use the upstream driver.

Closing this, we can reopen in case there is really need for it.

PR #984, opened 2020-11-12
Let's use our online check file

pvizeli:


pvizeli:
Should work but is not tested

agners:
Hm, I see, this is required to make the NetworkManager connectivity check actually work. ArchLinux sets it in `/usr/lib/NetworkManager/conf.d/20-connectivity.conf`. I think we should also set it in `/usr/lib/NetworkManager/conf.d/` as this is a distribution wide setting...

ludeeus:
Wherever this setting end up, we should also adjust the supervised-installer script to do the same, and cache that file hard in cloudflare

pvizeli:
Do you want to split it from a single config into a multi config import on a read-only system? Your choice.

agners:
@ludeeus 
> Wherever this setting end up, we should also adjust the supervised-installer script to do the same
Given that distributions already set something I am not sure we should do that. Supervised always feels to me we should leave the underlying OS largely alone.

> cache that file hard in cloudflare
Yep!

@pvizeli maybe actually move the whole thing to `/usr/lib`. While we have `/etc/NetworkManager/*` not writeable by users today, I like to have that option for the future maybe. You can overwrite settings made by the distribution in `/usr/lib/NetworkManager/` from `/etc/NetworkManager/`, which might come in handy later.

PR #986, opened 2020-11-13
Use systemd-resolved to announce hostname via mDNS and LLMNR

agners:


buildroot-external/rootfs-overlay/etc/NetworkManager/system-connections/default
```diff
@@ -2,6 +2,8 @@
 id=HassOS default
 uuid=f62bf7c2-e565-49ff-bbfc-a4cf791e6add
 type=802-3-ethernet
+llmnr=2
+mdns=2
```
pvizeli, line 6: Can we do, so we need at this also on our profile over supervisor

pvizeli (approved):

buildroot-external/rootfs-overlay/etc/NetworkManager/system-connections/default
agners: I guess we can set this safely in supervisor since it won't affect older releases which don't use systemd-resolved yet.

PR #999, opened 2020-11-16
Add on-board RTC support for ODROID-N2+

agners:
Make the NXP PCF8563TS on-board RTC the primary RTC of the system.


pvizeli (approved):

pvizeli:
We should send that also upstream

PR #1004, opened 2020-11-17
Fix DNS resolving

agners:


pvizeli (approved):

filipvh:
@agners Will this solve the issue that my hassio is not using the dhcp set dns sever?

PR #1007, opened 2020-11-20
Make sure ALSA modprobe parameters get applied

agners:


pvizeli:
Do we not need move the `/usr/modprobe.d` folder to `/usr/lib/modprobe.d`?

agners:
That is what the first commit is doing: https://github.com/home-assistant/operating-system/pull/1007/commits/953dc7009823aef4aed99d404e7f854ca066f985

systemd seems to already install its conf into the right folder, I did not find any other modprobe config files.

pvizeli (approved):

PR #1009, opened 2020-11-20
Increase OVA disk image default size to 32GB

agners:
The current default size of 6GB can fill up pretty quickly. Since most
disk images we offer resize dynamically its not really problem to ship
with a bigger default size. It avoids support cases when people forget
to increase the disk image size.


ludeeus:
I like it, it will help a lot of users!

frenck:
Yes please 👍

pvizeli:
We need just keep in mind, it takes a lot of storage on our build system

PR #1017, opened 2020-11-24
Add dev build workflow

agners:


pvizeli (commented):
I'm fine if we move away to bump the version to meta and use all time these mechanics for version handling

PR #1018, opened 2020-11-24
Compress images using xz instead of gzip

agners:
The xz compression allows higher compression rates and higher speeds,
a quick measurement lead to this numbers:
gzip --best: compression 131.11s, decompression 9.797s (299M)
xz -3 (single thread): compression 95.13s, decompression 14.902s (228M)
xz -3 (multi thread): compression 12.146s, decompression 14.902s (228M)

### DIFF
--- a/buildroot-external/board/asus/hassos-hook.sh
+++ b/buildroot-external/board/asus/hassos-hook.sh
@@ -21,5 +21,5 @@ function hassos_pre_image() {
 
 
 function hassos_post_image() {
-    convert_disk_image_gz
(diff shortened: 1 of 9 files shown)


pvizeli (commented):
I see no issue if etcher support it now correctly

agners:
> I see no issue if etcher support it now correctly
Checked with current balenaEtcher Version 1.5.111 on Linux and it all seems good. I assume that it is fine on all platforms today. :crossed_fingers:

PR #1029, opened 2020-11-27
Use hdaudio for VMware by default (#925)

agners:
The HD-Audio virtual sound device is known to work with Home Assistant
OS. Make sure to setup such a device by default.


agners:
This doesn't work like that.

PR #1041, opened 2020-12-01
Set NetworkManager default value / fix OTA updates

pvizeli:


buildroot-external/rootfs-overlay/etc/NetworkManager/NetworkManager.conf
```diff
@@ -9,3 +9,7 @@ unmanaged-devices=type:bridge;type:tun;type:veth
 
 [logging]
 backend=journal
+
+[connection]
+connection.lldp=2
```
agners: Link Layer Discovery Protocol, is that what you intended to enable here?

agners (approved):

PR #1046, opened 2020-12-02
Rewrite datactl command

agners:
Prepare the target partition as part of the datactl command. Rely on
partlabel for the target disk since we are always using GPT on the
target disk. Use systemd and partlabel mechanism to wait and find
the target data disk. Keep using the file system label to identify
the source disk.

Also use e2image instead of raw dd to move data. This should
speed up the processes significantly.


pvizeli (commented):
https://github.com/home-assistant/operating-system/blob/dev/buildroot-external/bootloader/mbr-part.rules

We should make sure we clean up the old partition (GPT or MBR) and the filesystem label.

And you need to test it on RPi, N2, and OVA - all they handle it different

agners:
> We should make sure we clean up the old partition (GPT or MBR) and the filesystem label.

Currently I just rename the old file system label, which seems to get it out of the way nicely. I could delete the partition no problem, but is it worth? I mean this way we would have a go back option in case it turns out there are problems with the external disk or similar.

> And you need to test it on RPi, N2, and OVA - all they handle it different

From what I can tell, the data partition is always handled using the file system label.

Yes I will test it, but I plan to do so after merge, then I can build a dev build for all machines using the new dev build workflow.

agners:
Btw, I am using a different partlabel now for the external disk: `hassos-data-external` (while keeping the file system label `hassos-data`). This allows us to easily identify if an external data partition is used if necessary later on. Since we are always using GPT for the external disk this should work fine on all platforms.

agners:
Tested on a N2 as well, the script seems to work fine.

One thing I noticed is if data partition doesn't exist, the OS boots (after 90s) but behaves weired (login is not possible anymore). While this most often should not be the case, it is a situation which happens when removing the external disk after a successful move. I don't think it has to do with this change, as the `mnt-data.mount` is unchanged. Just something I noticed. Will take a look at that was well.

pvizeli (approved):

Issue #1045, opened 2020-12-02
Has os 5.6 only only boots by unplug and plugin SSD


snippem:
- [x] Raspberry Pi [1/2/3/4]
- [ ] ODROID [C2/C4/N2(+)/XU4]
- [ ] ASUS Tinker [S]
- [ ] Intel NUC (or compatible)
- [ ] OVA (Open Virtualization Applicance, on Intel NUC or any other hardware, please add the Hypervisor you are using)

**Home Assistant OS release:**
- [x] Fresh installation of release 5.6
- [ ] Updated from version 5.5
- Additional information (if accessible):
<!--

snippem:
Hi all.

Yesterday i updated from OS 5.5 to OS 5.6 with all the other versions never had the problem that my ssd won't boot at powering up.
Reinstalled a clean version off hassos 5.6 does not solve this issue.
So i presume something changed with OS 5.6 that prevents the pi4 from booting up. Changing the sequence in booting (first power on the pi then plug in the ssd ) lets HA boot but that is not the nicest way to boot haha.
I have 2 exactly the same RPI4 4GB with a kingston 128GB usb ssd so i can test whats working or not.
From what i have read here https://github.com/home-assistant/operating-system/pull/939
I have to wait for a new firmware version for the raspberry pi4 that is now still in beta? or is there something else i can do?
It's not the end of the world i know but just interested if the future updates wil solve this problem.

agners:
#939 is part of OS 5.6, so you should get the new firmware with 5.6. Can you check which firmware your RPi4 is using now? (Should show up on the initial screen on HDMI, the screen should stay there when you unplug the SSD/SD card).

This firmware is supposed to improve the situation yes, but since you are saying that 5.5 and before did not show the problem it seems the opposite is the case here?

@tmm1 have you noticed something like that with the 2020-10-28 firmware?

snippem:
> #939 is part of OS 5.6, so you should get the new firmware with 5.6. Can you check which firmware your RPi4 is using now? (Should show up on the initial screen on HDMI, the screen should stay there when you unplug the SSD/SD card).
> 
> This firmware is supposed to improve the situation yes, but since you are saying that 5.5 and before did not show the problem it seems the opposite is the case here?
> 
> @tmm1 have you noticed something like that with the 2020-10-28 firmware?

Will check this later what firmware the raspberry with hassos 5.6 is on as i am working atm.
[...]

PR #1103, opened 2020-12-17
Disable Raspberry Pi EEPROM upgrade (#1045)

agners:
The EEPROM upgrade 2020-10-28 causes issues with JMS583 or JMS580
controller from Jmicron. Others reported that the same update fixes
reboot issues. Currently there is no Raspberry Pi 4 firmware which works
for all cases. Therefor don't ship an EEPROM upgrade so users can flash
and continue using what works for their setup.


pvizeli (approved):

mirh:
Could be related https://github.com/raspberrypi/rpi-eeprom/issues/227

Issue #1120, opened 2020-12-22
oDroid N2+: # CONFIG_HIDRAW is not set


mvieleers:
**Hardware Environment**

- [ ] Raspberry Pi [1/2/3/4]
- [X] ODROID [C2/C4/N2(+)/XU4] (Home Assistant Blue device)
- [ ] ASUS Tinker [S]
- [ ] Intel NUC (or compatible)
- [ ] OVA (Open Virtualization Applicance, on Intel NUC or any other hardware, please add the Hypervisor you are using)

**Home Assistant OS release:**
- [ ] Fresh installation of release x.y
- [X] Updated from version 5.8 
- Additional information (if accessible): Home Assistant Blue came with 5.8, updated to 5.9

version: 2020.12.1
installation_type: Home Assistant OS
dev: false
hassio: true
docker: true
virtualenv: false
python_version: 3.8.6
os_name: Linux
os_version: 5.9.15
arch: aarch64
timezone: Europe/Amsterdam

GitHub API: ok
Github API Calls Remaining: 4501
Installed Version: 1.9.0
Stage: running
Available Repositories: 767
Installed Repositories: 28

logged_in: true
subscription_expiration: January 8, 2021, 1:00 AM
relayer_connected: true
remote_enabled: true
remote_connected: true
alexa_enabled: false
google_enabled: true
can_reach_cert_server: ok
can_reach_cloud_auth: ok
can_reach_cloud: ok

host_os: Home Assistant OS 5.9
update_channel: stable
supervisor_version: 2020.12.7
docker_version: 19.03.13
disk_total: 113.9 GB
disk_used: 6.9 GB
healthy: true
supported: true
board: odroid-n2
supervisor_api: ok
version_api: ok
installed_addons: Check Home Assistant configuration (3.6.0), Samba share (9.3.0), Mosquitto broker (5.1), Terminal & SSH (8.10.0), deCONZ (6.6.1), AppDaemon 4 (0.3.2), Visual Studio Code (2.9.1), Home Assistant Google Drive Backup (0.102.0), rest980 Docker Image (20200205), php-nginx Docker Image (latest)

dashboards: 14
mode: auto-gen
resources: 18

api_endpoint_reachable: ok

- 20-12-22 09:57:27 INFO (SyncWorker_0) [supervisor.docker.dns] Starting DNS homeassistant/aarch64-hassio-dns with version 2020.11.0 - 172.30.32.3
[...]

PR #1122, opened 2020-12-22
Enable hidraw driver (#1120)

agners:
The hidraw driver is required by some IoT devices such as Wyze sense or
Jablotron JA-100. Enable the driver on all platforms by default.


buildroot-external/kernel/hassos.config
```diff
@@ -44,3 +44,6 @@ CONFIG_CIFS_ACL=y
 CONFIG_CIFS_DFS_UPCALL=y
 
 CONFIG_WIREGUARD=m
+
+CONFIG_HID=y
+CONFIG_HIDRAW=y
```
pvizeli, line 49: as module?

PR #1140, opened 2020-12-29
Add hassos-apparmor dependency to supervisor

agners:
The supervisor container requires the "hassio-supervisor" AppArmor
profile. Make sure our AppArmor service hassos-apparmor is a dependency
of the hassos-supervisor.service.


agners:
This isn't something which caused issues in practice but I noticed it when stopping/starting service manually.

frenck (approved):

Issue #1150, opened 2020-12-31
Missing firmware for rtl8169


superm1:
**Hardware Environment**

- [ ] Raspberry Pi [1/2/3/4]
- [ ] ODROID [C2/C4/N2(+)/XU4]
- [ ] ASUS Tinker [S]
- [x] Intel NUC (or compatible)
- [ ] OVA (Open Virtualization Applicance, on Intel NUC or any other hardware, please add the Hypervisor you are using)

Dell IOT edge gateway 3003 (w/ 8G of MMC based storage).

**Home Assistant OS release:**
- [x] Fresh installation of release 5.9-dev
- [ ] Updated from version x.y
- Additional information (if accessible):

**Supervisor logs:**

**Journal logs:**

**Kernel logs:**

```
[    1.181063] r8169 0000:02:00.0: can't disable ASPM; OS doesn't have ASPM control
[    1.182145] libphy: r8169: probed
[    1.182700] r8169 0000:02:00.0 eth0: RTL8106e, 14:b3:1f:08:f8:3a, XID 449, IRQ 91
[    1.182877] r8169 0000:04:00.0: can't disable ASPM; OS doesn't have ASPM control
[    1.183938] libphy: r8169: probed
[    1.184536] r8169 0000:04:00.0 eth1: RTL8106e, 14:b3:1f:08:f8:3b, XID 449, IRQ 92
[    4.595808] r8169 0000:02:00.0 enp2s0: renamed from eth0
[    4.712136] r8169 0000:04:00.0 enp4s0: renamed from eth1
[   10.504619] r8169 0000:02:00.0: Direct firmware load for rtl_nic/rtl8106e-1.fw failed with error -2
[   10.504627] r8169 0000:02:00.0: Unable to load firmware rtl_nic/rtl8106e-1.fw (-2)
[   10.515660] RTL8208 Fast Ethernet r8169-200:00: attached PHY driver [RTL8208 Fast Ethernet] (mii_bus:phy_addr=r8169-200:00, irq=IGNORE)
[   10.697033] r8169 0000:02:00.0 enp2s0: Link is Down
[   10.788030] r8169 0000:04:00.0: Direct firmware load for rtl_nic/rtl8106e-1.fw failed with error -2
[   10.788038] r8169 0000:04:00.0: Unable to load firmware rtl_nic/rtl8106e-1.fw (-2)
[...]

PR #1152, opened 2021-01-02
Add r8169 firmware into intel_nuc build (Fixes: #1150)

superm1:


superm1:
Also FYI there are a ton of USB dongles and docks with r8169, so this might actually be beneficial for non-`intel_nuc` builds too.

superm1:
@agners any thoughts on this?

agners (approved):

agners:
Sorry for the delay, looks good!

PR #1188, opened 2021-01-28
Linux: Update kernel 5.4.93

agners:


agners:
We'll jump to 5.10 on dev directly, updating to 5.4.93 on `rel-5` only.

Issue #1125, opened 2020-12-23
Upgrade to OS 5.9 on OVA caused corrupted EFI folder


mj-sakellaropoulos:
Just updated to OS 5.9 via UI, vm no longer boots.
VM is on Proxmox 6.2, OVA, UEFI OVMF

Upon investigation in ubuntu, garbage data was found in the EFI folder : 
![image](https://user-images.githubusercontent.com/36081545/102948765-a9d3a300-4494-11eb-9c4a-bcc828588aec.png)

Will dd the boot partition from release page and report back, suspect update process is broken somehow ?

( as first reported here: https://github.com/whiskerz007/proxmox_hassos_install/issues/96 )

mj-sakellaropoulos:
After repairing the boot partition, the file system seemed intact but I am stuck on the barebox bootloader and 100% CPU usage : 
![image](https://user-images.githubusercontent.com/36081545/102950629-fde08680-4498-11eb-8cac-a12d75c99105.png)

agners:
There are two partition (A/B update system), you might have booted the old 5.8 release.

Did you by chance had to reset/force poweroff the VM? Can you reproduce the issue? You are not the only report along those lines, see #1092. I use libvirt (which uses kvm underneath) and did a bunch of updates using OVA, I wasn't able to reproduce this issue.

agners:
Which version did you upgrade from?

mj-sakellaropoulos:
5.8 to 5.9 via UI

I booted system0 and system1 via GRUB, let me know if there are other procedures to follow for booting specific versions.

The vm was not forced off by me, it did the update, corrupted EFI and rebooted. When i looked at VNC, it was saying cannot find boot entry.

mj-sakellaropoulos:
Just to clarify from my perspective there are the following multiple failures:

- OS update failed (?)
- Bootloader update failed, killing EFI
- New bootloader from 5.9 OVA (.qcow2.gz) unable to boot existing system (maybe related to #1092 bug?)

If there are any log files I can provide let me know.

I will try to repro this issue to extract some more data.

mj-sakellaropoulos:
MAJOR UPDATE :

- Bootloader hang is caused by having IDE DVD-ROM attached for debugging, remove IDE and barebox works properly
- System update worked, and system is running fully functional as long as booted by barebox (just need to manually start docker with ```systemctl start docker```)

The ONLY issue was the EFI corruption although the cause remains _unknown_
[...]

PR #1190, opened 2021-01-28
Disable fsck.fat for boot partition (might help #1125)

agners:
There are incident reports on the internet where poeple report that
fsck.(v)fat actually leads to problems rather file system fixes. Around
the time when Home Assistant OS added fsck.fat for the boot partition,
reports of empty boot partitions or file with weired filenames started
to appear. This could be caused by fsck.fat.

Disable fsck on the boot partition.


lexathon:
Works for me. Had problems on 5.9 and 5.10, but just updated to 5.11 with no issues. Thanks.

PR #1211, opened 2021-02-05
RaspberryPi: Update firmware 0d458874a89921fbe460e422b239695e1e101e2b

agners:


pvizeli:
You should use my script, which also updates the checksum. Or is the checksum not needed any more there?

agners:
I did use your script, but I think something went wrong. Will give it another try.

denics:
Hi guys, thanks for this. would it be possible to have the modules for the camera too? it is a real pity to not be able to use it!

PR #1213, opened 2021-02-05
Rename hassos to haos

agners:
This "only" renames the `HASSOS_ID` for now, which has far reaching consequences: Rauc compatible is handled via hook, but this will need supervisor change since the CPE product name changes as well.

### DIFF
--- a/buildroot-external/ota/rauc-hook
+++ b/buildroot-external/ota/rauc-hook
@@ -2,7 +2,27 @@
 
 ##
 # Hooks
-env
+
+case "$1" in
+     install-check)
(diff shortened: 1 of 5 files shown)


buildroot-external/ota/rauc-hook
```diff
@@ -2,7 +2,28 @@
...
 # Hooks
-env
+
+case "$1" in
+     install-check)
+         if [ "$RAUC_MF_COMPATIBLE" = "$RAUC_SYSTEM_COMPATIBLE" ]; then
+             exit 0
+         fi
+         # Be compatible with hassos OS ID
+	 # shellcheck disable=SC2039
```
pvizeli, line 12:
```suggestion
	   # shellcheck disable=SC2039
```

PR #1218, opened 2021-02-08
support for JetHome JetHub D1/H1 devices

adeepn:
Add support for JetHome JetHub D1/H1 devices
( http://jethome.ru/ )

https://github.com/home-assistant/core/pull/46215
https://github.com/home-assistant/builder/pull/69
https://github.com/home-assistant/supervisor/pull/2532


agners:
Thanks for the pull request.

I am not familiar with JetHub, seems like an nice home automation platform to run Home Assistant on. Will this device become available outside of Russia?

Some technical things I noticed:
- U-Boot seems to get pulled in as a binary. This is kinda a no-go, as U-Boot is an important part of our update system. This way we have no control over U-Boot if we want to change the update logic etc.
- There seems to be a rather large binary checked in: buildroot-external/board/jethome/bin/aml_image_v2_packer. From what I can tell it packs the image in a custom way. Image build is already quite complex, plulling in a board specific method which relies on a binary blob is a no-go

adeepn:
> I am not familiar with JetHub, seems like an nice home automation platform to run Home Assistant on. Will this device become available outside of Russia?
Yes, but a little bit later.

> * U-Boot seems to get pulled in as a binary. This is kinda a no-go, as U-Boot is an important part of our update system. This way we have no control over U-Boot if we want to change the update logic etc.
U-Boot will be vailable via official JetHome repository https://github.com/jethome-ru/jethub-prebuilt-u-boot (as odroid xu4 for example). All boot logic loaded from boot.scr which located on first partition (Bootloader partition).

> * There seems to be a rather large binary checked in: buildroot-external/board/jethome/bin/aml_image_v2_packer. From whatI can tell it packs the image in a custom way. Image build is already quite complex, plulling in a board specific method which relies on a binary blob is a no-go
This binary is only way to build amlogic image for burning into MMC. May be better to move it to https://github.com/home-assistant/operating-system-blobs or to the official JetHome repository?

agners:
> > * U-Boot seems to get pulled in as a binary. This is kinda a no-go, as U-Boot is an important part of our update system. This way we have no control over U-Boot if we want to change the update logic etc.
> 
> U-Boot will be vailable via official JetHome repository https://github.com/jethome-ru/jethub-prebuilt-u-boot (as odroid xu4 for example). All boot logic loaded from boot.scr which located on first partition (Bootloader partition).

ODROID-XU4's U-Boot is compiled as part of the buildroot build. Maybe I got that wrong, is JetHub just using usptream U-Boot with s400 defconfig?

Some of the boot logic has dependencies on U-Boot configs, so that is why we rather want to have U-Boot builds under control of the OS repository.

> 
> > * There seems to be a rather large binary checked in: buildroot-external/board/jethome/bin/aml_image_v2_packer. From whatI can tell it packs the image in a custom way. Image build is already quite complex, plulling in a board specific method which relies on a binary blob is a no-go
> 
> This binary is only way to build amlogic image for burning into MMC. May be better to move it to https://github.com/home-assistant/operating-system-blobs or to the official JetHome repository?

Not what you mean by that. The binary seems to create a binary image, just like we do using our scripts for the other boards. I wonder why is that required? I rather prefer to not have some propitiatory image mangling tool handling our main OS partitions etc...  What special magic is in there?

We do have some special tools required to generate bootable U-Boot binaries for other Amlogic boards, but those are limited to U-Boot and early firmware (bl2/bl3). But this tool seems to handle the main OS partitions.

adeepn:
> Maybe I got that wrong, is JetHub just using upstream U-Boot with s400 defconfig?
At this moment for jethub d1/h1 u-boot building we use custom patched version of sources from AmLogic. We will investigate moving to opensource uboot, but this not first priority task. For all time we will provide support for Hassio OS and use HassOS as one of preflashed os for JetHome devices.

> Some of the boot logic has dependencies on U-Boot configs, so that is why we rather want to have U-Boot builds under control of the OS repository.
This is not problem, we will support all needed changes before we can use opensource uboot. And, of course, after that. 

> I don't know what you mean by "the only way". The binary seems to create a binary image, just like we do using our scripts for the other boards. 
JetHome D1 device only has eMMC memory on board, no SD card. JetHome H1 use eMMC as main storage. For flashing eMMC via AmLogic Burning tool or from usb drive we must prepare an image using the AmLogic image packer utility. There is no working source code for this utility. In order not to litter the repository with binary blobs, I have moved the utilities into an additional package (jethome-amltools, https://github.com/home-assistant/operating-system/pull/1218/commits/28d888d7e41e5c1299edb03ae6a6eb674cc2347f). 
As an alternative, I can propose to create a regular MBR image, and then convert it into AML via binary utils in a separate package.

agners:
I am closing this as there is no ongoing discussion or progress towards addressing the review feedback.

adeepn:
We have two moments:
1. closed uboot. it has all needed features enabled: all commands, loading custom boot script for hassos. uboot has custom amlogic version with closed patches, based on 2015.10 uboot.
2. closed tool (created by amlogic) for create mmc burn image.
All of these located in buildroot-external/packages and loaded from Jethome official repositories. 
What we need to do to merge support for JetHome devices in Hassos?

agners:
>     1. closed uboot. it has all needed features enabled: all commands, loading custom boot script for hassos. uboot has custom amlogic version with closed patches, based on 2015.10 uboot.

Even if it works all fine now, that is not good enough for a open source project. When we want to introduce new features, we need to deal with all platforms to make sure we don't break boards. If one of the board needs a special process to get a new U-Boot feature enabled, it will slow down the project as a whole.

Currently we have upstream U-Boot support for all boards, and we'd like to keep it that way. A couple of patches are ok, if they are in the process of being up-streamed. But a board being stuck on downstream vendor U-Boot is a no-go.

>     2. closed tool (created by amlogic) for create mmc burn image.
>        All of these located in buildroot-external/packages and loaded from Jethome official repositories.
>        What we need to do to merge support for JetHome devices in Hassos?

Same here: What if we want to adjust our image building process in a future release? If JetHome support is in the main repository, it will mean additional maintenance effort.

Now I am willing to make exception if a certain SoC is very new and popular, but doesn't have a good upstream support/or open toolchain to build a bootable image. But this is not the case here, the Amlogic S905 SoC is also used in ODROID-C2 and for that we have upstream U-Boot support and the image toolchain uses our regular image build flow.

There is a pull-request formalizing some of those rules. They are still under discussion, but in the grand scheme of things that is what we expect for a new board support:
https://github.com/home-assistant/architecture/blob/hardware-screening-os/adr/0017-hardware-screening-os.md

adeepn:
1. I got permission to publish u-boot sources. Can we build u-boot from custom sources from JetHome github repository? or You need to use mainstream u-boot sources?
2. We can build images for SD card without binary aml_image_packer. But JetHub D1 does not have SD card slot. For burning image to eMMC memory we need to use proprietary AmLogic utility.

PR #1233, opened 2021-02-18
Update ovf:id to get Synology VMM detect Guest Agent

fredrike:
The Synology VMM only detects running Guest Agent if the `ovf:id="102"`. I've done some extensive testing and this is the only way to get the VMM to detect Guest Agent.

I'm aware of #660 but I don't have access to ESXi so I can't verify that this change will work for those users too.

The `102` and `otherGuest64` have been matched from this page: https://wiki.abiquo.com/display/ABI310/Extended+OVF+Support+and+Template+Definition

### DIFF
--- a/buildroot-external/board/intel/ova/home-assistant.ovf
+++ b/buildroot-external/board/intel/ova/home-assistant.ovf
@@ -21,7 +21,7 @@
       <Product>Home Assistant Operating System</Product>
       <ProductUrl>https://www.home-assistant.io</ProductUrl>
     </ProductSection>
-    <OperatingSystemSection ovf:id="100" vmw:osType="other4xLinux64Guest">
+    <OperatingSystemSection ovf:id="102" vmw:osType="otherGuest64">
       <Info>The kind of installed guest operating system</Info>
       <Description>Linux</Description>
     </OperatingSystemSection>


pvizeli:
Maybe `101` work too?

fredrike:
No, I've tested that and it didn't work. I really don't know what @Synology are doing here but 102 is the one that works.

pvizeli:
Maybe we can fix it with using `ovf:required="false"` there and allow Synology to select an other model

fredrike:
Doesn't ESXi accept my change.

I'm unsure how Synology could "select an other model". But I'm happy to test it.

agners:
Maybe @ludeeus can help test this?

ludeeus:
Both 100 and 102 works on ESXi, personally I'd prefer the 102 variant;
image
"Other 2.6.x Linux (64-bit)" which 100 display, just looks messy and wrong :shrug:

pvizeli:
What is about https://github.com/home-assistant/operating-system/issues/660 ?

ludeeus:
#696 looks more to be the cause there

jens-maus:
I just stumbled over this issue. Please note that in the RaspberryMatic project I am maintaining we are also using a template `.ovf` file and there we perfectly use `ovf:id="100" vmw:osType="other4xLinux64Guest"`. And this .ovf works perfectly fine with Synology VMM as well as with all other virtual environments RaspberryMatic supports (e.g. also ESXi, QNAP, etc. etc.)

So I would suggest to simply compare your .ovf to the one used in the RaspberryMatic project:
https://github.com/jens-maus/RaspberryMatic/blob/master/buildroot-external/board/ova/template.ovf

fredrike:
@pvizeli are you waiting for me to do some more tests?

fredrike:
> I just stumbled over this issue. Please note that in the RaspberryMatic project I am maintaining we are also using a template `.ovf` file and there we perfectly use `ovf:id="100" vmw:osType="other4xLinux64Guest"`. And this .ovf works perfectly fine with Synology VMM as well as with all other virtual environments RaspberryMatic supports (e.g. also ESXi, QNAP, etc. etc.)
> 
> So I would suggest to simply compare your .ovf to the one used in the RaspberryMatic project:
> https://github.com/jens-maus/RaspberryMatic/blob/master/buildroot-external/board/ova/template.ovf

Well your own picture talks against you..

https://github.com/jens-maus/RaspberryMatic/wiki/images/ova/synology/synology-schritt0.png

Although most of the documentation is in German (so I could not understand what the purpose of RaspberryMatic is) you don't seem to have the **Guest Agent** running in your ova. This patch enables VMM to detect it from the Homeassistant-OS ova.

jens-maus:
> > So I would suggest to simply compare your .ovf to the one used in the RaspberryMatic project:
> > https://github.com/jens-maus/RaspberryMatic/blob/master/buildroot-external/board/ova/template.ovf
> 
> Well your own picture talks against you..
> 
> https://github.com/jens-maus/RaspberryMatic/wiki/images/ova/synology/synology-schritt0.png
> 
> Although most of the documentation is in German (so I could not understand what the purpose of RaspberryMatic is) you don't seem to have the **Guest Agent** running in your ova. This patch enables VMM to detect it from the Homeassistant-OS ova.

Sorry, but this is unfortunately a fairly old screenshost where qemu guest agent support was not yet added to RaspberryMatic. See here for a recent one:

Bildschirmfoto 2021-02-23 um 22 04 37

As you can spot (even in that german screenshot) there "Guest Agent" is perfectly mentioned as "running" and I can perfectly shutdown the ova machine or see statistics within the Synology VMM interface, thus the guest additions are working fine in that `ovf:id=100` machine. So I would indeed suggest to merge the template ovf used here in HAos with the one from RaspberryMatic as it cannot be the `ovf:id=100` alone that is the reason why the guest additions don't work on your end. Back then I have invested quite and amount of time myself to find the best `*.ovf` template for all diferent virtual platforms supported (ESXi, Proxmox VE, VirtualBox, Synology VMM, QNAP, etc.)

BTW: If you want,  read up about RaspberryMatic  on the main GitHub project pages of it. And if you want, simply download the latest `*.ova` from the releases page and start it in your Synology VMM to view the guest additions being perfectly recognized even thought it was build with https://github.com/jens-maus/RaspberryMatic/blob/master/buildroot-external/board/ova/template.ovf

fredrike:
> As you can spot (even in that german screenshot) there "Guest Agent" is perfectly mentioned as "running" and I can perfectly shutdown the ova machine or see statistics within the Synology VMM interface, thus the guest additions are working fine in that `ovf:id=100` machine. So I would indeed suggest to merge the template ovf used here in HAos with the one from RaspberryMatic as it cannot be the `ovf:id=100` alone that is the reason why the guest additions don't work on your end. Back then I have invested quite and amount of time myself to find the best `*.ovf` template for all diferent virtual platforms supported (ESXi, Proxmox VE, VirtualBox, Synology VMM, QNAP, etc.)

Yes I was a bit quick to draw conclusions, did test your ova and indeed the Guest Agent is detected.

Could it be that hass-os uses UEFI bios that requires the ID to be 102??

Will test with your ovf and the hass-os image and see if that works.

fredrike:
The RaspberryMatic ovf works fine for the Homeassistant-OS vmdk image (detects Guest Agent just fine).

Will do some diffs between the two ovf-files, but this seams really magical (thought that I've eliminated all options but apparently no).

agners:
The meaning of the ID seems to map to the following (from https://opennodecloud.com/howto/2013/12/25/howto-ON-ovf-reference.html):
```
"Linux 2.4.x 64-Bit" [98]
"Linux 2.6.x" [99]
"Linux 2.6.x 64-Bit" [100]
"Linux 64-Bit" [101]
"Other 64-Bit" [102]
```

In theory I guess `101` would be the best match, but if `102` works better, that is fine for me.

@jens-maus I'd rather prefer to stick with ours. I diffed the two template, there are differences here and there, some minor (naming etc), some bigger (like EFI vs. BIOS, or Audio/Networking configuration).

Issue #1235, opened 2021-02-18
CONFIG_RT_GROUP_SCHED=y leaves docker with no available cpu.rt_runtime_us resources to share


jens-maus:
**Hardware Environment**

- [ ] Raspberry Pi [1/2/3/4]
- [ ] ODROID [C2/C4/N2(+)/XU4]
- [ ] ASUS Tinker [S]
- [ ] Intel NUC (or compatible)
- [X] OVA (Open Virtualization Applicance, on Intel NUC or any other hardware, please add the Hypervisor you are using)

**Home Assistant OS release:**
- [X] Fresh installation of development version 6.0dev
- [ ] Updated from version x.y
- Additional information (if accessible):

**Supervisor logs:**

**Journal logs:**

**Kernel logs:**

**Description of problem:**

The use of `CONFIG_RT_GROUP_SCHED=y` in [docker.config](https://github.com/home-assistant/operating-system/blob/dev/buildroot-external/kernel/docker.config#L5) results in docker and its child containers to not be able to use system functions like `sched_setscheduler()` to raise a process' realtime priority in case of time critical applications even thought `SYS_NICE` capability is assigned to a container.

Note the following default /sys output in a fresh HomeAssistantOS install:

1. cpu.rt_runtime_use cgroup of system:
   ```sh
   # cat /sys/fs/cgroup/cpu/cpu.rt_runtime_us 
   950000
   ```
2. cpu.rt_runtime_use cgroup of docker:
   ```sh
   # cat /sys/fs/cgroup/cpu/docker/cpu.rt_runtime_us 
   0
   ```
3. cpu.rt_runtime_use cgroup of a started docker container:
    ```sh
    # cat /sys/fs/cgroup/cpu/docker/fc74c542b573e272a600b46c76fbd32dd0d74c8e04914f00cfe504871e97e9b9/cpu.rt_runtime_us 
    0
    ```    

Please note that the output of (2) and (3) suggests that not only docker itself is not allowed to use functions like `sched_setscheduler()` to raise the realtime priority, but also its started containers won't be able to use processes which try to raise the real-time priority of the scheduler. One example of an add-on trying to use a process which uses `sched_setscheduler()` can be found here: https://github.com/jens-maus/RaspberryMatic/issues/1087#issuecomment-780888083. However, other applications might exist which require a process to raise the realtime priority of processes (e.g. like real-time audio apps, etc.)

[...]

PR #1236, opened 2021-02-19
Add --cpu-rt-runtime to allow Docker allocate real-time CPU time (#1235)

agners:


pvizeli (approved):

pvizeli:
https://github.com/home-assistant/supervisor/pull/2593

jens-maus:
This refs https://github.com/home-assistant/operating-system/issues/1235

jens-maus:
I tested the changes suggested by this PR in HAos and I could reproduce that at least docker has now enough `cpu.rt_runtime_us` resources assigned:

```
# cat /sys/fs/cgroup/cpu/docker/cpu.rt_runtime_us
950000
```

However, the suggested change by @pvizeli in https://github.com/home-assistant/supervisor/pull/2593 seems not be enough to get each container started by docker enough real-time shares so that applications using `sched_setscheduler()` are able to run correctly. So even after using a supervisor developer build with the changes by  https://github.com/home-assistant/supervisor/pull/2593 and the new `realtime` option used in `config.json` of the add-on ends up in ...

```
/ # cat /sys/fs/cgroup/cpu/cpu.rt_runtime_us
0
/ # multimacd -l 0 -c
[...]
2021/02/21 18:55:42.923 <Warning> Could not set scheduling priority: Operation not permitted
2021/02/21 18:55:42.923 <Warning> Could not set scheduling priority: Operation not permitted
2021/02/21 18:55:42.923 <Warning> Could not set scheduling priority: Operation not permitted
``` 

Reviewing the changes suggested by https://github.com/home-assistant/supervisor/pull/2593 it looks like only the `--ulimit rtprio=99` option is added by the new `realtime` option when starting the docker container of add-ons. However, AFAIK the `--cpu-rt-runtime=950000` option is also required to get the `cpu.rt_runtime_us` cgroup raised accordingly.

See https://docs.docker.com/config/containers/resource_constraints/#configure-the-realtime-scheduler

pvizeli:
https://github.com/home-assistant/supervisor/pull/2602

Need also an `SUPERVISOR_CPU_RT=1` on the env @agners

pvizeli (changes requested):
`SUPERVISOR_CPU_RT=1` for supervisor startup

jens-maus:
@agners @pvizeli

Sorry to bring this up here again, but I actually tried to check if the changes mentioned here (in HAos and the supervisor) actually work so that the new `realtime` option in an add-on actually brings up the container with enough real-time cpu time now. For this I installed my own dev version of HAos (from latest `dev` branch) with the changes from @agners and switched immediately to the beta channel so that supervisor 2021.02.12 was installed and is running. In fact, while I can see that `dockerd` is perfectly running with `--cpu-rt-runtime` and the new `SUPERVISOR_CPU_RT` env variable is propagated to the supervisor, my test add-on  container is still outputting the following

```
# cat /sys/fs/cgroup/cpu/cpu.rt_runtime_us
0
```

which suggests that the `cpu.rt_runtime_us` setup does not work as expected even thought I have `"realtime": true` defined in the `config.json` of the test add-on. So did someone from you actually test that the new combination of the HAos changes work nicely together with the new 2021.02.12 changes regarding real-time changes for add-ons.

Issue #1242, opened 2021-02-23
WiFi dongle drivers fail to load


yozik04:
**Hardware Environment**

- [x] Raspberry Pi [1/2/3/4]
- [ ] ODROID [C2/C4/N2(+)/XU4]
- [ ] ASUS Tinker [S]
- [ ] Intel NUC (or compatible)
- [ ] OVA (Open Virtualization Applicance, on Intel NUC or any other hardware, please add the Hypervisor you are using)

**Home Assistant OS release:**
- [ ] Fresh installation of release x.y
- [x] Updated from version 2020.x
- Additional information (if accessible):

Version | core-2021.2.1
-- | --
Installation Type | Home Assistant OS
Development | false
Supervisor | true
Docker | true
Virtual Environment | false
Python Version | 3.8.7
Operating System Family | Linux
Operating System Version | 5.4.83-v7
CPU Architecture | armv7l
Timezone | Europe/Tallinn

**Supervisor logs:**

Irrelevant
**Journal logs:**

Irrelevant
**Kernel logs:**

Relevant log:
```
[   12.485748] ieee80211 phy0: rt2x00lib_request_firmware: Info - Loading firmware file 'rt2870.bin'
[   12.485905] rt2800usb 1-1.2:1.0: Direct firmware load for rt2870.bin failed with error -2
[   12.485928] ieee80211 phy0: rt2x00lib_request_firmware: Error - Failed to request Firmware
```

Full log:
```
[    0.000000] Booting Linux on physical CPU 0xf00
[    0.000000] Linux version 5.4.83-v7 (builder@afcfacd4d5bd) (gcc version 9.3.0 (Buildroot -g1a551c07)) #1 SMP Sun Jan 31 02:31:00 UTC 2021
[    0.000000] CPU: ARMv7 Processor [410fc075] revision 5 (ARMv7), cr=10c5387d
[    0.000000] CPU: div instructions available: patching division code
[    0.000000] CPU: PIPT / VIPT nonaliasing data cache, VIPT aliasing instruction cache
[    0.000000] OF: fdt: Machine model: Raspberry Pi 2 Model B Rev 1.1
[    0.000000] Memory policy: Data cache writealloc
[    0.000000] Reserved memory: created CMA memory pool at 0x37400000, size 64 MiB
[    0.000000] OF: reserved mem: initialized node linux,cma, compatible id shared-dma-pool
[    0.000000] On node 0 totalpages: 242688
[...]

PR #1248, opened 2021-02-26
Add Ralink rt27xx/rt28xx/rt30xx firmware (#1242)

agners:
Add Ralink firmware for devices which have the driver enabled. The
firmwares are with a total of 20KiB rather small.


pvizeli:
Why do we add a Wireless USB card to a board with wifi supported?

agners:
I just added it everywhere where the drivers have been enabled. I feel an enabled driver warrants the firmwares to be present.

Alternatively we can disable the driver, but I think currently all RPis use the same kernel config.

agners:
@pvizeli removed most WiFi drivers from Tinker Board, as drivers without firmware really are useless.

For Raspberry Pi, this is a bit more painfull as we use defconfig from the kernel tree. I just don't add the Ralink firmware where on-board WiFi is there.

pvizeli (commented):
LGTM. It was just a question :) Do you want not to add this to odroid?

PR #1277, opened 2021-03-18
Rename NetworkManager default profile

agners:
Rename the NetworkManager default profile to "Home Assistant OS
default". Improve documentation on how to reset to default
configuration.


cogneato (commented):
This looks good assuming that the ha command is requiring an update to support that. Removing the system-connections directory and rebooting restored the "HassOS default" profile.

PR #1278, opened 2021-03-18
Restore AppArmor profile when not present on the system

agners:
Restore our default AppArmor profile when not present on the system.
This allows to implement factory reset.


buildroot-external/rootfs-overlay/usr/libexec/hassos-apparmor
```diff
@@ -1,12 +1,18 @@
...
 PROFILES_DIR="/mnt/data/supervisor/apparmor"
 CACHE_DIR="${PROFILES_DIR}/cache"
 REMOVE_DIR="${PROFILES_DIR}/remove"
 
 # Check folder structure
-mkdir -p "${PROFILES_DIR}"
+if [ ! -d "${PROFILES_DIR}" ]; then
+    echo "[INFO]: AppArmor profile missing, downloading..."
+    mkdir -p "${PROFILES_DIR}"
+    systemctl start network-online.target
```
pvizeli, line 13: we can also add network.target to wanted on the service file

agners: That would delay AppArmor on every boot. With this method, we only do it if the profile is missing. Since that is almost never the case, I prefer to have that dependency not by default.

PR #1314, opened 2021-04-08
Use interface-name to exclude veth

agners:
The type veth is not a valid type (see [1] for how to obtain a list of
valid device types. Use interface name with globbing to filter veth.

Note: It seems that NetworkManager did not manage veth so far, so this
change seems not to be relevant in practise.

[1] https://people.freedesktop.org/~lkundrak/nm-docs/NetworkManager.conf.html#device-spec

### DIFF
--- a/buildroot-external/rootfs-overlay/etc/NetworkManager/NetworkManager.conf
+++ b/buildroot-external/rootfs-overlay/etc/NetworkManager/NetworkManager.conf
@@ -5,7 +5,7 @@ autoconnect-retries-default=0
 rc-manager=file
 
 [keyfile]
-unmanaged-devices=type:bridge;type:tun;interface-name:veth*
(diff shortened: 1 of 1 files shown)


buildroot-external/rootfs-overlay/etc/NetworkManager/NetworkManager.conf
```diff
@@ -5,7 +5,7 @@ autoconnect-retries-default=0
 rc-manager=file
 
 [keyfile]
-unmanaged-devices=type:bridge;type:tun;type:veth
+unmanaged-devices=type:bridge;type:tun;interface-name:veth*
```
pvizeli, line 8: According: https://developer.gnome.org/NetworkManager/stable/NetworkManager.conf.html#device-spec

```suggestion
unmanaged-devices=type:bridge;type:tun;driver:veth
```

PR #1347, opened 2021-04-29
Use built-in Meson DRM driver

agners:
It seems that the crash of the Meson DRM driver on shutdown can also be
fixed by compiling it in. The driver is also built-in in LibreELEC,
hence this is better tested by the upstream community.

Note the underlying issue seems to be a disabled clock: Since the
introduction of meson_drv_shutdown some registers are touched at a very
late stage. Those clock get disabled in meson_ee_pwrc_shutdown. It seems
that when the driver is built-in, meson_drv_shutdown gets called before
meson_ee_pwrc_shutdown and hence sidesteps the problem.

Note: This increases the kernel by a bit since DRM needs to be built-in
as well. Configure some less common used file systems as modules
(ext3/NFS).


devova:
Thank you a lot. Does it mean that it would enable hw video decoding?

PR #1368, opened 2021-05-19
Recreate Supervisor container on OS upgrade/downgrade

agners:
When the operating system gets upgraded or downgraded the Supervisor
start script might start the Supervisor slightly differently (e.g. with
RT scheduling support). If the container has already been created, a OS
upgrade or downgrade won't recreate the Supervisor container.

### DIFF
--- a/buildroot-external/rootfs-overlay/usr/sbin/hassos-supervisor
+++ b/buildroot-external/rootfs-overlay/usr/sbin/hassos-supervisor
@@ -11,6 +11,7 @@ set -e
 # Init supervisor
 SUPERVISOR_DATA=/mnt/data/supervisor
 SUPERVISOR_STARTUP_MARKER="/run/supervisor/startup-marker"
+SUPERVISOR_STARTSCRIPT_VERSION="${SUPERVISOR_DATA}/.hassos-supervisor-version"
 SUPERVISOR_IMAGE="homeassistant/${SUPERVISOR_ARCH}-hassio-supervisor"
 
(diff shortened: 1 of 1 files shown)


agners:
This addresses an issue discussed in the RaspberryMatic issue tracker: https://github.com/jens-maus/RaspberryMatic/issues/1087#issuecomment-839794191

Thanks @stefangries for testing 6.0.rc1 and @jens-maus @pvizeli to pin point the issue.

buildroot-external/rootfs-overlay/usr/sbin/hassos-supervisor
```diff
@@ -11,6 +11,7 @@ set -e
 # Init supervisor
 SUPERVISOR_DATA=/mnt/data/supervisor
 SUPERVISOR_STARTUP_MARKER="/run/supervisor/startup-marker"
+SUPERVISOR_STARTSCRIPT_VERSION="${SUPERVISOR_DATA}/.hassos-supervisor-version"
```
pvizeli, line 14: I think we can but it to `/mnt/data/` or `/var/lib/supervisor/`, I don't think the supervisor need or should have access to this file. But that is my only comment on it, others looks fine I guess.

agners: Hm, `/var/lib/supervisor/` would need another bind mount. Its not that important, we can move it easily if needed. I'll just store it in `/mnt/data/` for now.

Issue #1339, opened 2021-04-27
HA not booting in KVM


MrksHfmn:
**Description of problem:**
I am trying to boot HA in a VM (KVM), however it only boots up to "Loading MS-DOS executable '/mnt/system/bzImage'" and stops there (I think). Also no IP is obtained from the DHCP server. The PCI device is a network card that is passed through via PCI passthrough.

**Hardware Environment**
- [ x] Generic x86-64 (like Intel NUC)

**Home Assistant OS release:**
hassos_ova-5.13.qcow2.xz

**virsh import**
```
virt-install \
        --import \
        --name home-assistant \
        --memory 4096 \
        --vcpus 4 \
        --cpu host \
        --disk home-assistant.qcow2,format=qcow2,bus=virtio \
        --nonetworks \
        --os-type=linux \
        --os-variant=debian9 \
        --graphics none \
        --noautoconsole \
        --boot uefi \
        --host-device=pci_0000_01_00_1
```

**virsh dumpxml home-assistant**

```
<domain type='kvm' id='17'>
  <name>home-assistant</name>
  <uuid>13357bd0-000c-496c-95af-602ebdb7999c</uuid>
  <metadata>
    <libosinfo:libosinfo xmlns:libosinfo="http://libosinfo.org/xmlns/libvirt/domain/1.0">
      <libosinfo:os id="http://debian.org/debian/9"/>
    </libosinfo:libosinfo>
  </metadata>
  <memory unit='KiB'>4194304</memory>
  <currentMemory unit='KiB'>4194304</currentMemory>
  <vcpu placement='static'>4</vcpu>
  <resource>
    <partition>/machine</partition>
  </resource>
  <os>
    <type arch='x86_64' machine='pc-q35-5.2'>hvm</type>
    <loader readonly='yes' type='rom'>/usr/share/ovmf/OVMF.fd</loader>
    <nvram>/var/lib/libvirt/qemu/nvram/home-assistant_VARS.fd</nvram>
    <boot dev='hd'/>
  </os>
  <features>
    <acpi/>
    <apic/>
  </features>
  <cpu mode='custom' match='exact' check='full'>
    <model fallback='forbid'>Snowridge</model>
    <vendor>Intel</vendor>
    <feature policy='require' name='ss'/>
    <feature policy='require' name='vmx'/>
    <feature policy='require' name='pdcm'/>
    <feature policy='require' name='hypervisor'/>
    <feature policy='require' name='tsc_adjust'/>
[...]

PR #1378, opened 2021-05-26
Enable serial console as well (#1339)

agners:
Using console focused virtualization environments such as virsh having
a serial console is the easiest way to interact with a virtual machine.
It also saves resources since no video memory needs to be allocated.
Enable serial console besides tty1 by default.

Note: The bootloader as well as the kernel shows its boot messages on
all consoles. However, only the last console is mapped to /dev/console,
which systemd is using to show service startup messages. Putting tty1 as
last console makes sure that systemd messages are still shown on the
console screen.

### DIFF
--- a/buildroot-external/board/pc/ova/hassos-hook.sh
+++ b/buildroot-external/board/pc/ova/hassos-hook.sh
@@ -10,7 +10,7 @@ function hassos_pre_image() {
     cp "${BINARIES_DIR}/barebox.bin" "${BOOT_DATA}/EFI/BOOT/BOOTx64.EFI"
     cp "${BR2_EXTERNAL_HASSOS_PATH}/bootloader/barebox-state-efi.dtb" "${BOOT_DATA}/EFI/barebox/state.dtb"
 
-    echo "console=tty1" > "${BOOT_DATA}/cmdline.txt"
+    echo "console=ttyS0 console=tty1" > "${BOOT_DATA}/cmdline.txt"
 }
 
 


buildroot-external/board/pc/ova/hassos-hook.sh
```diff
@@ -10,7 +10,7 @@ function hassos_pre_image() {
     cp "${BINARIES_DIR}/barebox.bin" "${BOOT_DATA}/EFI/BOOT/BOOTx64.EFI"
     cp "${BR2_EXTERNAL_HASSOS_PATH}/bootloader/barebox-state-efi.dtb" "${BOOT_DATA}/EFI/barebox/state.dtb"
 
-    echo "console=tty1" > "${BOOT_DATA}/cmdline.txt"
+    echo "console=ttyS0 console=tty1" > "${BOOT_DATA}/cmdline.txt"
```
pvizeli, line 13: 🤷 I don't know if that work as suspected on all kind of hypervisors/vm to be a default option. I mean people can edit the cmdline.txt if they need it. If it works over all systems, why not.

agners: The problem is, if you run on a headless system, its rather hard to change it (since you need to enable graphical access via VNC or similar etc. to only change that setting).

From my testing it doesn't seem to cause problems when serial port is not enabled (tested with KVM and Virtual Box, in the latter no serial port is the default it seems).

PR #1388, opened 2021-06-01
Enhancement: add thinkpad_acpi module

herrfrei:
The thinkpad_acpi module is needed to adjust the battery charge thresholds (see [tlp homepage](https://linrunner.de/tlp/faq/battery.html)) 
when using a newer Thinkpad laptop (T430(s)/T530/W530/X230 and all later) for hassos.


homeassistant:
Hi @herrfrei, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;operating-system#1388).

Thanks!

buildroot-external/board/pc/generic-x86-64/kernel.config
```diff
@@ -23,3 +23,6 @@ CONFIG_MMC_SDHCI=y
 CONFIG_MMC_SDHCI_ACPI=y
 CONFIG_MMC_SDHCI_PCI=y
 CONFIG_X86_INTEL_LPSS=y
+
+# Lenovo Thinkpad ACPI module
```
agners, line 27: This seems obvious from the name, so:

```suggestion
```

agners (changes requested):

agners:
How do you change the threshold then? I mean, we don't have provisions in HAOS to change it. Are there reasonable defaults when the module is loaded?

herrfrei:
Changing these values requires writing to `/sys/class/power_supply/BATx/charge_start_threshold` and `/sys/class/power_supply/BATx/charge_stop_threshold`. BATx might be BAT0, BAT1, ...

I'm not sure what the default values are, may be they differ between devices.

agners (approved):

PR #1393, opened 2021-06-06
Verify Power Supply Size

Mariusthvdb:
Because of intermittent inadequate power issues and https://jamesachambers.com/raspberry-pi-4-ubuntu-20-04-usb-mass-storage-boot-guide/ explicitly stating 3.5A is needed to run an SSD off of a Rpi, added the paragraph

### DIFF
--- a/Documentation/partition.md
+++ b/Documentation/partition.md
@@ -77,5 +77,5 @@ $ systemctl status mnt-data.mount
 If the data partition was moved to your USB drive you should see ```sh Active: active (mounted)  ``` in the output. Also, it will show, which drive got mounted as /mnt/data (```sh Where ``` and ```sh what ``` section of the output)
 
 
-## Power supply
+## Verify Power Supply Size (3.5A strongly recommended)
 Using an attached SSD will most likely draw more power than the official Power supply that go with the Rpi can deliver. Advice is to use one that provides at least 3.5A These will not be available in every region. Alternative can be to use a powered USB hub. Connect the Hub to 1 of the USB slots of your Rpi, and connect the SSD to the Hub. The PS that came with the Hub will power the attached device(s), and the PS for the Pi will only have to take care of the Pi.
--- a/Documentation/partition.md
+++ b/Documentation/partition.md
@@ -75,3 +75,7 @@ Within the Home Assistant interface you won't see if the move was succesful. To 
 $ systemctl status mnt-data.mount
 ```
 If the data partition was moved to your USB drive you should see ```sh Active: active (mounted)  ``` in the output. Also, it will show, which drive got mounted as /mnt/data (```sh Where ``` and ```sh what ``` section of the output)
+
+
+## Power supply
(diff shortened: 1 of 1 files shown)


Mariusthvdb:
for more background please see: https://community.home-assistant.io/t/pi4-ssd-too-much-to-handle-for-powersupply-replace-with/313606/14

(happening since HA 2021.6)

Documentation/partition.md
```diff
@@ -75,3 +75,7 @@ Within the Home Assistant interface you won't see if the move was succesful. To
 $ systemctl status mnt-data.mount
 ```
 If the data partition was moved to your USB drive you should see ```sh Active: active (mounted)  ``` in the output. Also, it will show, which drive got mounted as /mnt/data (```sh Where ``` and ```sh what ``` section of the output)
+
+
+## Verify Power Supply Size (3.5A strongly recommended)
+Using an attached SSD will most likely draw more power than the official Power supply (PS) that comes with the Rpi can deliver. Advice is to use one that provides at least 3.5A These will not be available in every region. Alternatively use a powered USB hub. Connect the Hub to 1 of the USB slots of your Rpi, and connect the SSD to the Hub. The PS that came with the Hub will power the attached device(s), and the PS for the Rpi will only have to take care of the Rpi.
```
agners, line 81: This is a generic section. It's ok to mention Raspberry Pi, but I think we need to be a bit more clear. Also the official Raspberry Pi 4 power supply has a voltage of 5.1V to counter voltage drop a bit, but I doubt that 5.1V power supplies are commonly available.

```suggestion
Using an USB attached SSD can draw quite some power. For instance on Raspberry Pi 3 the official Raspberry Pi power supply (PSU) only provides 2.5A which can be too tight. Use a power supply which can at least provide 3.5A. Alternatively use a powered USB hub. Connect the Hub to one of the USB slots of your Raspberry Pi, and connect the SSD to the Hub. The power supply that came with the Hub will power the attached device(s).
```

Mariusthvdb: That is a perfect improvement, thanks

Mariusthvdb:
hey Stefan, anything needed on my part? I did commit your changes did I?

Documentation/partition.md
```diff
@@ -75,3 +75,7 @@ Within the Home Assistant interface you won't see if the move was succesful. To
 $ systemctl status mnt-data.mount
 ```
 If the data partition was moved to your USB drive you should see ```sh Active: active (mounted)  ``` in the output. Also, it will show, which drive got mounted as /mnt/data (```sh Where ``` and ```sh what ``` section of the output)
+
+
+## Verify Power Supply Size (3.5A strongly recommended)
```
agners, line 80:
```suggestion
Check Power Supply Rating
```

RaptorPR443:
I am using the Argon one m.2 case with a kingston b&m key with the rpi 4 4gb and the argon 3.5a charger they sale. Been running it 24/7 for 6 days with HA 6.0 RC3 with 0 issues, no power failures.

The new rpi i purchased came with usb boot enabled from factory. i installed HA in m.2 ssd from rpi os from sd card then shut down and removed sd and automatically booted from usb without changing the factory bootloader.

PR #1400, opened 2021-06-12
pc/ova: support virtio-fs

andrew-d:
This enables support for [virtio-fs](https://virtio-fs.gitlab.io/), a shared filesystem for virtual machines. This allows very low-overhead file sharing between the host and guest when running a VM. I want to use this for sharing media files from Home Assistant to the host, where other programs/VMs can access them.

I tested this by running `make ova` locally to confirm that the Linux kernel built properly, which it did.


homeassistant:
Hi @andrew-d, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;operating-system#1400).

Thanks!

agners (approved):

pvizeli:
Don't forget that it's need also https://github.com/home-assistant/supervisor/issues/2564 otherwise if you doing it manual, it still end trouble

agners:
Yeah I was wondering that too: this is not a block device but a file system, so it doesn't work as root disk drive (like virtio disk). In other words, you have to explicitly mount such a file system. So this requires something like @pvizeli is mentioning.

That said, I don't think it hurts to enable the drive on OS side already to start experimenting/development etc.

andrew-d:
Yeah, my plan was to start out with a test add-on, running privileged, that could mount a `virtio-fs` filesystem from the host and e.g. copy files to/from the VM. If that works out well, then integrating with the supervisor or something else would be a good next step.

andrew-d:
@agners - I don't particularly mind when, but do you have a rough idea of when a version of HassOS will be released with these options? I need to do a VM migration at some point in the future, and if it'll be "soon", then I'll hold off; if not, I can do it now. Either is fine, so no pressure - just curious :smiley:

agners:
Another release 6 version is on my todo, probably towards the end of this week.

andrew-d:
> Another release 6 version is on my todo, probably towards the end of this week.

Thanks! Much appreciated :grinning:

deviantintegral:
Glad to see this added - this should be perfect for saving video streams outside of a VM image. @andrew-d if you end up starting an addon for this, please link it as I'd be glad to test it out.

PR #1412, opened 2021-06-14
Remove unnecessary ODROID-N2 kernel patches

agners:
For ODROID-N2, release 6 is using Linux 5.9.16. On dev branch we most
likely will move to latest stable and skip 5.10. Drop those unnecessary
patches.


agners:
This patches haven't been used in any build since release 6 is already using 5.10.42 (and the patches are in the version specific 5.10.37 directory).

PR #1418, opened 2021-06-16
Use compressed squashfs for kernel

agners:


pvizeli:
I still prefer LZ4 over LZO just the size is not that important as the speed

I think we can ditch in that case the ext4 module on uboot?

agners:
> I still prefer LZ4 over LZO just the size is not that important as the speed

Yes me too. But currently U-Boot doesn't support it :(

I have a preliminary patch, but not sure if that is acceptable. Since LZO works currently, I prefer to wait and see if it makes it upstream before switching to LZ4. With squashfs we can easily switch to a different compression algorithm, just adding a configuration and change the `mksquashfs` command.

> I think we can ditch in that case the ext4 module on uboot?

Not yet: If update fails, and U-Boot script falls back to the previous partition, it will need ext4 support. I think we need to have support for both during 1 release cycle or so, then we can get rid of it.

pvizeli (approved):

robloh:
> Not yet: If update fails, and U-Boot script falls back to the previous partition, it will need ext4 support. I think we need to have support for both during 1 release cycle or so, then we can get rid of it.

Will this also happen if someone upgrades for the first time 6 or 12 months from now (ie, skipping in-between versions)? If so then may I make the suggestion to keep the ext4 module for much longer?

agners:
> Will this also happen if someone upgrades for the first time 6 or 12 months from now (ie, skipping in-between versions)?

Yes.

> If so then may I make the suggestion to keep the ext4 module for much longer?

Probably make sense to keep it around, it doesn't hurt much. That said, skipping too many versions might not work for other reasons anyways. I think actually it would be a good idea if we disallow skipping a major version. This is what big distributions do too (e.g. from a Ubuntu LTS release, you can only update to the next safely, you can not skip a LTS version).

PR #1437, opened 2021-06-30
Using image name from version file

pvizeli:
Is needed for the new supervisor

### DIFF
--- a/buildroot-external/package/hassio/builder/hostapp.sh
+++ b/buildroot-external/package/hassio/builder/hostapp.sh
@@ -73,7 +73,9 @@ done
 
 # Install supervisor
 docker pull "${SUPERVISOR}:${SUPERVISOR_VERSION}"
-docker tag "${SUPERVISOR}:${SUPERVISOR_VERSION}" "${SUPERVISOR}:latest"
+
+# Need match with the tag used by OS
+docker tag "${SUPERVISOR}:${SUPERVISOR_VERSION}" "homeassistant/${ARCH}-hassio-supervisor:latest"
 
 # Install Plugins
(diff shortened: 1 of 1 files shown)


buildroot-external/package/hassio/builder/hostapp.sh
```diff
@@ -26,13 +26,13 @@ while [[ $# -gt 0 ]]; do
...
 done
 
-SUPERVISOR="homeassistant/${ARCH}-hassio-supervisor"
-DNS="homeassistant/${ARCH}-hassio-dns"
-AUDIO="homeassistant/${ARCH}-hassio-audio"
-CLI="homeassistant/${ARCH}-hassio-cli"
-MULTICAST="homeassistant/${ARCH}-hassio-multicast"
-OBSERVER="homeassistant/${ARCH}-hassio-observer"
-LANDINGPAGE="homeassistant/${MACHINE}-homeassistant:landingpage"
+SUPERVISOR=$(curl -s ${VERSION_URL} | jq -e -r '.images.supervisor' | sed -r "s/\{arch\}/${ARCH}/g")
```
ludeeus, line 29: Instead of doing 13 calls now with this, add a new variable that holds the result of `curl -s ${VERSION_URL}` and use that new variable everywhere `curl -s ${VERSION_URL}` is used

ludeeus (changes requested):

pvizeli:
It's not only that. We need also to rewrite the hassos-supervisor script etc. that he gets the right image. But then it's the question of how we deal with OTA updates. I think we need to look deeper into this after @agners vacation. It's not that simple but we need solve it for making new releases

PR #1445, opened 2021-07-05
Use git submodule for buildroot

agners:
This change drops buildroot from the operating system tree
and uses git submodule instead. This allows to manage changes
to buildroot in a (forked) buildroot git repository, which makes it
easier to manage the patches (rebase/prepare for upstream).


pvizeli:
https://github.com/buildroot/buildroot is not official, only a 3party mirror. We should look into how we can archive that from the official git repository. However, I'm fine with that

jens-maus:
> I prefer to use git instead of having to deal with patch files. I currently ended up managing a buildroot git repo with the HAOS patches locally, and regenerated the `buildroot-patches` directory from it, and then applying them to the `buildroot` directory. This all seems more complicated then necessary.

I would like to just provide my two cents here on that matter since I already partly discussed that with @agners in advance. Using `git submodules´ is  IMHO a bit over-the-top for the very few buildroot upstream patches that this project has to deal with. The mechanism of having/maintaining a top-level `buildroot-patches` directory with patch files is not only IMHO quite straight forward, it is also way more easily understandable for third-parties. In fact, in my side project (RaspberryMatic) I am doing exactly the same, here:

https://github.com/jens-maus/RaspberryMatic/tree/master/buildroot-patches

However, on top of that I do have a `create_patches.sh` skript which very quickly and reproducably creates these patches on the fly. In addition, I also do not maintain a top-level buildroot directory with the whole buildroot sources in that project, but download the official buildroot tarball upon the top-level make execution and then apply the patches before running the real buildroot build job.

IMHO the charm of this whole procedure still lies in its simplicity and that the patches or immediately visible and understandable for humans :) Moving all the logic into a dedicated buildroot repository and maintaining the patches there in dedicated branches seems to IMHO obfuscate the point that pre-patches are applied to the buildroot environment and that Home Assistant OS is not based on a vanilla buildroot environment.

But of course it is up to you to decide. Just want to drop my two cents here since I still believe that the current mechanism is capable enough while still maintaining a nice simplicity of things. The only thing that I would improve here would be to drop the top-level buildroot directoy with all the vanilla buildroot sources and make the top-level makefile download the official buildroot tarballs instead.

agners:
> Using git submodules is IMHO a bit over-the-top for the very few buildroot upstream patches that this project has to deal with.

I think the tendency will be that changes to buildroot will grow, or there will be more churn: I try to keep the diff small and  upstream changes acceptable upstream. But also during that process changes will accumulate. In the end, the more development we do on OS, the more the more churn will be in this patchset.

> However, on top of that I do have a create_patches.sh skript which very quickly and reproducably creates these patches on the fly. In addition, I also do not maintain a top-level buildroot directory with the whole buildroot sources [...]

I was curious how that `create_patches.sh` script looks like in the absence of a buildroot git tree. It seems that you rely on `.orig` files being there, and taking the diff from it? IMHO, that script is a bit a "poor man's git".

I get the argument that it is easier to tell the difference over vanilla buildroot by just having a glance at `buildroot-patches`. However, as soon as you want to dig into those changes a bit deeper, git is much more capable: a single `git log -p` on the changes branch gives the list of changes, and `git diff --stat 2021.02.3` a nice list of files with the amount of changes to each one of them.

> Furthermore, it also substantially increased the checked out directory/file size since the submodule will contain the whole buildroot history which is not at all necessary for this project.

Storage is cheap. A full checkout today is 211MiB. :sweat_smile: 

Not having the buildroot history is actually what I don't like about the current and your proposed approach: During development, I very often end up asking myself if there was a certain reason of some lines of code. `git log/blame <filename>` are very useful during development to get insights over the history of changes. The current approach only gives the update commit, and your approach would not give any history. Also having a full checkout allows to check the master branch quickly, to see if something has been fixed/changed which I am working on.

What we could do is an intermediate solution which allows a git tree if preferred:
  - Drop `buildroot` from the source tree
  - Make `Makefile` smart:
    - if `buildroot/.git` is missing, download the tarball and apply `buildroot-patches`
    - if `buildroot/.git` is present, assume the developer is maintaining the directory
  - Introduce `create_patches.sh` which does a `git format-patch --output-directory=../buildroot-patches <base>`

It has the downside that there are two "modes", and if somebody adds a patch to `buildroot-patches` every developer using the git method needs to manually reapply it to his branches...

> But of course it is up to you to decide. Just want to drop my two cents here [...]

Thanks for the feedback, it is very much appreciated! I see your points, but I think the benefits outweigh the downside, at least for how I use buildroot currently.

Let's give the `git submodule` approach a try, and see how it goes.

jens-maus:
> What we could do is an intermediate solution which allows a git tree if preferred:
> 
> * Drop `buildroot` from the source tree
> * Make `Makefile` smart:
>   
>   * if `buildroot/.git` is missing, download the tarball and apply `buildroot-patches`
>   * if `buildroot/.git` is present, assume the developer is maintaining the directory
> * Introduce `create_patches.sh` which does a `git format-patch --output-directory=../buildroot-patches <base>`
> 
> It has the downside that there are two "modes", and if somebody adds a patch to `buildroot-patches` every developer using the git method needs to manually reapply it to his branches...

To me this sounds like a good compromise between completly relying on `git submodules` only and handling these upstream patches in this "poor man's" style I still prefer ;-)

So I would suggest to go this road and modify the build system to perform exactly how you proposed, thus keeping "buildroot-patches" and just regenerate them from the new home-assistant/buildroot repository using `git format-patch`. And if a dev decided to use git submodules and checkout everything in `buildroot/.git` then the patches could be omitted or regenerated on the fly. 

> > But of course it is up to you to decide. Just want to drop my two cents here [...]
> 
> Thanks for the feedback, it is very much appreciated! I see your points, but I think the benefits outweigh the downside, at least for how I use buildroot currently.

I fully understand that and at least we can perfectly agree that dropping the top-level buildroot directory is a good idea. :) Nevertheless, I still believe that a git repository should be (if possible) self contained without the necessity of having to use git submodules. So if we could agree on your proposed procedure above I would really appreciate it because it would also make it easier for me to share those buildroot upstream patches between my main RaspberryMatic project and this project because the development road of these two otherwise independent projects are still quite linked.

agners:
I am still considering the two options download tarball vs. git submodule.

One of my concern with the download tarball is that it adds some complexity to the Makefile. Is your Makefile handling the case when new patches get added/removed?

I feel that this case is really better handled with `git submodule update`. Which makes me prefer it. The only downside I see is that two commits need to be made for every change: One to the buildroot repository, and one to bump the git submodule. However, it has the advantage that the buildroot changes can be better reviewed as those are direct changes (as opposed to patchfiles added to the tree). I probably would add a Github Action to bump the git submodule automatically.

jens-maus:
Ok, as you already merged your draft the only question that I still have is, why did you not fork https://github.com/buildroot/buildroot for https://github.com/home-assistant/buildroot ? Even thought the github repository of buildroot is just a mirror of the off-github version buildroot maintains, it would have made web-based GitHub comparisons more easy if it would have been created as a fork via GitHub. So can you perhaps redo this repository again so that it is visible as a fork and so GitHub would allow to perform more easy web-based comparisons between the fork and the original repository? Otherwise one has to always use the git command line to perform the comparisons to identify which patches you added.

PR #1451, opened 2021-07-09
Supoport compressed squashfs for kernel (#1418)

agners:
* Add squashfs with LZ4 and LZO compression to Barebox

* Add squashfs with LZO compression to U-Boot

* Use "raw" as type for kernel partition which is compatible with ext4
  image as well as squashfs.

Note: This does not use squashfs for the kernel but merly adds support
for it. This will allow to downgrade/fallback to the non-squashfs
version. Using squashfs for the kernel partition is implemented in dev
(for release 7).


agners:
This is mostly a backport from dev and tested to be working on a RPi 4 64-bit (a "failed" downgrade now can also boot a release 7).

PR #1452, opened 2021-07-09
Drop superseeded CRDA utility

agners:
The CRDA (Central Regulatory Domain Agent) utility has been used as a
user space helper to load regulatory information for WiFi drivers.
However, since Linux 4.15 the kernel can load the regulatory information
directly from a signed firmware file "regulatory.db".

The regulatory.db file is provided by the WIRELESS_REGDB package, which
has been already installed since its a dependency of CRDA.

Drop CRDA and select WIRELESS_REGDB package explicitly to make sure the
regulatory.db file is present.


agners:
See also:
https://www.kernel.org/doc/html/latest/networking/regulatory.html#how-to-get-regulatory-domains-to-the-kernel

Issue #1450, opened 2021-07-09
Current measurement for RPi PoE+ hat not working


georgeboot:
**Hardware Environment**

- [x] Raspberry Pi [1/2/3/4]
- [ ] ODROID [C2/C4/N2(+)/XU4]
- [ ] ASUS Tinker [S]
- [ ] Generic x86-64 (like Intel NUC)
- [ ] OVA (Open Virtualization Appliance, on Intel NUC or any other hardware, please add the Hypervisor you are using)

**Home Assistant OS release:**
- [x] Fresh installation of release 6.1
- [ ] Updated from version x.y

**Problem**
I can't measure the power the PoE+ hat provides through the OS.

It's supposed to be measured using ([source](https://www.jeffgeerling.com/blog/2021/review-raspberry-pis-poe-hat-june-2021)):

```
cat /sys/devices/platform/rpi-poe-power-supply@0/power_supply/rpi-poe/current_now
```

Perhaps there are some modules that need to be enabled in the kernel to support the new hat out of the box?

agners:
It seems that the PoE+ Hat support has been added with Raspberry Pi Kernel `1.20210527-1`. I'll update with the next release.

agners:
Release 6.2 is out, please test and reopen in case this is still a problem.

georgeboot:
@agners this still doesn't work on the latest version:

```
Logger: homeassistant.components.command_line
Source: components/command_line/__init__.py:48 
Integration: command_line (documentation, issues) 
First occurred: 15:55:35 (331 occurrences) 
Last logged: 21:25:34

Command failed (with return code 1): cat /sys/devices/platform/rpi-poe-power-supply@0/power_supply/rpi-poe/current_now
```

```yml
sensor:
- platform: command_line
  name: RPi CPU Temperature
  command: "cat /sys/class/thermal/thermal_zone0/temp"
  unit_of_measurement: "°C"
  value_template: "{{ value | multiply(0.001) | round(1) }}"
```

agners:
Please open a new issue with all the details provided.

Did you load the `rpi-poe-plus` overlay?

georgeboot:
A combination of setting the overlay to `rpi-poe-plus` (I was still using the old `rpi-poe` overlay) and reading from `/sys/devices/platform/rpi-poe-power-supply/power_supply/rpi-poe/current_now` (without the `@0`) instead, fixed it for me.

PR #1456, opened 2021-07-12
RaspberryPi: Update kernel 5.10.17 - 1.20210527-1 (#1450)

agners:
Git hash 081f560bde7188ca6f40cf297bd55c680e0179c0


jens-maus:
@agners Please note, that in RaspberryMatic I use the following `BR2_LINUX_KERNEL_CUSTOM_TARBALL_LOCATION` config option to directly reference the `1.20210527-1` release tag rather than using the commit id. IMHO this should be more intuitive:

```
BR2_LINUX_KERNEL_CUSTOM_TARBALL_LOCATION="https://github.com/raspberrypi/linux/archive/raspberrypi-kernel_1.20210527-1.tar.gz"
```

cf. https://github.com/jens-maus/RaspberryMatic/blob/master/buildroot-external/configs/raspmatic_rpi4_defconfig#L29

agners:
Definitely prefer tags too. I remember a while back implementing exactly that locally, but had to roll back for some reason. I just can't remember why I had to roll back :sweat_smile:. I'll try again, maybe that was some other problem or the problem is gone, let's see.

PR #1457, opened 2021-07-12
Update hassos-overlay

ukrolelo:
Feature request is on forum,everything is detailed explained.
Over 17K+ views on this issue.
We need to be able to copy fstab from overlay so the fstab settings on remote mount will stay after update/upgrade on squashfs.

https://community.home-assistant.io/t/nas-mount-on-hass-os/320218

### DIFF
--- a/buildroot-external/rootfs-overlay/usr/libexec/hassos-overlay
+++ b/buildroot-external/rootfs-overlay/usr/libexec/hassos-overlay
@@ -17,6 +17,11 @@ if [ ! -f /mnt/overlay/etc/hosts ]; then
     cp -fp /etc/hosts /mnt/overlay/etc/hosts
 fi
 
+if [ ! -f /mnt/overlay/etc/fstab ]; then
+    cp -fp /etc/fstab /mnt/overlay/etc/fstab
+fi
+
+
 # TimeSync
 if [ ! -f /mnt/overlay/etc/systemd/timesyncd.conf ]; then
     mkdir -p /mnt/overlay/etc/systemd


homeassistant:
Hi @ukrolelo, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;operating-system#1457).

Thanks!

ludeeus:
This should not be handled here, but with <https://github.com/home-assistant/supervisor/issues/2564>

ukrolelo:
> This should not be handled here, but with [home-assistant/supervisor#2564](https://github.com/home-assistant/supervisor/issues/2564)

Sorry i am new to this pull requests. So you are saying that this problem should be solved with different way?
As i see on supervisor issue 2564 it is still opened, i don't see any instruction how i could test it.To setup and set the mount options.
In my case, if fstab will be copied then the user will just add it to overlay and everything will be resolved.
But i do realise that devs see bigger picture and know which way to go.
What should i do :-P ?

ludeeus:
The supervisor should be responsible for mount/remount and making mounted drives available to core/add-ons, and the user should configure those with CLI/UI.
The reason #2564 is still open, is that no one has yet added that capability to the supervisor.

ukrolelo:
Thank you. Should/can i remove this pull or just leave it like this?

ludeeus:
I'll leave that decision up to @agners / @pvizeli

agners:
> Thank you. Should/can i remove this pull or just leave it like this?

So you aim to resolve it in Supervisor? I'll close this PR then.

ukrolelo:
Personaly, I think  for the time being that the fstab will be pushed on top of the squashfs would be a good workaround for thoose 16K+ views that in total is on the forum(people interested to resolve this thing).And until it will be resolved with supervisor (which will take long time, as it is GUI+backend, i don't have knowledge to do that).
P.S. if the commit that i did is actualy doing what i explained.

ukrolelo:
I have made a tutorial to do it manualy, but it breaks after update of HASS.Because of the new fstab file.
https://community.home-assistant.io/t/solved-hassos-mount-nas-network-share/303292

ludeeus:
"thoose 16K+ view" does not mean much, and is not a reason to add anything, your linked FR has 12 votes, if anything I personally count those 12 way higher than xxxx views to a topic.

But IMO workarounds should not be implemented because it's faster, it should be handled correctly, implementing workarounds means that we can't remove it, and there will not be a real incentive for a proper soulution.

ukrolelo:
I see, then please close. Thank you. Will watch the supervisor issue further.
<3

ukrolelo:
Wait guyz!!! one question, now i see that you have empty fstab in the folder with comment
fstab
Cleanup run
Is it possible to remove to do the cleanup? :D and everything will be fine afterwards? You are not using fstab anywere in the code as i see in search.

agners:
> Is it possible to remove to do the cleanup? :D and everything will be fine afterwards? You are not using fstab anywere in the code as i see in search.

That wouldn't help you since we bind mount files individually to overlay. Just removing it does not magically makes overlay work. You would have to add it to `buildroot-external/rootfs-overlay/usr/libexec/hassos-overlay` as well.

The problem is that if we add that now, we kinda have to support it even once we have the comfortable solution. I'd rather prefer to only support one way.

pvizeli:
I close this because it needs to be solved with the open RFC from the supervisor. Just a number of views don't allow to hack things into a system

ukrolelo:
Roger that, thx guyz. Will watch the supervisor RFC

ukrolelo:
@ludeeus 
Well now i have FR40, how many i need? :D
https://community.home-assistant.io/t/nas-mount-on-hass-os/320218/19
21.4K for hassio remote mount
2.4K for my tutorial on hassos mount

Would it be visible to just allow to not clean fstab? :d
maybe we can find a logic solution?
As i don't see any updates on fusefs remote mount on supervisor.


> "thoose 16K+ view" does not mean much, and is not a reason to add anything, your linked FR has 12 votes, if anything I personally count those 12 way higher than xxxx views to a topic.
> 
> But IMO workarounds should not be implemented because it's faster, it should be handled correctly, implementing workarounds means that we can't remove it, and there will not be a real incentive for a proper soulution.

ukrolelo:
@ludeeus @pvizeli 
FR 51

PR #1459, opened 2021-07-13
Improve Xen support

agners:


.github/workflows/pr-checks.yml
```diff
@@ -15,6 +15,6 @@ jobs:
       with:
         dockerfile: Dockerfile
     - name: Check shell scripts
-      uses: ludeeus/action-shellcheck@1.0.0
+      uses: ludeeus/action-shellcheck@1.1.0
       with:
-        ignore: "buildroot"
+        ignore: "buildroot" "buildroot-external/package/xe-guest-utilities/xe-linux-distribution"
```
jens-maus, line 20: FYI: When having implemented this Xen Support myself in RaspberryMatic I ran into an issue with the ludeeus/action-shellcheck action not accepting full path ignores like you did here. Perhaps @ludeeus can help, but it seems you can only specify directory path based ignores here. See here for my action errors:

https://github.com/jens-maus/RaspberryMatic/runs/3065555262?check_suite_focus=true#step:6:54

In fact, I solved that by using `ignore: buildroot-external/package/xe-guest-utilities` instead. However, this will of course ignore other files in this directory as well. Perhaps @ludeeus can enhance the action-shellcheck to accept full path ignores as well?

ludeeus: you can already do that, however... full path in a runner usually is not what you think it is

jens-maus: Well, all I can say is that the whole `"buildroot-external/package/xe-guest-utilities/xe-linux-distribution"` does not work when added as an `ignore:` path.

agners: But it seems to, this previously failed:

https://github.com/home-assistant/operating-system/runs/3059129944

Then I added that additional path.

jens-maus: Are you sure? Then what about that?
https://github.com/home-assistant/operating-system/actions/runs/1027389070/workflow

agners: Oh wow, why is this PR green then? I am confused!

agners: Its beyond me why GitHub does not show the fact in the main PR view that a workflow is not parable...

Anyways, I guess I'll ignore the directory too, thanks for bringing this up!

I hope I can get rid of the custom script once this is merged:
https://github.com/xenserver/xe-guest-utilities/pull/116

jens-maus: Ok, perfect. Even better would be if @ludeeus could change that `ignore:` line logic of his shellcheck action so that even a full path to a file (and not only subdir) can be used. Or did I fundamentally miss here something why this is not possible?!?

PR #1464, opened 2021-07-14
Support development builds in pull-requests

agners:
This adds support for development builds in pull-requests. This allows to test a change before merging on actual hardware.

A PR needs to be labeled with `run-dev-build` as well as the board labels the build should be done for.

The development build get a custom version number with the PR number prepended (e.g. `7.0.dev2021071400123`).


.github/workflows/dev.yml
```diff
@@ -4,15 +4,18 @@ name: Development build
 
 on:
   workflow_dispatch:
+  pull_request:
```
ludeeus, line 7: This will fail when the pull request is opened from a fork

agners: This is from a fork... Seems to work?

ludeeus: Forks does not have access to secrets

agners: I see, but using `pull_request_target` should work right? I just can't test it until merged :)

agners:
```suggestion
  pull_request_target:
````

ludeeus: it does, use with care <https://securitylab.github.com/research/github-actions-preventing-pwn-requests/>

PR #1465, opened 2021-07-15
Support development builds in PRs

agners:
Implement a "/ghaction build" command to trigger a development build.
The comment must come from the owner or member of the organization.
This will only generate a build for the machines the PR has been
labeled with.

This is an alternative to #1464.


agners:
This implementation has the downside that the build is not associated with the PR really.

agners:
Close in favor of #1464.

PR #1474, opened 2021-07-26
board: hardkernel: Fix random crash on Linux kernel 5.10 and newer

agners:
Avoid adding memreserve since the device tree already reserves the
memory regions in reserved-memory nodes in device tree. This avoids
conflicting no-map setting and makes sure memory is properly reserved.


agners:
It is still unclear how the issue will be fixed upstream it could be a change similar to this in U-Boot or a revert of the Linux patch as done in #1476.

Close in favor of a work-around in Linux. The work around in Linux also works in case of a OTA fallback.

Issue #1448, opened 2021-07-08
lvm2-monitor.service fails to start


agners:
**Hardware Environment**

- [ ] Raspberry Pi [1/2/3/4]
- [ ] ODROID [C2/C4/N2(+)/XU4]
- [ ] ASUS Tinker [S]
- [ ] Generic x86-64 (like Intel NUC)
- [ ] OVA (Open Virtualization Appliance, on Intel NUC or any other hardware, please add the Hypervisor you are using)
- [x] any

**Home Assistant OS release:**
- [ ] Fresh installation of release x.y
- [ ] Updated from version x.y
- Additional information (if accessible):

**Supervisor logs:**

**Journal logs:**

**Kernel logs:**

**Description of problem:**

It seems that since OS release 6.1 the lvm2-monitor.service doesn't start.
```
# systemctl status lvm2-monitor.service
● lvm2-monitor.service - Monitoring of LVM2 mirrors, snapshots etc. using dmeventd or progress polling
     Loaded: loaded (/usr/lib/systemd/system/lvm2-monitor.service; enabled; vendor preset: enabled)
     Active: failed (Result: exit-code) since Tue 2021-06-22 08:15:04 UTC; 2 weeks 2 days ago
       Docs: man:dmeventd(8)
             man:lvcreate(8)
             man:lvchange(8)
             man:vgchange(8)
    Process: 163 ExecStart=/usr/sbin/lvm vgchange --monitor y --ignoreskippedcluster (code=exited, status=203/EXEC)
   Main PID: 163 (code=exited, status=203/EXEC)
```

agners:
It seems that LVM2 is not installed completely, despite being a dependency. There is already a fix in latest 2020.02.x branch, hence with the next Buildroot update this should be resolved:
https://git.busybox.net/buildroot/commit/?id=6d758f59e65c33b067076cdb89a8404d5471c7ae

That said, it seems that LVM is pulled in by `libblockdev` which is required by `udisks`. But `libblockdev` only needs device-mapper capablitiy of the LVM package. It is worth considering opting out of `BR2_PACKAGE_LVM2_STANDARD_INSTALL` to get rid of much of lvm, which we don't need anyways.

PR #1481, opened 2021-07-28
buildroot: package/lvm2: install dmsetup only (#1448)

agners:
LVM2 is not really required in the embedded use case. Opt out of
installing the standard installation which will install only dmsetup.
This requires a backported fix for the lvm2 package to not install
unnecessary systemd services.

Fixes: #1448
Related: #1449


agners:
This is basically a backport of #1449.

PR #1517, opened 2021-08-28
updated eq3_char_loop and rpi-rf-mod package to latest versions

jens-maus:
This PR updates the eq3_char_loop and rpi-rf-mod buildroot packages to the latest available versions and adds also proper hash files for download verification.


buildroot-external/package/rpi-rf-mod/rpi-rf-mod.mk
```diff
@@ -11,15 +11,16 @@
...
 RPI_RF_MOD_SITE = $(call github,jens-maus,RaspberryMatic,$(RPI_RF_MOD_VERSION))
 RPI_RF_MOD_LICENSE = Apache-2.0
 RPI_RF_MOD_DEPENDENCIES = host-dtc
-#RPI_RF_MOD_LICENSE_FILES = LICENSE
+RPI_RF_MOD_LICENSE_FILES = LICENSE
 
 ifeq ($(BR2_PACKAGE_RPI_RF_MOD_DTS_RPI),y)
   # RaspberryPi DTS file
   RPI_RF_MOD_DTS_FILE = rpi-rf-mod
+  RPI_RF_MOD_DTS_FILE_ALT = rpi-rf-mod-rpi1
```
agners, line 23: Maybe this should be `RPI_RF_MOD_DTS_FILES` and then a for loop below? I am fine with this for now.

jens-maus: You are right, this could of course be a bit more optimized. will have a look for it with the next update. However, if you look at the use case in the `if` calls below you will see that using `RPI_RF_MOD_DTS_FILE` is handled different than `RPI_RF_MOD_DTS_FILE_ALT` in regard of the output name of the dtbo.

PR #1543, opened 2021-09-08
Update docs hybrid >> hyprid to match the implementation

adamfeldman:
https://github.com/home-assistant/operating-system/search?q=hyprid&type=code

### DIFF
--- a/Documentation/development.md
+++ b/Documentation/development.md
@@ -5,7 +5,7 @@
 `BOOT_SYS`:
 
 - efi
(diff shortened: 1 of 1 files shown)


agners:
`hybrid` is the correct writing afaik. I'd rather prefer to fix the naming in the implementation.

agners:
Implemented in #1548

adamfeldman:
Thank you @agners!

PR #1555, opened 2021-09-27
Add NVMe boot support for RPi Compute Module 4

agners:


agners:
It seems that U-Boot 2021.04 has another bug which causes the script to fail quite often

```
Device 0: Vendor: 0x144d Rev: 1B2QEXM7 Prod: S4EWNF0M531930D
            Type: Hard Disk
            Capacity: 953869.7 MB = 931.5 GB (1953525168 x 512)
... is now current device
Scanning nvme 0:1...
Found U-Boot script /boot.scr
2411 bytes read in 0 ms
## Executing script at 02400000
Bad data crc
SCRIPT FAILED: continuing...
```

Manually running the boot command using `run bootcmd_nvme0` then works fine. I'll check with U-Boot 2021.10-rc4 and consider moving to it if it turns out to work better.

agners:
This turned out to be a bug in U-Boot which is also present in the latest version. This patch fixes it:
https://lists.denx.de/pipermail/u-boot/2021-September/461882.html

Since the update to v2021.10-rc5 is done now, this PR is obsolete and gets replaced by #1557.

PR #1569, opened 2021-10-04
Add Amber as new machine

agners:
This adds initial Home Assistant Amber support. The change
introduces a new machine `amber` in the `raspberrypi` board
directory structure.

Currently the board introduces patches for the custom device
tree.


jens-maus:
Please note that enabling the serial console per default might conflict with onboard UART uses like using a RPI-RF-MOD/HM-MOD-RPI-PCB. So if you keep it enabled per default this might imply a different manual setup to get the GPIO board UART free for these and other modules.

agners:
> Please note that enabling the serial console per default might conflict with onboard UART uses like using a RPI-RF-MOD/HM-MOD-RPI-PCB. So if you keep it enabled per default this might imply a different manual setup to get the GPIO board UART free for these and other modules.

Yes I am aware. For now I'd like to enable `enable_uart=1` since it allows to get debug logs from the boot firmware. I intend to disable that again later.

UART5 is connected to the on-board CP2102N USB to UART converter. Ultimately it will be the UART where a console is enabled by default.

PR #1573, opened 2021-10-06
Replace Stale Bot with Stale GitHub Action

agners:


.github/workflows/stale.yml
```diff
@@ -0,0 +1,47 @@
...
+      # The 90 day stale policy
+      # Used for:
+      # - Issues & PRs
+      # - No PRs marked as no-stale or pinned
+      # - No issues marked as no-stale, help-wanted or pinned
+      - name: 90 days stale issues & PRs policy
+        uses: actions/stale@v4
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          days-before-stale: 90
```
ludeeus, line 22: This was 60 before

.github/workflows/stale.yml
```diff
@@ -0,0 +1,47 @@
...
+          days-before-close: 7
+          operations-per-run: 150
+          remove-stale-when-updated: true
+          stale-issue-label: "stale"
```
ludeeus, line 26: This was "wontfix" before.
All applicable issues with that label need to get the "stale" as well
https://github.com/home-assistant/operating-system/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3Awontfix

ludeeus (approved):

.github/workflows/stale.yml
agners: Yes, I moved to 90 days on purpose as 60 days go by fast :) Also aligns with Core.

.github/workflows/stale.yml
agners: Yes, and I like to move to the `stale` label. I think `wontfix` really has a different meaning, it assumes that there is an issue and we won't address it. In my mind, stale is really not about not fixing the bug, its more about the assumption that the bug fixed somehow differently by updating (other) software.

I guess they will get the `stale` label the first time this action runs. I am considering removing the `wontfix` label though, so its clear that those are just `stale` issues.

PR #1612, opened 2021-10-26
Add support for Btrfs

annuges:
Add support for Btrfs file system to the kernel to allow host os and addons to mount external disks formatted this way


buildroot-external/kernel/hassos.config
```diff
@@ -15,6 +15,7 @@ CONFIG_VFAT_FS=y
 CONFIG_SQUASHFS=y
 CONFIG_SQUASHFS_XATTR=y
 CONFIG_SQUASHFS_LZ4=y
+CONFIG_BTRFS_FS=y
```
pvizeli, line 18: Why not just as a module? On embedded systems, ZFS or BTRFS is quite useless or just over-engineered @agners ?

PR #1625, opened 2021-11-02
Support Realtek 8152/8153/8156 USB Ethernet adapter (#1604)

agners:


jarkastr:
i have a tp link UE305, which appears to use a realtek 8153 chipset. Tried installing home assistant OS 7.0 and it would not recognize the usb ethernet adapter. I followed the troubleshooting in this link: https://bbs.archlinux.org/viewtopic.php?id=232764 and i believe it is being recognized as an HID device. 

I am at about the limits of my troubleshooting abilities, so happy to run whatever you may need to fix/verify.

agners:
@jarkastr ugh, that devices seems to need some additional user space tool to change the mode of the device to get it working. I'd rather prefer not to support such complicated devices.

jarkastr:
@agners Is there a list of support USB ethernet adapters somewhere? I'll buy another one if necessary

jarkastr:
@agners I apologize, I thought I was running 7.0, but I was running 6.6. I ended up going with the weird proprietary Ethernet adapter for this laptop with an Intel driver. Happy to test the USB Ethernet adapter in a later version if you need a tester.

agners:
> I ended up going with the weird proprietary Ethernet adapter for this laptop with an Intel driver.

So that adapter worked then?

> Happy to test the USB Ethernet adapter in a later version if you need a tester.

You mean the TP Link UE305? Ok, I'll keep it in mind.

jarkastr:
> > I ended up going with the weird proprietary Ethernet adapter for this laptop with an Intel driver.
> 
> So that adapter worked then?

Yes, proprietary ThinkPad Ethernet adapter worked. Pretty sure there is not a chipset in the adapter and it is just a physical connection to the motherboard because not enough physical space on laptops these days. 

> > Happy to test the USB Ethernet adapter in a later version if you need a tester.
> 
> You mean the TP Link UE305? Ok, I'll keep it in mind.

Yes, the TP Link UE305 worked today when I tested it on 7.0

PR #1652, opened 2021-11-16
Updated to lastest generic_raw_uart version

alexreinert:


agners:
Should this go into release 7 still? If so, I'll add a `REL-7` label and backport it before the next release candidate.

alexreinert:
If there is no change freeze, this would be nice.

jens-maus:
I can approve this PR and flag it REL-7 if this is fine with you @agners

alexreinert:
It seems, that there is an issue on RPI3 boards. I will do some further investigations and provide a new PR, when I fixed it.

PR #1653, opened 2021-11-21
Updated to latest generic_raw_uart package

alexreinert:


jens-maus (approved):
After having checked that this version works now fine, I can approve it for integration in `rel-7`.

However please check @agners that it is fine that @alexreinert has sent in this PR based on the :rel-7 branch, because I think PR should always be submitted to the :dev branch, right?

agners:
> However please check @agners that it is fine that @alexreinert has sent in this PR based on the :rel-7 branch, because I think PR should always be submitted to the :dev branch, right?

Yes this needs to be against `dev` branch. Changed accordingly.

Issue #826, opened 2020-08-19
Invalid OVF checksum algorithm: SHA256


JohnGauthier:
**HassOS release with the issue:**
4.12 OVA, downloaded 8/19/2020

**Description of problem:**
VMware vCenter 7 (vSphere Client version 7.0.0.10600) error on deploying OVF template:

> _**The provided manifest file is invalid: Invalid OVF checksum algorithm: SHA256 .**_

vicfergar:
I have followed [this steps](https://medium.com/@rootedshell/failed-to-open-virtual-machine-sha1-digest-of-file-xxxxxxxx779-disk1-vmdk-does-not-match-manifest-7d0d94db4451) and it works for me :wink:

Apocrathia:
Deleting the manifest file didn't work. Now getting the following error:

> Issues detected with selected template. Details: - 71:7:VALUE_ILLEGAL: Value ''3'' of Parent element does not refer to a ref of type DiskControllerReference.

This issue is still present in 4.19 OVA.

DennisFaucher:
Same issue 

Issues detected with selected template. Details: - 71:7:VALUE_ILLEGAL: Value ''3'' of Parent element does not refer to a ref of type DiskControllerReference.

B3DTech:
So the OVA that Home Assistant releases was designed for VMWare Player, or Workstation, or a consumer version.  On ESXi/vSphere, it has no concept of a SATA controller.  Only IDE or SCSI.  Extract the OVF and VMDK.  Change the controller section starting at line 55 to:

      <Item>
        <rasd:Address>0</rasd:Address>
        <rasd:Caption>SCSIController</rasd:Caption>
        <rasd:Description>SCSI Controller</rasd:Description>
        <rasd:ElementName>SCSIController</rasd:ElementName>
        <rasd:InstanceID>3</rasd:InstanceID>
        <rasd:ResourceSubType>lsilogic</rasd:ResourceSubType>
        <rasd:ResourceType>6</rasd:ResourceType>
      </Item>

Then you can either repackage the OVA, or use the OVF/VMDK to create your VM.

DennisFaucher:
Thank you. I will try this and report back.

DennisFaucher:
Same error (
Selected OVA file is not a valid template.) . Maybe I did not package the OVA properly. I unzipped it, changed that section and re-zipped. Attached changed file. (Added .TXT to the name)
[...]

PR #1674, opened 2021-12-14
Adjust ova to support VMware ESXi (#826)

agners:
Use OpenSSL to generate OVA manifest file. It seems that sha256sum adds a
space after the hash algorithm which causes "Invalid OVF checksum algorithm"
on certain VMware virtualization products.

Using OpenSSL avoids the space and makes the manifest file compatible
with VMware products.

Use the common LSI Logic SCSI controller for disks. This is supported by Virtual Box
as well as VMware.

### DIFF
--- a/buildroot-external/board/pc/ova/hassos-hook.sh
+++ b/buildroot-external/board/pc/ova/hassos-hook.sh
@@ -32,7 +32,7 @@ function hassos_post_image() {
 
     cp -a "${BOARD_DIR}/home-assistant.ovf" "${OVA_DATA}/home-assistant.ovf"
     qemu-img convert -O vmdk -o subformat=streamOptimized "${HDD_IMG}" "${OVA_DATA}/home-assistant.vmdk"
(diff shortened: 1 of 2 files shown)


buildroot-external/board/pc/ova/hassos-hook.sh
```diff
@@ -32,7 +32,7 @@ function hassos_post_image() {
 
     cp -a "${BOARD_DIR}/home-assistant.ovf" "${OVA_DATA}/home-assistant.ovf"
     qemu-img convert -O vmdk -o subformat=streamOptimized "${HDD_IMG}" "${OVA_DATA}/home-assistant.vmdk"
-    (cd "${OVA_DATA}" || exit 1; sha256sum --tag home-assistant.* >home-assistant.mf)
+    (cd "${OVA_DATA}" || exit 1; openssl sha256 home-assistant.* >home-assistant.mf)
```
jens-maus, line 35: IMHO this should better be

```
"${HOST_DIR}/bin/openssl" sha256
```

so that the openssl version build during the buildroot run is used and not the openssl binary installed in the system itself.

pvizeli (approved):

PR #1681, opened 2021-12-18
Enable ForwardToSyslog in journald

mib1185:
This will enable `ForwardToSyslog` in journald, so that all log messages become available via `/dev/log` socket.
So we get the possibility to have addons (_maybe core or user_) to gather logs from this socket and - for example - forward them to a remote syslog server or do some kind of monitoring and alerting based on defined search pattern...

### DIFF
--- a/buildroot-external/rootfs-overlay/etc/systemd/journald.conf
+++ b/buildroot-external/rootfs-overlay/etc/systemd/journald.conf
@@ -1,3 +1,4 @@
 [Journal]
 Storage=auto
 Compress=yes
+ForwardToSyslog=yes


agners:
> This will enable ForwardToSyslog in journald, so that all log messages become available via /dev/log socket.

I don't think that this will work just like that. `/dev/log` is a symlink to `/run/systemd/journal/dev-log`, which in turn is only available on the host OS. Inside containers (where Add-ons live) it won't be accessible.

That said, recently we exposed the systemd-journal-gatewayd service to the Supervisor (see #1627). This socket allows to access messages in a more structured manor (see the [systemd-journal-gatewayd man](https://www.freedesktop.org/software/systemd/man/systemd-journal-gatewayd.service.html) page). We could consider making this available to Add-ons (via a flag in the [Add-on configuration](https://developers.home-assistant.io/docs/add-ons/configuration)).

mib1185:
> I don't think that this will work just like that. `/dev/log` is a symlink to `/run/systemd/journal/dev-log`, which in turn is only available on the host OS. Inside containers (where Add-ons live) it won't be accessible.

yeah sorry, i was not precise enough on my initial post - when i say to make it available inside the AddOns, I mean to map the `/dev/log` readonly into the container (eq. `-v /dev/log:/dev/log:ro`)

> That said, recently we exposed the systemd-journal-gatewayd service to the Supervisor (see #1627). This socket allows to access messages in a more structured manor (see the [systemd-journal-gatewayd man](https://www.freedesktop.org/software/systemd/man/systemd-journal-gatewayd.service.html) page). We could consider making this available to Add-ons (via a flag in the [Add-on configuration](https://developers.home-assistant.io/docs/add-ons/configuration)).

is it intended to add a "send to remote syslog" feature to the supervisor based on the `systemd-journal-gatewayd` ? 🤔

mib1185:
Hi @agners do you think, we can go ahead with this, or is there meanwhile another solution available?

agners:
I don't think this will work as you expect it. Currently we map `/dev/` into add-ons already. We use cgroups to adjust device access rights, but mapping the whole `/dev` from the host to each addon solves a lot of headache with udev (this way the host udev creates symlinks e.g. for serial deviecs and all addons have them instantly).

However, this also means that each addon gets `/dev/log` from the host. And the host `/dev/log` is a symlink. E.g. this is from within the SSH add-on:
```
[core-ssh ~]$ ls -l /dev/log
lrwxrwxrwx    1 root     root            28 Feb 12 01:11 /dev/log -> /run/systemd/journal/dev-log
[core-ssh ~]$ ls -la /run/systemd/journal/dev-log
ls: /run/systemd/journal/dev-log: No such file or directory
```

Also, it seems to me that this is somewhat circular: `/dev/log` is journald itself, so does `ForwardToSyslog` make systemd forward to itself? I guess systemd-journald is smart enough to handle this properly.

But for this to work in addons we would also have to bind mount `/run/systemd/journal/dev-log` into the add-on container.

We do have an option which bind mounts the systemd-journald database into an addon.

@mdegat01 made add-ons which uses Promtail to forward logs to a Loki instance. Do those add-ons maybe fulfill your use case?

mdegat01:
For reference, the promtail addon is [here](https://github.com/mdegat01/addon-promtail). I have a [loki addon](https://github.com/mdegat01/addon-loki) in the same repo but promtail can ship the logs anywhere, doesn't have to be to that addon. If you prefer a different log shipper one could be built in the similar way, just enable the `journald` option in the addon's configuration so it see the system journal and read from it.

If that doesn't work for you then perhaps you could explain a bit more about what you're trying to do? Because as agners noted, there's more steps here. This change alone isn't going to affect addons. Addons still will not have access to the syslog as there is no way to get `/run/systemd/journal/dev-log` mapped into an addon.

> We could consider making this available to Add-ons (via a flag in the [Add-on configuration](https://developers.home-assistant.io/docs/add-ons/configuration)).

That sounds interesting, maybe a new `service` addons request from supervisor?

mib1185:
Sorry for very late response, but I got it managed to extract the logs from journald files and pull them over to a remote syslog server --> [ha-addon-syslog](https://github.com/mib1185/ha-addon-syslog) ... it is not perfect, yet but it already works :)

PR #1684, opened 2021-12-19
Devices description update

lszeremeta:
I noticed that the device list is a bit outdated. I updated it according to: https://www.home-assistant.io/installation/

Additionally, [it looks like the Raspberry Pi Compute Module 4 is supported](https://community.home-assistant.io/t/ha-on-raspberry-pi-compute-modules/237182/10). Maybe it is worth adding some information about it here and on the website?

### DIFF
--- a/Documentation/boards/README.md
+++ b/Documentation/boards/README.md
@@ -20,7 +20,7 @@ The following boards/devices are supported:
   - Odroid-XU4
 - Asus
   - Tinker Board
-- Intel NUC
+- Generic x86-64 (UEFI, not suited for virtualization)
   - Intel NUC5CPYH
   - Intel NUC6CAYH
   - Intel NUC10I3FNK2
--- a/Documentation/boards/README.md
+++ b/Documentation/boards/README.md
@@ -5,26 +5,27 @@
 The following boards/devices are supported:
 
 - Raspberry Pi
-  - Pi 4 Model B (1 GB, 2 GB and 4 GB model) 32-bit
-  - Pi 4 Model B (1 GB, 2 GB and 4 GB model) 64-bit (recommended)
-  - Pi 3 Model B and B+ 32-bit
+  - Pi 4 Model B (1 GB, 2 GB, 4 GB and 8 GB model) 64-bit (recommended)
+  - Pi 4 Model B (1 GB, 2 GB, 4 GB and 8 GB model) 32-bit
   - Pi 3 Model B and B+ 64-bit (recommended)
+  - Pi 3 Model B and B+ 32-bit
   - Pi 2 (not recommended)
   - Pi Zero-W (not recommended)
   - Pi (not recommended)
 - Hardkernel
-  - Odroid-C2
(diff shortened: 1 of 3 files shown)


homeassistant:
Hi @lszeremeta, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;operating-system#1684).

Thanks!

Documentation/boards/README.md
```diff
@@ -39,17 +40,17 @@ Notes:
...
-|Generic x86-64 |`make generic_x86_64` |[generic_x86_64](../../buildroot-external/configs/generic_x86_64_defconfig)|[generic-x86-64](./generic-x86-64/)|
-|OVA            |`make ova`            |[ova](../../buildroot-external/configs/ova_defconfig)|[ova](./ova/)|
+|Pi4B 64-bit   |`make rpi4_64`       |[rpi4_64](../../buildroot-external/configs/rpi4_64_defconfig)|[raspberrypi](./raspberrypi/)|
+|Pi4B 32-bit   |`make rpi4`          |[rpi4](../../buildroot-external/configs/rpi4_defconfig)|[raspberrypi](./raspberrypi/)|
+|Pi3B 64-bit   |`make rpi3_64`       |[rpi3_64](../../buildroot-external/configs/rpi3_64_defconfig)|[raspberrypi](./raspberrypi/)|
+|Pi3B 32-bit   |`make rpi3`          |[rpi3](../../buildroot-external/configs/rpi3_defconfig)|[raspberrypi](./raspberrypi/)|
+|Pi2           |`make rpi2`          |[rpi2](../../buildroot-external/configs/rpi2_defconfig)|[raspberrypi](./raspberrypi/)|
+|Pi Zero       |`make rpi0_w`        |[rpi0_w](../../buildroot-external/configs/rpi0_w_defconfig)|[raspberrypi](./raspberrypi/)|
+|Pi            |`make rpi`           |[rpi](../../buildroot-external/configs/rpi_defconfig)|[raspberrypi](./raspberrypi/)|
```
agners, line 49: Pi Zero and Pi configurations actually got recently dropped, we should remove them from this table too.

agners (changes requested):
Thanks for the update, one small nit.

Documentation/boards/README.md
lszeremeta: Done. I removed Pi Zero and Pi from the tables. In addition I removed note for Raspberry Pi 4 (all RAM versions are supported so it's not needed). What about Raspberry Pi Compute Module 4? It's officially supported or not? No info about that on the [Installation docs](https://www.home-assistant.io/installation/raspberrypi).

agners (approved):

Documentation/boards/README.md
agners: > What about Raspberry Pi Compute Module 4?

It should work with the rpi4 image and is generally supported. That said, we don't have instructions etc. Then also different IO board might need a different device tree...

PR #1695, opened 2021-12-27
Use LSI Logic SCSI controller

agners:
For some reason, the vmdk disk format's descriptor contains the
controller type as well. By default, qemu-img sets it to "ide", which
seems not optimal especially for VMware's ESXi. Set adapter type to
commonly supported "lsilogic".


agners:
Note that I haven't actual prove that this benefits VMware Workstation/ESXi installations etc. as there seem to be also an option at installation time. But setting it is recommended it forums/blogs (e.g. [here](https://blogs.vmware.com/openstack/2017/07/)).

PR #1713, opened 2022-01-13
Linux: Update kernel 5.10.91

agners:


healeydave:
Any idea if there's anything in this kernel update to address ffmpeg hwaccel issues on Rpi?

agners:
This kernel update doesn't affect Raspberry Pi (see labels). For Raspberry Pi, we don't use upstream kernels, but the Raspberry Pi teams kernel https://github.com/raspberrypi/linux/. I usually update the Raspberry Pi kernel when they tag a new version.

PR #1757, opened 2022-02-21
Add support for AArch64/ARM64 EFI architecture

Doridian:
This uses GRUB because Barebox does not support ARM64 and the only supported bootloader by BR2 seems to be GRUB for this scenario.

I had to use `qemuarm-64` in several places as machine type as `generic-aarch64` versions of some containers (namely supervisor) do not exist and/or work correctly.

I have done basic functionality testing with `rauc`
- The current partition gets marked as good, HA seems to start up correctly
- I can swap partition to B, but of course as I cannot create update packages it does not boot, but it does get selected and then (correctly) fail to boot

Here is my test script I use (the commented out lines are required before the first boot to create the necessary flash0 and flash1 img files)
[testrun.sh.txt](https://github.com/home-assistant/operating-system/files/8111216/testrun.sh.txt)


homeassistant:
Hi @Doridian, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;operating-system#1757).

Thanks!

.github/workflows/matrix.json
```diff
@@ -11,6 +11,12 @@
     "runner": "x86-64-runner",
     "label": "board/generic-x86-64"
   },
+  {
+    "id": "generic-aarch64",
+    "defconfig": "generic_aarch64",
+    "runner": "aarch64-64-runner",
```
agners, line 17:
```suggestion
   "runner": "aarch64-runner",
```

agners:
Hm, there is a limitation in our PR dev build infrastructure: We can only build for machines which already exist in the `dev` branch (since that is where the metadata are coming from to create the build matrix).

Anyways, I will create a build locally, should be a quick thing.

agners:
@Doridian uploaded the builds to the root of https://os-builds.home-assistant.io/.

Can you share the qemu command line options you are using to test the images?

Doridian:
@agners I attached my QEMU test script to the initial post, completely forgot to do that initially.

agners:
It seems that installing the raucb currently fails:

```sh
rauc install haos_generic-aarch64-8.0.dev20220221.raucb
...
LastError: Installation error: Failed updating slot boot.0: failed to run slot hook: Child process exited with code 1
```

We assume that there is a txt file in the root and back it up. Because there is no txt, the rauc hook `buildroot-external/ota/rauc-hook` fails. You can work around by just adding a `cmdline.txt`:

```sh
touch /mnt/boot/cmdline.txt
```

Which brings up another question: Can we add that file (can be empty for now) and make its content append to the kernel boot arguments? Might be not really Grub standard of doing things, but would align nicely with the other platforms.

agners (changes requested):
It seems that a barebox file gets copied to /etc currently. Can you adjust the if statement at `buildroot-external/scripts/rauc.sh:install_bootloader_config` as well to accommodate the new `BOOTLOADER` value (just check explicitly for barebox).

buildroot-external/board/generic-aarch64/grub.cfg
```diff
@@ -0,0 +1,61 @@
...
+CMDLINE="rootwait zram.enabled=1 zram.num_devices=3 apparmor=1 security=apparmor rootwait systemd.machine_id=$MACHINE_ID cgroup_enable=memory fsck.repair=yes"
+
+# root is a full HDD/partition definition in GRUB format like hd0,gpt1
+# We extract the part before the comma to then append our own partition index
+# at the end. This is hacky but the best way I found
+regexp --set 1:boothd (.+),.+ ${root}
+
+menuentry "Slot DEFAULT (for RAUC to fall back onto)" {
+    linux (${boothd},gpt2)/Image root=PARTUUID=8d3d53e3-6d49-4c38-8349-aff6859e82fd $CMDLINE rauc.slot=A
+}
```
agners, line 53: I guess technically that is not even necessary is it? It would fallback to the first entry which is A as well..?

Doridian:
@agners
For your points:

The third boot option:
I have adjusted the GRUB config to allow removing the third boot option (it was initially necessary due to the RAUC config template necessitating a difference between "unselected, default to A" and "selected, use A" profiles, but i circumnavigated that with a default of 99, which is surely invalid and can be used as a flag)

The cmdline.txt
I have not found a way in GRUB to read an arbitrary file into a variable at all. I can "source" a file, I can load the env file, and I can exec a file. But I have not seen a way to load a plain text file
Therefor I have made the cmdline.txt that format (`set CMDLINE="...."`), so it can be updated as all the other versions do, but with a slightly modified format.

buildroot-external/board/generic-aarch64/cmdline.txt
```diff
@@ -0,0 +1 @@
+set CMDLINE="rootwait zram.enabled=1 zram.num_devices=3 apparmor=1 security=apparmor systemd.machine_id=$MACHINE_ID cgroup_enable=memory fsck.repair=yes"
```
agners, line 1: This should only contain boot arguments which the user should adjust. For now I would leave it empty. Also, I think set is not strictly necessary correct? Can this be:

```suggestion
cmdline=""
```

buildroot-external/board/generic-aarch64/grub.cfg
```diff
@@ -0,0 +1,58 @@
...
+    fi
+    if [ "$B_OK" -eq 1 -a "$B_TRY" -eq 1 ]; then
+        B_TRY=0
+    fi
+    default=0
+fi
+
+save_env A_TRY A_OK B_TRY B_OK ORDER MACHINE_ID
+
+source ($root)/cmdline.txt
```
agners, line 45: I then would add the default args here:

```suggestion
default_cmdline="rootwait zram.enabled=1 zram.num_devices=3 apparmor=1 security=apparmor systemd.machine_id=$MACHINE_ID cgroup_enable=memory fsck.repair=yes"
source ($root)/cmdline.txt
```

buildroot-external/board/generic-aarch64/grub.cfg
```diff
@@ -0,0 +1,58 @@
...
+
+# root is a full HDD/partition definition in GRUB format like hd0,gpt1
+# We extract the part before the comma to then append our own partition index
+# at the end. This is hacky but the best way I found
+regexp --set 1:boothd (.+),.+ ${root}
+
+menuentry "Slot A (OK=$A_OK TRY=$A_TRY)" {
+    linux (${boothd},gpt2)/Image root=PARTUUID=8d3d53e3-6d49-4c38-8349-aff6859e82fd $CMDLINE rauc.slot=A
```
agners, line 53:
```suggestion
    linux (${boothd},gpt2)/Image root=PARTUUID=8d3d53e3-6d49-4c38-8349-aff6859e82fd $default_cmdline $cmdline rauc.slot=A
```

buildroot-external/board/generic-aarch64/grub.cfg
```diff
@@ -0,0 +1,58 @@
...
+}
+
+menuentry "Slot B (OK=$B_OK TRY=$B_TRY)" {
+    linux (${boothd},gpt4)/Image root=PARTUUID=a3ec664e-32ce-4665-95ea-7ae90ce9aa20 $CMDLINE rauc.slot=B
```
agners, line 57:
```suggestion
    linux (${boothd},gpt4)/Image root=PARTUUID=a3ec664e-32ce-4665-95ea-7ae90ce9aa20 $default_cmdline $cmdline rauc.slot=B
```

buildroot-external/board/generic-aarch64/kernel.config
```diff
@@ -0,0 +1,147 @@
+CONFIG_EFI_STUB=y
+
```
agners, line 2: This seems to come from generic-x86-64. Lots in here is quite Intel specific and probably will never be seen on a aarch64 platform.

I'd suggest to go with something like this:

```
CONFIG_EFI_STUB=y

CONFIG_VIRTIO=y
CONFIG_VIRTIO_PCI=y
CONFIG_VIRTIO_NET=y
CONFIG_VIRTIO_BALLOON=m
CONFIG_VIRTIO_INPUT=m
CONFIG_VIRTIO_BLK=y
CONFIG_VIRTIO_BLK_SCSI=y
CONFIG_VIRTIO_CONSOLE=m
CONFIG_VIRTIO_VSOCKETS=m
CONFIG_VIRTIO_MMIO=y
CONFIG_SCSI_VIRTIO=y
CONFIG_HW_RANDOM_VIRTIO=y

CONFIG_DRM_VIRTIO_GPU=m
```

agners (changes requested):

Doridian:
@agners Your assumption about set being optional is correct. Implemented your suggestions, still boots in QEMU

agners:
Moved the new board to `buildroot-external/board/arm-uefi/` (just in case we introduce a dedicated machine for virtualization at one point).

Also added qcow2 and vdi disk images which are useful especially since the machine is mainly used for virtualization right now.

agners (approved):

PR #1761, opened 2022-02-23
Bump U-Boot to 2022.01

agners:


acostach:
Hi @agners , does the current u-boot patches included for nvme work for you? Is the nvme listed in u-boot?
I've tried them on top of this tag https://source.denx.de/u-boot/u-boot/-/commits/v2022.01 at d637294e and the mass storage controller is not detected and, of course, the nvme drive isn't either:

```
U-Boot 2022.01 (Jan 10 2022 - 18:46:34 +0000)
...
U-Boot> pci
Scanning PCI devices on bus 0
BusDevFun  VendorId   DeviceId   Device Class       Sub-Class
_____________________________________________________________
00.00.00   0x14e4     0x2711     Bridge device           0x04

U-Boot> nvme scan
U-Boot> nvme dev

no nvme devices available
```
Bumping to the latest revision in master branch of u-boot + your current u-boot nvme patches here make the controller to get  detected but not the nvme drive. However, your v2021.10 patches make nvme work on top of that revision, so I was wondering if you used a specific SHA of 2022.01 which makes nvme work for you.

Thank you

PR #1762, opened 2022-02-23
Use GRUB bootloader for all UEFI platforms

agners:
Replace Barebox with GRUB for x86-64 systems.

GRUB has a more friendly menu by default. The boot menu seems to also show up properly more often than the Barebox menu did. Furthermore GRUB supports aarch64, which allows us to use the same boot loader for the new generic-aarch64 board support.


agners:
This builds on the work of @Doridian in #1757.

For those platforms we already have `cmdline.txt`, and our rauc hook actually migrates the `cmdline.txt` (in case there are user modification.

To make `cmdline.txt` work with GRUB, we either need to prepend `cmdline=` to that file, or make GRUB really support a "vanilla" `cmdline.txt` file. I'd prefer the latter, but I understand that this seems to require a special GRUB module. I wonder how hard it is to make that happen? I will have a look at this.

Doridian:
@agners It does not require a special "mode". What would be needed is someone to write a GRUB module (and contribute it upstream or just put it into these builds here in HAOS)

Every command in GRUB is a module, pretty much (some modules have multiple commands, like loadenv).

The problem comes in that the only way for GRUB to read a file is raw is `cat`, but `cat` will just print it, no way to put that into variable (as GRUB does not have output redirection like a shell would).
The other commands GRUB has for interacting with files is `load_env` which requires a special env file and `source` which is like `#include` or a bash `source` or `.` operator.

So someone would either need to enhance `cat` to add a `--set=VARIABLE`, or make a whole new module to load a file.

buildroot-external/board/pc/generic-x86-64/cmdline.txt
```diff
@@ -1 +1 @@
-console=tty1
+cmdline="console=tty1"
```
Doridian, line 1: If this is done in x86 with the console arg in cmdline.txt, maybe do the same for ARM64?

agners:
> So someone would either need to enhance cat to add a --set=VARIABLE parameter, or make a whole new module to load a file.

Probably extending an existing one would be quicker. The `load_env` seems more appropriate to me, an additional argument/paramter with the variable, then the hole content would get loaded as value to that variable.

buildroot-external/board/pc/generic-x86-64/cmdline.txt
agners: Not sure if its really required, but we can add it if it helps :man_shrugging: 

Afaik, the ARM64 variant used the serial console. I'd like to keep that if possible. Does ARM64 has a graphical console? Enabling one `-device virtio-gpu-pci` did not give me a graphical console when I tried.

Doridian: I mean, the QEMU ARM64 model does not seem to have graphics at all from what I could find yeah.

agners: Now with `CONFIG_DRM_VIRTIO_GPU=m` it does seem to support graphics.

agners: Turns out, to get boot up console we actually have to specify `console=tty1`. I'll adjust that for `generic-aarch64`.

PR #1767, opened 2022-02-25
Fix migration from Barebox GRUB

agners:
Create GRUB env which defaults to the boot slot we are updating to. This
makes sure that the newly installed OS version will be booted on next
reboot even if installed on boot slot B.

### DIFF
--- a/buildroot-external/board/pc/generic-x86-64/hassos-hook.sh
+++ b/buildroot-external/board/pc/generic-x86-64/hassos-hook.sh
@@ -9,7 +9,14 @@ function hassos_pre_image() {
 
     cp "${BOARD_DIR}/../grub.cfg" "${EFIPART_DATA}/EFI/BOOT/grub.cfg"
     cp "${BOARD_DIR}/cmdline.txt" "${EFIPART_DATA}/cmdline.txt"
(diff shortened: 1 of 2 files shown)


agners:
@Doridian the migration is not as trivial as  I hoped it would be :see_no_evil: . Currently, when updating from OS 7.4 -> 8.0 dev, it works if updated to boot slot A. However, when updating to boot slot B, it doesn't work since OS 7.4 is still assuming barebox and the GRUB boot script defaults to boot slot A. On reboot, the new boot loader is used, but it boots into the old OS.

I came up with this solution which creates a default environment depending on boot slot the update is installed to. Maybe you have a better idea how to handle this?

The problem also exists in the other direction: Downgrading to 7.x won't update the Barebox boot state properly. But that direction I am not particularly interested in supporting.

Doridian:
@agners The only cleaner thing I could think of right now, and I don't know how that would work if at all, would pretty much add a way to RAUC to see if the bootloader changed, and if yes, make sure to apply the changes for the NEW bootloader, not the old one. But since the RAUC.conf is baked into the image, we'd have to somehow read the rauc.conf from the to-be-flashed image as opposed to the current running image.

PR #1794, opened 2022-03-16
Implement A/B switch when moving back to Barebox

agners:
When downgrading to OS 7 or older from a OS 8 installation, the bootloader
is changed back to Barebox. However, in this case Barebox does not know
which boot slot is correct (and rauc won't update Barebox' state as it
is configured to work with GRUB in OS 8+).

This implements a poor mans version to switch the boot partition. It
clears Barebox state and deploys one of the two pre-configured default
states, booting either boot slot A or B.


pvizeli (approved):
sure

PR #1804, opened 2022-03-20
Add RTL_81XX firmware to raspberry pi devices

cr0wbar:
### What?
Home assistant's operating system's kernel has support for rtl 8188eu Wifi USB dongles modules. Here is an output of `lsmod` on  `Home Assistant OS 7.5`

```
r8188eu               421888  0
lib80211               16384  1 r8188eu
```
Unfortunately there is no firmware shipped along with the kernel so the module fails to work. Here is the output from `dmesg`

```
[  143.167261] r8188eu 1-1.5:1.0: Direct firmware load for rtlwifi/rtl8188eufw.bin failed with error -2
[  143.167308] r8188eu 1-1.5:1.0: Firmware rtlwifi/rtl8188eufw.bin not available
```

The changes in this PR cause `buildroot` to package the required firmware files so that the module loads correctly.

### Please note
I am trying to build an image for rpi2 (my machine) but I keep encountering errors. I can see the firmware files being present in the `build` directory so I am fairly confident it will work. Happy to have a chat about it so that I can test it properly!

PR #1811, opened 2022-03-23
Enable wext backend for wpa_supplicant to support r8188eu

cr0wbar:
## What
Some wireless drivers still rely on the old wireless extension kernel driver, not having that built in `wpa_supplicant` means that they cannot be used on Rpi devices to connect to a wifi network. Here we also enable the NL80211 driver explicitly, which is more modern and seems to be enabled by default, to make sure it's not turned off once we add the wext setting and break retrocompatibility.

## Test
Made on a RPI2 device with the 7.5 tag and matching commit for `buildroot-external`, works without any issue on a WPA2 network.


jens-maus:
@cr0wbar 

> Some wireless drivers...

I am just curious, which "wireless drivers" are you talking about that still rely on wext?

cr0wbar:
> @cr0wbar 
> 
> 
> 
> > Some wireless drivers...
> 
> 
> 
> I am just curious, which "wireless drivers" are you talking about that still rely on wext?
The one I have tried to make work is the r8188 packaged with the kernel. I made a commit a couple of days ago to include the firmware for that driver, after including it wpa_supplicant was complaining in the logs that wext was not supported and nl80211 did not support authentication and connection. Does that make sense? I am not really an expert on the matter and open to suggestions of course.

cr0wbar:
@jens-maus 

> @cr0wbar 
> 
> 
> 
> > Some wireless drivers...
> 
> 
> 
> I am just curious, which "wireless drivers" are you talking about that still rely on wext?

Apologies, I guess kernel modules for wireless devices is probably more appropriate.

The one I have tried to make work is the r8188 module packaged with the kernel. I made a commit a couple of days ago to include the firmware for that driver, after including it wpa_supplicant was complaining in the logs that wext was not supported and nl80211 did not support authentication and connection. Does that make sense? I am not really an expert on the matter and open to suggestions of course.

agners:
Really don't like to add this legacy options :disappointed: 

Do we know this has no side effect for other drivers etc? Are these options which are enabled in modern distros (e.g. Debian bullseye)?

cr0wbar:
> Really don't like to add this legacy options disappointed
> 
> Do we know this has no side effect for other drivers etc? Are these options which are enabled in modern distros (e.g. Debian bullseye)?

It definitely appears in the debian [ManPages](https://manpages.debian.org/bullseye/wpasupplicant/wpa_supplicant.8.en.html)

After downloading `wpasupplicant_2.9.0-21_amd64.deb` and unpacking it on my machine I ran 
`./sbin/wpa_supplicant ` to see what drivers where supported, I got
```
wpa_supplicant v2.9
Copyright (c) 2003-2019, Jouni Malinen <j@w1.fi> and contributors

This software may be distributed under the terms of the BSD license.
See README for more details.

This product includes software developed by the OpenSSL Project
for use in the OpenSSL Toolkit (http://www.openssl.org/)

usage:
  wpa_supplicant [-BddhKLqqstuvW] [-P<pid file>] [-g<global ctrl>] \
        [-G<group>] \
        -i<ifname> -c<config file> [-C<ctrl>] [-D<driver>] [-p<driver_param>] \
        [-b<br_ifname>] [-e<entropy file>] [-f<debug file>] \
        [-o<override driver>] [-O<override ctrl>] \
        [-N -i<ifname> -c<conf> [-C<ctrl>] [-D<driver>] \
        [-m<P2P Device config file>] \
        [-p<driver_param>] [-b<br_ifname>] [-I<config file>] ...]

drivers:
  nl80211 = Linux nl80211/cfg80211
  wext = Linux wireless extensions (generic)
  wired = Wired Ethernet driver
  macsec_linux = MACsec Ethernet driver for Linux
  none = no driver (RADIUS server/WPS ER)
options:
  -b = optional bridge interface name
  -B = run daemon in the background
  -c = Configuration file
  -C = ctrl_interface parameter (only used if -c is not)
  -d = increase debugging verbosity (-dd even more)
  -D = driver name (can be multiple drivers: nl80211,wext)
  -e = entropy file
  -f = log output to debug file instead of stdout
  -g = global ctrl_interface
  -G = global ctrl_interface group
  -h = show this help text
  -i = interface name
  -I = additional configuration file
  -K = include keys (passwords, etc.) in debug output
  -L = show license (BSD)
  -m = Configuration file for the P2P Device interface
  -N = start describing new interface
  -o = override driver parameter for new interfaces
  -O = override ctrl_interface parameter for new interfaces
  -p = driver parameters
  -P = PID file
  -q = decrease debugging verbosity (-qq even less)
  -s = log output to syslog instead of stdout
  -t = include timestamp in debug messages
  -T = record to Linux tracing in addition to logging
       (records all messages regardless of debug verbosity)
  -u = enable DBus control interface
  -v = show version
  -W = wait for a control interface monitor before starting
example:
  wpa_supplicant -Dnl80211 -iwlan0 -c/etc/wpa_supplicant.conf
```

While this is the output on the rebuilt hassos with this PR

```# wpa_supplicant
Successfully initialized wpa_supplicant
wpa_supplicant v2.9
Copyright (c) 2003-2019, Jouni Malinen <j@w1.fi> and contributors

This software may be distributed under the terms of the BSD license.
See README for more details.

usage:
  wpa_supplicant [-BddhKLqqtuvW] [-P<pid file>] [-g<global ctrl>] \
        [-G<group>] \
        -i<ifname> -c<config file> [-C<ctrl>] [-D<driver>] [-p<driver_param>] \
        [-b<br_ifname>] [-e<entropy file>] [-f<debug file>] \
        [-o<override driver>] [-O<override ctrl>] \
        [-N -i<ifname> -c<conf> [-C<ctrl>] [-D<driver>] \
        [-p<driver_param>] [-b<br_ifname>] [-I<config file>] ...]

drivers:
  nl80211 = Linux nl80211/cfg80211
  wext = Linux wireless extensions (generic)
options:
  -b = optional bridge interface name
  -B = run daemon in the background
  -c = Configuration file
  -C = ctrl_interface parameter (only used if -c is not)
  -d = increase debugging verbosity (-dd even more)
  -D = driver name (can be multiple drivers: nl80211,wext)
  -e = entropy file
  -f = log output to debug file instead of stdout
  -g = global ctrl_interface
  -G = global ctrl_interface group
  -h = show this help text
  -i = interface name
  -I = additional configuration file
  -K = include keys (passwords, etc.) in debug output
  -L = show license (BSD)
  -M = start describing new matching interface
  -N = start describing new interface
  -o = override driver parameter for new interfaces
  -O = override ctrl_interface parameter for new interfaces
  -p = driver parameters
  -P = PID file
  -q = decrease debugging verbosity (-qq even less)
  -t = include timestamp in debug messages
  -u = enable DBus control interface
  -v = show version
  -W = wait for a control interface monitor before starting
example:
  wpa_supplicant -Dnl80211 -iwlan0 -c/etc/wpa_supplicant.conf
```
Without this patch the list of drivrs only contained
```
drivers:
  nl80211 = Linux nl80211/cfg80211
```
I understand your concern for this legacy drivers, however, here I am not really enabling the kernel layer, but rather its support within `wpa_supplicant`, the kernel  is already built-in with that driver. Does that help?
In other words here is what I think it'd be reasonable to do.
* Either support for `wext` is removed entirely and not supported forward (perfectly viable option, some old parts now present in hassos can be trimmed off)
* We can support `wext` in `wpa_supplicant` so that is not lingering in limbo

cr0wbar:
> Really don't like to add this legacy options :disappointed: 
> 
> 
> 
> Do we know this has no side effect for other drivers etc? Are these options which are enabled in modern distros (e.g. Debian bullseye)?

Upon further research, I actually found that _wext_ is explictly activated in the hassos config https://github.com/home-assistant/operating-system/blob/dev/buildroot-external/kernel/hassos.config#L61-L63, so this PR should just add support to drivers that use that in the wpa_supplicant package. Should i provide more info @agners ? I am happy to help!

agners:
Hm, I see, yeah I was remembering something about WEXT and was going to do some research.

So turns out we enabled it globally since folks were missing it and it was already enabled for RPis, see #1389.

That being said, this was mostly about enabling proc to get information from the wireless stack. Enabling wext so we can use it for drivers is a bit more involved.

Since `r8188eu`/`CONFIG_R8188EU` is actually upstream (it is in staging), and various other distributions also still enable WEXT (and that driver), it is ok from my side to enable wext.

Note that `BR2_PACKAGE_WPA_SUPPLICANT_NL80211` is set to `default y`, and hence was already enabled. Adding it explicitly certainly doesn't hurt.

agners:
It seems that the necessary firmware is part of `BR2_PACKAGE_LINUX_FIRMWARE_RTL_81XX` (enabled in #1098).

Let's add wext support for all machines and enable the WiFi driver globally (in `buildroot-external/kernel/device-support.config`).

cr0wbar:
> It seems that the necessary firmware is part of `BR2_PACKAGE_LINUX_FIRMWARE_RTL_81XX` (enabled in #1098).
> 
> 
> 
> Let's add wext support for all machines and enable the WiFi driver globally (in `buildroot-external/kernel/device-support.config`).

I can try and do it on my own, shall I create a new PR for that?

cr0wbar:
> > I can try and do it on my own, shall I create a new PR for that?
> 
> You can make the changes part of this PR.

Looks like it's already part of the global config from #1389 , but in hassos.config. Shall I delete  [this](https://github.com/home-assistant/operating-system/blame/31b9c46a790434147f135c8c7c4934cc1a5b0807/buildroot-external/kernel/hassos.config#L60-L65) block and add it to `device-support.config` as it would be better suited to be in that file or some other reason? Correct me if I am wrong but my understanding is that both files would contribute to the configuration of all devices, but `device-support` was to enable device modules.

Also I believe #1098 is for odroid boards only. I have added support for raspberry pi boards in #1804 , the same devices this PR covers. Should I enable the firmware files globally as well or leave them for specific device configuration?

agners:
> Shall I delete [this](https://github.com/home-assistant/operating-system/blame/31b9c46a790434147f135c8c7c4934cc1a5b0807/buildroot-external/kernel/hassos.config#L60-L65) block and add it to device-support.config as it would be better suited to be in that file or some other reason?

It is a device independent setting, so I think its fine where it is today.

Sorry, what I mean was to enable `CONFIG_R8188EU` in `device-support.config`.

> Correct me if I am wrong but my understanding is that both files would contribute to the configuration of all devices, but device-support was to enable device modules.

That is correct.

> Also I believe https://github.com/home-assistant/operating-system/pull/1098 is for odroid boards only. I have added support for raspberry pi boards in https://github.com/home-assistant/operating-system/pull/1804 , the same devices this PR covers. Should I enable the firmware files globally as well or leave them for specific device configuration?

Those are USB based right? I think it make sense to support them on all platforms. So yes, please enable them globally.

agners (approved):

PR #1829, opened 2022-04-02
Add bootcode.bin to Raspberry pi boot folder

cr0wbar:
## What 

`bootcode.bin` is missing from the `boot` folder of raspberry pi devices (2/3) this makes them unbootable (only checked for rpi2). This line includes `bootcode.bin` from the raspberry pi firmware package so that it can boot.

## Testing
create an image with the latest commit from the dev branch and it booted.


agners:
That sounds like a rather big issue :monocle_face: 

It seems `BR2_PACKAGE_RPI_FIRMWARE_VARIANT_PI` has been enabled in OS 7 by default. In OS 8, due to the upgrade to Buildroot 2022.02, the package name is `BR2_PACKAGE_RPI_FIRMWARE_BOOTCODE_BIN`, which is not enabled for RPi2/3 currently :sweat_smile: . This should be straight forward to fix.

Btw, I think `buildroot-external/package/rpi-firmware/rpi-firmware.mk` is not necessary anymore since Buildroot 2022.02 has the new `BR2_PACKAGE_RPI_FIRMWARE_VARIANT_PI4_X` and `BR2_PACKAGE_RPI_FIRMWARE_VARIANT_PI4_CD` options (which I already enabled). I'll create a new PR! Thanks for bringing this up!

cr0wbar:
Hey, no problem, thank you looking into this. i was just confused on why the images were not booting. Looking forward to that PR.

Cheers!

- Guglielmo

agners:
Alternative fix in PR #1831.

agners:
Btw, it seems that this does not break booting when _upgrading_, the reason is that we don't delete additional files in the boot directory... So `bootcode.bin` was still present from the initial installation. Probably also why it did not come up so far :sweat_smile:

PR #1876, opened 2022-04-27
Recreate defconfigs using savedefconfig target

agners:
Buildroot allows to generate minimal defconfigs using the savedefconfig
target. Regenerate all our configurations so they all look alive and are
minimalistc.


agners:
Verified that this doesn't change the generated full configuration of Buildroot. So this has no affect to the built output.

PR #1907, opened 2022-05-16
Linux: Update kernel 5.15.40

gregkh:
Signed-off-by: Greg Kroah-Hartman <gregkh@linuxfoundation.org>


gregkh:
No reason at all, I just figured you all wanted to do this in stages, one release at a time :)

I'll bump this to .40 if you think that's better.

agners:
I usually bump the stable kernel before I do the next OS release, often build a dev release and do some real world testing before building the actual release.

> I'll bump this to .40 if you think that's better.

Yes please. OS 8.1 will likely be released soon.

gregkh:
Now changed to .40

Want me to do this every week when I release these kernels to keep these devices up to date and secure?

jens-maus:
> Want me to do this every week when I release these kernels to keep these devices up to date and secure?

You really have the time to do that and bump the kernel version of HAos here together with your linux core development responsibilities at https://github.com/torvalds/linux? ;-)

I am impressed and would definitly (+1) vote for it if you would regularly contribute to this project.

gregkh:
I think you mean to point to https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/  kernel development doesn't happen on github :)

And yes, bumping the kernel version here wouldn't be that difficult, I'll try to remember now that I rely more on HA and want to see it succeed.

jens-maus:
> And yes, bumping the kernel version here wouldn't be that difficult, I'll try to remember now that I rely more on HA and want to see it succeed.

Perfect. But we could also in theory think about adding a github workflow/dependency check that automatically generates kernel update PRs as soon as you have released a new kernel ;)

agners:
Actually, there is a rudimentary script committed at `scripts/update-kernel-upstream.sh`. I guess wrapping that into a GH Action shouldn't be a big deal. What I am a bit worried is to have countless of "unnecessary"  bumps in the git history/change log (which don't get released).

Ideally it should be like Dependabot, where you can just leave the PR sitting there and you get it updated to the latest automatically. Then we can merge whenever we actually plan to release.

gregkh:
"Bumps" are good, as they provide a way to bisect back in case of regressions.  Jumping large version numbers is always harder in the end if there is a problem.

Anyway, it's your call as to if you want an "update for every stable release" or "one update a week", or "one update a month" or something else. I just want to make sure my HA device is running the latest and most stable/secure kernel release there is at the moment.  Thanks!

agners:
> "Bumps" are good, as they provide a way to bisect back in case of regressions. Jumping large version numbers is always harder in the end if there is a problem.

Yeah true, but then I sometimes trigger builds quite a bit apart, so there won't be any binaries. But then again, having the commits already in source can be helpful still. I guess a "clean change log" is kinda lame reason to not do the right thing: Make sure dev is pointing to the latest stable. :sweat_smile: 

Out of curiosity, what hardware are you running HAOS on?

Issue #1050, opened 2020-12-02
Hassos ova for arm cpu ESXi Arm Edition?


shirou93:
Why not release hassos ova for arm? VMware released ESXi Arm Edition and RPI4 8gb make sense.

Is it possible convert rpi.img to working vmdk?

agners:
@shirou93 patch welcome :-D Yeah you can convert the disk image to a vmdk and try create a machine which boots. The utility `qemu-img` can do that.

ipha:
I tried building a generic arm vm image, but I couldn't get a bootloader working. Best I can tell barebox doesn't support arm EFI.

GRUB could be an option, but I don't know if it supports the A/B partitions.

shirou93:
@ipha @agners  i try but not booting. Maybe bootloader was broken when convert img to vmdk.

agners:
Hm, yeah bootloader is a problem: We use U-Boot in all Arm builds. It seems that ESXi Arm is expecting a UEFI bootloader such as Grub2. Gleening at this blog post it seems that booting something else is possible in theory, but it seems that requires some hacking: https://blogs.vmware.com/arm/2020/10/23/uefi-less-vms-with-esxi-arm-fling-1-1/

myericbaba:
anyone try to follow this guideline for HA?
https://www.virtuallyghetto.com/2020/10/how-to-run-raspberry-pi-os-as-a-vm-on-esxi-arm.html

NeverEatYellowSwissSnow:
I am interested in this too. Would be so nice to run HA OS on ESXi ARM!

agners:
As noted in #1549, there are now images named `generic-aarch64` which support UEFI boot on arm64. I haven't tested them on ESXi for ARM, I am keen to here if the image also works on ESXi.

There are development builds available at https://os-builds.home-assistant.io/8.0.dev20220223/.

Grandma-Betty:
> As noted in #1549, there are now images named `generic-aarch64` which support UEFI boot on arm64. I haven't tested them on ESXi for ARM, I am keen to here if the image also works on ESXi.
> 
> There are development builds available at https://os-builds.home-assistant.io/8.0.dev20220223/.

Just tried it with the same settings as HAOS runs on ESXi-ARM, without any luck.
I have used the **.vmdk** disk image as well as the **.img** file from source.
I also tried to play around with hard disk controller settings too, no luck.
It's not booting.

agners:
[...]

PR #1915, opened 2022-05-17
Add open-vm-tools to AArch64 for better VMware support (#1050)

agners:


agners:
This does not seem to compile:
```
checking build system type... x86_64-pc-linux-gnu
checking host system type... aarch64-buildroot-linux-gnu
configure: error: Unknown architecture.
```

It seems that VMware tools are rather old (10.3.5) and don't support AArch64.

claplace:
stable 11.3.x and 12.0.x have arm64 support.

Issue #1911, opened 2022-05-16
HAOS 8.0 fails to boot after update on CM4 with NVMe SSD


Doridian:
### Describe the issue you are experiencing

I am running HAOS on a CM4 with an NVMe SSD.
After updating from HAOS 7.6 to 8.0 in the GUI, HomeAssistant fails to come up.
It does request an IP via DHCP on boot and will respond to pings on it, however, according to my DHCP server logs, those requests only ever come from `U-Boot.armv8`, not the actual OS.
Plugging in a screen, I see absolutely nothing.

Even repeated reboot attempts get me absolutely nowhere, not back to 7.6, no logs, nothing.

### What operating system image do you use?

rpi4-64  (Raspberry Pi 4/400 64-bit OS)

### What version of Home Assistant Operating System is installed?

8.0

### Did you upgrade the Operating System.

Yes

### Steps to reproduce the issue

1. Use NVMe SSD with HAOS 7.6
2. Update to 8.0 in the GUI
3. Stopped booting

### Anything in the Supervisor logs that might be useful for us?

```txt
Nothing
```

### Anything in the Host logs that might be useful for us?

```txt
Nothing
```

Doridian:
Adding to this:
- Fresh install of HAOS 8.0 to NVMe SSD -> Same symptoms, complete boot fail
- Fresh install of HAOS 7.6 to the same NVMe SSD -> Immediate boot success

Doridian:
> Any idea how we can get it to run?
> 
> Its complete dead, no SSH nothing

Only way I could recover mine is pull SSD, reflash it with 7.6, restore backup.

KartoffelToby:
@Doridian did you know were the backups are stored on the nvme? :D

Doridian:
> @Doridian did you know were the backups are stored on the nvme? :D

I always download them to my NAS (so I have them ready in case of drive failure).
But I'd go digging around on the last partition of the NVMe, that seems to contain everything.
If you can't find anything there, maybe just image the whole partition and restore it, no idea how good that is as a solution, however.

agners:
@KartoffelToby You can replace the `u-boot.bin` file on the first partition with the version from OS 7.6. For your continence, I extracted it from HAOS 7.6: [rpi4-64 u-boot.img.tar.gz from HAOS 7.6](https://github.com/home-assistant/operating-system/files/8702067/u-boot.img.tar.gz)

pickonedev:
[...]

PR #1957, opened 2022-06-01
Apply NVMe fixes to all Raspberry Pi boards (#1911)

agners:


rolf-tx:
Is there any build that have the patches applied. I just tried 8.1 and 8.2 beta and both fail to boot from SDcard if NVME drive is installed (I don't want to wipe out the one working 8.1 32bit I have om the nvme)

agners:
> Is there any build that have the patches applied. I just tried 8.1 and 8.2 beta and both fail to boot from SDcard if NVME drive is installed (I don't want to wipe out the one working 8.1 32bit I have on the nvme)

The patches are part of 8.2.

The not-booting when NVMe is plugged-in is a different problem tracked in #1887.

PR #1976, opened 2022-06-15
Fix network.md ha cli network wifi example

nel:
Removing `--enabled` option from ha cli network update example as it is not supported


homeassistant:
Hi @nel, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;operating-system#1976).

Thanks!

Issue #1706, opened 2022-01-06
Docker key.json has invalid contents and the system refuses to boot


mdcollins05:
### Describe the issue you are experiencing

I'm 2 for 2 now on this. Each reboot after an upgrade to the OS somehow leads to a corrupt key.json file on my Home Assistant OS install. Each time I have to boot via a recovery USB drive, mount the right partition (in this case `/dev/sda7`) and delete the `/etc/docker/key.json` file to fix the system. I don't intentionally manage this file in any way, nor do I care about the contents, so the Docker generated version is just fine with me.

As a test, I fixed the file, booted into Home Assistant and then restarted back into my recovery USB to inspect the contents. They were valid JSON instead of the invalid content.

### What operating system image do you use?

generic-x86-64 (Generic UEFI capable x86-64 systems)

### What version of Home Assistant Operating System is installed?

6.6

### Did you upgrade the Operating System.

Yes

### Steps to reproduce the issue

1. Update Home Assistant OS via supervisor in the web interface
2. Home Assistant then fails to reboot into a valid system

### Anything in the Supervisor logs that might be useful for us?

```txt
Nothing in there from the update or prior to the successful boot of my system.
```

### Anything in the Host logs that might be useful for us?

```txt
Nothing
```

## System Health

version | core-2021.11.1
-- | --
installation_type | Home Assistant OS
dev | false
hassio | true
docker | true
user | root
virtualenv | false
python_version | 3.9.7
os_name | Linux
os_version | 5.10.88
arch | x86_64
timezone | America/Los_Angeles

<details><summary>Home Assistant Community Store</summary>

GitHub API | ok
-- | --
Github API Calls Remaining | 4890
Installed Version | 1.16.0
Stage | running
Available Repositories | 932
Installed Repositories | 9

</details>

<details><summary>Home Assistant Cloud</summary>

logged_in | true
-- | --
subscription_expiration | January 13, 2022, 4:00 PM
relayer_connected | true
remote_enabled | false
remote_connected | false
alexa_enabled | true
google_enabled | false
remote_server | us-west-2-1.ui.nabu.casa
can_reach_cert_server | ok
can_reach_cloud_auth | ok
can_reach_cloud | ok

</details>

<details><summary>Home Assistant Supervisor</summary>

host_os | Home Assistant OS 7.1
-- | --
update_channel | stable
supervisor_version | supervisor-2021.12.2
docker_version | 20.10.9
disk_total | 219.4 GB
[...]

PR #1988, opened 2022-06-24
Remove key.json file if it appears to be corrupted (#1706)

agners:


buildroot-external/rootfs-overlay/usr/libexec/docker-failure
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+if [ "$SERVICE_RESULT" = "exit-code" ] && [ "$EXIT_STATUS" = "1" ]; then
+    echo "Docker exited with exit status 1, this might be caused by corrupted key.json."
+    size=$(stat -c %s "/etc/docker/key.json")
+    echo "key.json: ${size} bytes"
```
frenck, line 6: Wouldn't it be better to check the contents / check if the json can be parsed?

agners: Yeah, I was actually thinking the same. It will still be a heuristic in the end, but I guess there is a quite good chance that there are other errors which lead to the same exit code, where we don't want to remove the `key.json`.

frenck (approved):
LGTM

PR #2025, opened 2022-07-20
Linux: Update kernel 5.15.55

agners:


FrankJaco:
When attempting to update from 8.2 to 8.4 on generic x86 hardware it seems to go thru the motion, but not update.  It works on it for quite awhile then the host reboots, but after it comes up, the upgrade is still advertised and looking at the terminal I am still on 8.2.   I have tried this 3 times with the same result. Any ideas?  What can I look at to troubleshoot the issue?

fj

bfellner:
Mine didn't work at all 8.2 to 8.4.  I did the upgrade and then I could no longer access the machine.  It wasn't even pulling the static DHCP IP address (not responding to ping).  I also had nothing when I plugged in a monitor.  I took the drive out and proceeded to reinstall clean with 8.4.  Same thing.  I'm going to revert back to 8.2.

agners:
> When attempting to update from 8.2 to 8.4 on generic x86 hardware it seems to go thru the motion, but not update. It works on it for quite awhile then the host reboots, but after it comes up, the upgrade is still advertised and looking at the terminal I am still on 8.2. I have tried this 3 times with the same result. Any ideas? What can I look at to troubleshoot the issue?

Check the Supervisor logs, it may contain some hints. It sounds similar to #1978, if you can try what is mentioned in https://github.com/home-assistant/operating-system/issues/1978#issuecomment-1189158058 and post the result in that issue it might be helpful.

PR #2034, opened 2022-07-27
Allow to Wipe essential boot files to trigger boot from USB host

agners:
This makes the Red+Blue Button cause the boot loader to wipe start4.elf,
which is essential for the boot loader to boot from eMMC. With the file
missing, the Raspberry Pi firmware will continue its boot flow and boot
from USB host next. This allows to run the Home Assistant OS Installer
from a USB flash drive again.


frenck (approved):
Nice!

PR #2038, opened 2022-07-30
Add LED control through device tree overrides

agners:


scxs535:
I use follow code to disable the PWR LED of Raspberry Pi4b 8Gb with HA OS ( core 9.4) in boot/config.txt file
dtparam=pwr_led_trigger=none
dtparam=pwr_led_activelow=off
it works fine still with operating system to 8.5

but it doesn't work anymore after upgrade operating system to 9.0

agners:
Try removing the second line `dtparam=pwr_led_activelow=off`.

scxs535:
i just tried it, it doesn't work, the led is still on

scxs535:
thanks, 
it works now perfect with only this code in config.txt on operating system 9.0: 
dtparam=pwr_led_activelow=off

agners:
This PR is about LEDs on Yellow, this is not directly related to regular Raspberry Pis.

Issue #2064, opened 2022-08-15
Cannot connect to Raspberry PI 4 over Bluetooth


gcampax:
### Describe the issue you are experiencing

I tried to set it up a RPi4 running the most recent Home Assistant as a bluetooth speaker.
With some fiddling, I got it to pair to my phone and laptop, but sound does not work.
The logs from the audio container show:
> E: [pulseaudio] bluez5-util.c: org.bluez.Media1.RegisterEndpoint() failed: org.freedesktop.DBus.Error.UnknownMethod: Method "RegisterEndpoint" with signature "oa{sv}" on interface "org.bluez.Media1" doesn't exist

This seems identical to https://github.com/Arkq/bluez-alsa/issues/542 which suggests it is a build configuration problem: bluez (bluetoothd) needs to be built with --enable-a2dp --enable-avrcp

### What operating system image do you use?

rpi4  (Raspberry Pi 4/400 32-bit OS)

### What version of Home Assistant Operating System is installed?

8.4

### Did you upgrade the Operating System.

Yes

### Steps to reproduce the issue

1. Enable Bluetooth. This popped up in the UI and I confirmed.
2. Pair a phone or laptop with Bluetooth. I did it by hand by calling DBus methods using the web terminal (I wish there was a UI for it)
3. Connect to Home Assistant from the phone or laptop.

### Anything in the Supervisor logs that might be useful for us?

```txt
See above
```

### Anything in the Host logs that might be useful for us?

```txt
See above
```

agners:
Indeed, it seems that BlueZ is by default not compiled with this options in Buildroot. There is the Buildroot `BR2_PACKAGE_BLUEZ5_UTILS_PLUGINS_AUDIO` option, I'll enable it in OS 9.

agners:
Can you try the latest development build `9.0.dev20220816`? It should have the options configured correctly.

gcampax:
Happy to try it! How do I do that? I tried:
```
ha os update --version 9.0.dev20220816
```
and I got
> Error: Error raised from OTA Webserver: 404

gcampax:
Ok, testing this build I'm making progress, in that I'm able to connect from my phone in HSP/HFP mode, but actual audio playing fails.
I'm seeing these errors in the logs:
> E: [pulseaudio] backend-native.c: socket(SEQPACKET, SCO) Address family not supported by protocol
> E: [pulseaudio] backend-native.c: accept(): Not a socket

[...]

Issue #1746, opened 2022-02-09
bluetoothctl won't keep devices connected


shawly:
### Describe the issue you are experiencing

I can't connect any bluetooth device to HA OS through bluetoothctl. Tried a bluetooth remote camera shutter which registers as input-keyboard and a Logitech MX Master.

I can reproduce the issue on two different HA OS instances on two different hosts with two different BT devices (a BT 5.0 USB Dongle and the internal BT module from an Intel NUC).

I cannot reproduce the issue on standard Linux machines or on the host machine itself, only within HA OS this problem exists. BT and BLE does work fine though in HA.

### What operating system image do you use?

ova (for Virtual Machines)

### What version of Home Assistant Operating System is installed?

7.3

### Did you upgrade the Operating System.

No

### Steps to reproduce the issue

1. In a terminal execute `bluetoothctl`
2. Execute `scan on`
3. Copy the mac address of the device
4. Execute `pair <mac>`
5. Execute `trust <mac>`
6. Execute `connect <mac>`

Example: 

```
[bluetooth]# pair 11:22:33:FF:5F:AB
Attempting to pair with 11:22:33:FF:5F:AB
[CHG] Device 11:22:33:FF:5F:AB Connected: yes
[CHG] Device 11:22:33:FF:5F:AB Modalias: usb:v05ACp022Cd011B
[CHG] Device 11:22:33:FF:5F:AB UUIDs: 00001124-0000-1000-8000-00805f9b34fb
[CHG] Device 11:22:33:FF:5F:AB UUIDs: 00001200-0000-1000-8000-00805f9b34fb
[CHG] Device 11:22:33:FF:5F:AB ServicesResolved: yes
[CHG] Device 11:22:33:FF:5F:AB Paired: yes
Pairing successful
[CHG] Device 11:22:33:FF:5F:AB ServicesResolved: no
[CHG] Device 11:22:33:FF:5F:AB Connected: no
bluetooth]# trust 11:22:33:FF:5F:AB
[CHG] Device 11:22:33:FF:5F:AB Trusted: yes
Changing 11:22:33:FF:5F:AB trust succeeded
[bluetooth]# connect 11:22:33:FF:5F:AB
Attempting to connect to 11:22:33:FF:5F:AB
[CHG] Device 11:22:33:FF:5F:AB ServicesResolved: yes
[CHG] Device 98:D3:31:FB:6F:F1 Connected: yes
Failed to connect: org.bluez.Error.NotAvailable
[...]

PR #2065, opened 2022-08-16
Enable BlueZ Audio and HID plug-ins (#2064, #1746)

agners:


buildroot-external/configs/generic_aarch64_defconfig
```diff
@@ -71,6 +71,9 @@ BR2_PACKAGE_LIBCGROUP=y
 BR2_PACKAGE_LIBCGROUP_TOOLS=y
 BR2_PACKAGE_BLUEZ5_UTILS=y
 BR2_PACKAGE_BLUEZ5_UTILS_CLIENT=y
+BR2_PACKAGE_BLUEZ5_UTILS_PLUGINS_AUDIO=y
+BR2_PACKAGE_BLUEZ5_UTILS_PLUGINS_HID=y
+BR2_PACKAGE_BLUEZ5_UTILS_PLUGINS_HOG
```
jens-maus, line 76: To me this looks like you are missing the final `=y` at the end of the line. Isn't it?

agners: Right, thanks for spotting this!

```suggestion
BR2_PACKAGE_BLUEZ5_UTILS_PLUGINS_HOG=y
```

agners: FWIW, it seems that Kconfig/Buildroot automatically makes a `BR2_PACKAGE_BLUEZ5_UTILS_PLUGINS_HOG=y` out of it. That is why I didn't notice when building locally. Still, should be with `=y`.

PR #2073, opened 2022-08-19
rpi4: Enable arm_boost=1 to unlock 1.8Ghz CPU 

mhaas:
The official Raspberry Pi OS enables a "boosted" 1.8GHz mode since their Debian bullseye based release [source]. This commit brings this feature to HASS OS.

See the official [documentation on arm_boost], which states the following:

> This change should be safe for all such boards [...]
>  New Raspberry Pi OS images from Bullseye onwards come with this setting by default.

[source]: https://www.raspberrypi.com/news/bullseye-bonus-1-8ghz-raspberry-pi-4/
[documentation on arm_boost]: https://www.raspberrypi.com/documentation/computers/config_txt.html#arm_boost-raspberry-pi-4-only

### DIFF
--- a/buildroot-external/board/raspberrypi/config.txt
+++ b/buildroot-external/board/raspberrypi/config.txt
@@ -75,6 +75,8 @@ dtparam=audio=on
 # Enable DRM VC4 V3D driver on top of the dispmanx display stack
 dtoverlay=vc4-fkms-v3d
 max_framebuffers=2
+# Enable boost from 1.5Ghz to 1.8Ghz on compatible models
+arm_boost=1
 
 [all]
 #dtoverlay=vc4-fkms-v3d
--- a/buildroot-external/board/raspberrypi/config.txt
+++ b/buildroot-external/board/raspberrypi/config.txt
@@ -75,7 +75,7 @@ dtparam=audio=on
 # Enable DRM VC4 V3D driver on top of the dispmanx display stack
 dtoverlay=vc4-fkms-v3d
 max_framebuffers=2
-# Enable boost from 1.5Ghz to 1.8Ghz on compatible models
+# Enable boost from 1.5GHz to 1.8GHz on compatible models
 arm_boost=1
 
 [all]
--- a/buildroot-external/board/raspberrypi/config.txt
+++ b/buildroot-external/board/raspberrypi/config.txt
@@ -75,7 +75,7 @@ dtparam=audio=on
 # Enable DRM VC4 V3D driver on top of the dispmanx display stack
 dtoverlay=vc4-fkms-v3d
 max_framebuffers=2
-# Enable boost from 1.5GHz to 1.8GHz on compatible models
+# Enable boost from 1.5Ghz to 1.8Ghz on compatible models
 arm_boost=1
 
 [all]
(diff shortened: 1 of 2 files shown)


mhaas:
With this change applied to my config.txt, my core voltage goes up to 0.95V from 0.880V under load.

agners:
Hm, interesting, it seems Raspberry PI folks do not 100% trust that it won't have side effects though, since it is behind a flag and only gets deployed with new installations, at least that is how I interpret this sentence:

> New Raspberry Pi OS images from Bullseye onwards come with this setting by default. 

Now, `config.txt` is also not updated in our case, so this change would affect only new installations. On a quick glance, I lean towards merging this, but I'd like to do some more research.

agners:
It seems that the CM4 does not receive the booost. From Eben Uptons response it seems not really a technical limitation, more because of a more conservative approach (from comments in https://www.raspberrypi.com/news/bullseye-bonus-1-8ghz-raspberry-pi-4/):

> CM4 remains at 1.5GHz by default. With Compute Module we are much more focused on providing a stable, consistent, platform for our industrial customers than on pushing the performance envelope.

I any case, I am fine to enable this for Raspberry Pi 4. It could lead to some Pis to get a bit warmer than usual if they are under load, but the Pi seems to behave well when thermally limited.

incarvr6:
So how to enable on rp4? Is it yaml config or something else?

agners:
Raspberry Pi's before v1.2 only got validated for 1.5GHz. Setting `arm_freq=1800` is overclocking. Means, it *might* work, but if you are unlucky, your particular device is not capable to do it. So setting that by default in HAOS is a no-go.

The `arm_boost=1` is a new flag, which enables "smarts" in the Raspberry Pi firmware: Essentially it checks if the system is running on a newer RPi 4, and only if so, enables 1.8GHz. That is not overclocking, but just clocking at the highest validated speed...

SpicyLimes:
After shutting down my RPi 4B (running Home Assistant), I removed the SD Card and mounted all partitions on a Linux system - I was easily able to access and modify the `HASSOS-Boot` partitions' "config.txt", however the `arm_boost=1` does not work for me (checked via HA's Terminal using the command `cat /sys/devices/system/cpu/cpu0/cpufreq/cpuinfo_max_freq`).

After 3 "Reboots" (not "Restarts") of Home Assistant, I was not able achieve the 1.8GHz objective. However, after adding `arm_freq=1800` and `over_voltage=4` overlays, I was able to achieve the 1.8GHz speeds. My RPi 4B 4GB board is in-fact a v1.2 board, so I am not certain as to why I was not able to use `arm_boost=1` - maybe because I am using a Home Assistant 32bit, but normally that doesn't really matter - at least as far as I know. I am also NOT seeing any increase in temperature, even running a Genie Server and multiple other resource-intensive Add-On's. 

Oh well - objective achieved either way.

mhaas:
I also noticed that ghe CPU frequency seems to be stuck at 1.8 GHz. But this is not the case when you check the statistics of the scaling governor:


Screenshot_20220920-165506_Home Assistant

Perhaps the cpu now ramps up very quickly to 1.8GHz so the script that checks the frequency only sees 1.8GHz.

I can share my script to get the distribution tomorrow if desired.

SpicyLimes:
> I can share my script to get the distribution tomorrow if desired.

@mhaas Yes, please! That would be great.

Issue #2092, opened 2022-08-30
Boot Loop Issue


igloo15:
### Describe the issue you are experiencing

I came home from work and noticed my home assistant was not working. I checked on the machine and rebooted it. Upon rebooting the system began to boot loop. 

I hooked up a monitor to the intel nuc machine and viewed the boot logs. It seems that as the Network Manager Scripts Dispatcher starts it begins to get errors.

`Unable to read configuration index 0 descriptor/all`

Then after a few of those messages it says `Finished Network Manager Service`. Followed after that is a message saying `Network Target is Online` then `Waiting Until Kernel Time Synchronized` with a timer counting to 1 minute 30 seconds. When it hits 1 minute 30 seconds a ton of text flashes by and it reboots.

There were two slots A & B and I tried both of them but both did the same thing.  

### What operating system image do you use?

generic-x86-64 (Generic UEFI capable x86-64 systems)

### What version of Home Assistant Operating System is installed?

latest

### Did you upgrade the Operating System.

No

### Steps to reproduce the issue

1. Unsure it just randomily started rebooting
2.
3.
...

### Anything in the Supervisor logs that might be useful for us?

```txt
Can't see supervisor logs stuck at boot
```

### Anything in the Host logs that might be useful for us?

```txt
usb 2-1: unable to read configuration index 0 descriptor/all

waiting until Kernel Time Sychronization
```

### Additional information

Intel Nuc machine
Home Assistant 2022.8.6
Two Coral USB devices
One USB Zigbee Controller
One USB Zwave Controller

agners:
```
usb 2-1: unable to read configuration index 0 descriptor/all
```

Those messages are likely caused by a USB device. They should not be fatal. Can you try to disconnect the USB devices one by one to see which one is causing that?

`Waiting Until Kernel Time Synchronized` counting down to 0 means that the system isn't able to communicate with the NTP server. Are you sure your internet connectivity is fine at that point?

However, even if the NTP is not accessible, this should not lead to a system reboot afterwards.

The system reboots automatically if Docker cannot start properly. This can happen if there is data corruption on the data partition. This partition is shared between the two instances (A/B), so that would explain why both options fail.

One method to debug it from this point is by pressing e at the GRUB menu and appending the string `systemd.unit=rescue.target` to the line starting with `linux ..` and press F10 to access an emergency console.  From there you can check what Docker reported in the last boot by using `journalctl -b -1 -u docker.service`.

[...]

PR #2097, opened 2022-08-31
Check free disk space before starting Docker (#2092)

agners:
It seems that Docker can fail to start if there is no space left on the
device. Try to free up some space in that case by asking journald to
limit its size to 256MiB.

This should work for any storage larger than ~2.5GiB (as the journals
maximum size is 10% of the disk size). It still should leave enough
logs to diagnose problems if necessary.

Note: We could also limit the size of the journal in first place, but
that isn't sustainable: Once that space is used up, we run into the
same problem again.

By only asking journalctl to free up if necessary, we kinda (miss)use
the journal as way to "reserve" some space which we can free up at boot
if necessary.

### DIFF
--- a/buildroot-external/rootfs-overlay/etc/systemd/system/docker.service.d/failure.conf
+++ b/buildroot-external/rootfs-overlay/etc/systemd/system/docker.service.d/failure.conf
@@ -2,4 +2,5 @@
 FailureAction=reboot
 
 [Service]
+ExecStartPre=/usr/libexec/docker-disk-check
 ExecStopPost=/usr/libexec/docker-failure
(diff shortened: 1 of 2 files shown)


agners:
This is a bit crude, but it seems to work in my testing. The regular root reserve unfortunately doesn't work as most services run in containers as root. Better would be to use actual disk quota, and enforcement. But disk quotas are a pain to manage, and there is overhead associated to it.

An alternative to the current solution would be to store a "reserve" file, which we can delete at startup. But that is rather crude too, and leads to unnecessary writes (as the file needs to really be written out, a sparse file would not do the job here).

PR #2103, opened 2022-09-02
Bump commit interval to 30s

agners:
A higher file system commit interval can help to decrease the amount of
writes. In tests, a commit interval of higher than 30s seems not to help
much in practice. Settle with 30s for now.

Some tests about this change can be found here:
https://community.home-assistant.io/t/researching-i-o-write-improvements/456227

### DIFF
--- a/buildroot-external/rootfs-overlay/usr/lib/systemd/system/mnt-data.mount
+++ b/buildroot-external/rootfs-overlay/usr/lib/systemd/system/mnt-data.mount
@@ -12,6 +12,7 @@ Wants=systemd-fsck@dev-disk-by\x2dlabel-hassos\x2ddata.service systemd-growfs@mn
 What=/dev/disk/by-label/hassos-data
 Where=/mnt/data
 Type=ext4
+Options=commit=30
 
 [Install]
 WantedBy=local-fs.target


agners:
It seems like big distributions like Ubuntu leave it as default. But those are also geared towards desktop type systems, so might not be a good indicator. [ArchLinux has it](https://wiki.archlinux.org/title/ext4#Increasing_commit_interval) as a "performance improvement" tip set to 60s in the example. [Debian Wiki](https://wiki.debian.org/%20SSDOptimization) has a tip to set to 600s. ChromeOS synchronizes it with `vm.dirty_expire_centisecs`, and sets both to 600s (see [files/00-sysctl.conf](https://chromium.googlesource.com/chromiumos/overlays/chromiumos-overlay/+/master/chromeos-base/chromeos-base/files/00-sysctl.conf#102), there [have been discussions](https://bugs.chromium.org/p/chromium/issues/detail?id=464959) to lower it, but it seems it stayed at 600s. It seems that the main reason is not to save writes, but to make sure fewer wake-ups are necessary and therefor maximize runtime when running on battery). 

All in all, I expect this to create a bit more issues where files get corrupted (like the key.json corruption we are seeing, see https://github.com/home-assistant/operating-system/issues/1706). But in the end those are software bugs which should be fixed (either the software needs to use a save write mechanism, or handle empty files properly).

pvizeli:
Let's test the 30s. I'm more worry about docker layers

pvizeli (approved):

PR #2116, opened 2022-09-06
Move Docker configuration to daemon.json

agners:
This is more readable than passing arguments to the daemon directly. It also shortens the ExecStart command significantly, which is stored in every log entry in systemd-journald.


pvizeli:
should we cleanup the overlay on the OTA?

agners:
> should we cleanup the overlay on the OTA?

We still need the key.json, it is now just explicitly referenced (see `"deprecated-key-path"`). I was wondering if the key file is still necessary (since it is saying "deprecated"), but it seems that current Docker still requires it.

pvizeli:
Without bind mount, I think new installation or reseted overlay will not have that folder on the overlay. So maybe `/mnt/data/docker/key.json` is better?

agners:
> Without bind mount, I think new installation or reseted overlay will not have that folder on the overlay.

Right, but in tests it seems that the Docker daemon automatically creates the directory.

> So maybe `/mnt/data/docker/key.json` is better?

We would need to migrate from the overlay location. I know that its not nice that we peek into the overlay directly, but it seems the simplest solution. The key file will become unnecessary in the future:

https://github.com/moby/moby/pull/43555
https://github.com/moby/moby/pull/41295

HangrilyJon:
Are end users able to get write access or the ability to override/supersede the default settings provided in this file? A (admittedly small) users have already commented on #2135 who need this ability, mostly to set their own default base network.

PR #2145, opened 2022-09-19
Add LXD VM Guest Agent Support

### LINKED ISSUE #676: Enable 9p VIRTIO kernel module
**HassOS release with the issue:**

**Supervisor logs:**

**Journal logs:**

**Kernel logs:**

**Description of problem:**
As part of a migration from a Supervised Linux install to HASSOS on KVM I am missing the ability to share a directory with the host os.  This is possible within KVM but the OS is missing the required 9p kernel module.  This leads to the following error when trying to mount:

mount -t 9p -o trans=virtio,version=9p2000.L hostshare /tmp/host_files
mount: unknown filesystem type '9p'

I believe the two required module are 9pnet and 9pnet_virtio

bryanyork:
Fixes #676


buildroot-external/package/lxd-guest-agent/99-lxd-agent.rules
```diff
@@ -0,0 +1 @@
+ACTION=="add", SYMLINK=="virtio-ports/org.linuxcontainers.lxd", TAG+="systemd", ACTION=="add", RUN+="/bin/systemctl start lxd-agent.service"
```
pvizeli, line 1: Can we not use systemd unit VM detection like all other agent using?

bryanyork: What would you recommend? A separate check for "/dev/lxd/sock" (indicating we are in a guest LXD) that is called by systemd?

agners: No, systemd can have conditions in descriptive form. There is `ConditionVirtualization=` which can be used to detect virtualization. So we can just set that in `lxd-agent.service`, I think, but we should test it with and without LXC if things behave as expected.

https://www.freedesktop.org/software/systemd/man/systemd.unit.html#ConditionVirtualization=
https://www.freedesktop.org/software/systemd/man/systemd-detect-virt.html#

bryanyork: I understand now. Hmm, looking my test system there is not a good way to detect a LXD VM specifically, systemd-detect-virt shows just qemu. Apologies, I scraped this together from the debian package lxd-agent-loader. Will dig in to see if I can better detect this. Would it not be a good enough test to check for /dev/lxd/sock in this case if the existing systemd detection is not good enough?

bryanyork: Also, we may consider updating the virtualization checks for the qemu agent, as that tries and fails to load in this setup.

agners: > systemd-detect-virt shows just qemu

Hm, I see there is no `lxd` :cry: 

Ideally we'd add lxd to systemd-detect-virt, then the Qemu problem would be solved as well :smile:

bryanyork: I just looked at the systemd source it looks like they are detecting /sys/class/dmi/id/sys_vendor which shows as. I just looked and LXD shows up properly in /sys/class/dmi/id/board_name. 

I can try and create a PR for systemd, but that is a tall order. Any chance we can implement a hack and I can follow up later with the correct systemd PR?

buildroot-external/board/pc/ova/kernel.config
```diff
@@ -12,6 +12,16 @@ CONFIG_VMWARE_VMCI=y
 CONFIG_VMWARE_BALLOON=y
 CONFIG_VMWARE_PVSCSI=y
 
+# These options are for LXD Guest Support
+CONFIG_NET_9P=y
+CONFIG_NET_9P_VIRTIO=y
+CONFIG_9P_FS=y
+CONFIG_9P_FS_POSIX_ACL=y
```
jens-maus, line 19: Only comment from my side: It is really necessary to integrate all these 9P config option directly into a kernel or isn't it better to compile them as kernel modules instead? IMHO, compiling them as loadable kernel modules should be a better approach.

buildroot-external/package/lxd-guest-agent/99-lxd-agent.rules
agners: I'd prefer if we have a PR with systemd already open before patching it on our side, so the process is at least kicked off :)

Seems that worth a try submitting a PR which just reads `/sys/class/dmi/id/board_name`?

pvizeli: We can apply the PR also into our build system, so we not have to wait to get approved by systemd @bryanyork

buildroot-external/package/lxd-guest-agent/lxd-guest-agent.mk
```diff
@@ -0,0 +1,21 @@
...
+define LXD_GUEST_AGENT_INSTALL_INIT_SYSTEMD
+	$(INSTALL) -D -m 644 $(LXD_GUEST_AGENT_PKGDIR)/lxd-agent.service \
+		$(TARGET_DIR)/usr/lib/systemd/system/lxd-agent.service
+	$(INSTALL) -D -m 755 $(LXD_GUEST_AGENT_PKGDIR)/lxd-agent-setup \
+		$(TARGET_DIR)/usr/lib/systemd/lxd-agent-setup
+	$(INSTALL) -d $(TARGET_DIR)/etc/systemd/system/multi-user.target.wants
+	ln -fs /usr/lib/systemd/system/lxd-agent.service \
+		$(TARGET_DIR)/etc/systemd/system/multi-user.target.wants/lxd-agent.service
+        $(INSTALL) -D -m 644 $(LXD_GUEST_AGENT_PKGDIR)/99-lxd-agent.rules \
+                $(TARGET_DIR)/usr/lib/udev/rules.d/99-lxd-agent.rules
```
agners, line 18: FWIW, there also linter errors on those two lines, see:
https://github.com/home-assistant/operating-system/actions/runs/3079333274/jobs/4979283480

bryanyork: Fixed / Removed those rules as I have the systemd patch in place now.

buildroot-external/package/lxd-guest-agent/99-lxd-agent.rules
bryanyork: Working systemd patchset added. We are behind the latest systemd version and they refactored things. I'll work on a PR to systemd for HEAD now.

bryanyork:
@agners @pvizeli Updated this PR to address issues and submitted upstream PR to systemd: https://github.com/systemd/systemd/pull/24802

bryanyork:
@agners @pvizeli Updates here from the systemd PR:

In later version of systemd past v250 series, LXD VM's are "properly" detected as KVM.
The creator / maintainer of LXD wants to keep it that way as changing will break other units expecting to see KVM. He again recommends targeting `ConditionPathExists=/dev/virtio-ports/org.linuxcontainers.lxd` as a test here.

I'm happy to keep the patch in, as v250 is still broken and detects qemu. I'll see if we can backport the fix to the v250 series.

I would prefer to get something in so I can follow mainline for my installation. Thoughts on next steps to merge this PR?

bryanyork:
Updated. I think this is the best option to merge. Please let me know your thoughts. It seems I can't convince folks at systemd or LXD to return as such, so we'll need to target like this to support my use-case. Perhaps in the future we will see a better conditional that is backwards compatible.

buildroot-external/package/qemu-guest-agent/qemu-guest.service
```diff
@@ -3,6 +3,8 @@ Description=QEMU Guest Agent
 After=syslog.target network.target
 ConditionVirtualization=|kvm
 ConditionVirtualization=|qemu
+# TODO: Remove Once https://github.com/systemd/systemd-stable/issues/218 is closed (And we can separate LXD and KVM)
```
agners, line 6: I'll remove the comment for now, as I think it is incorrect.

```suggestion
```

agners (changes requested):

agners (approved):

bryanyork:
Thanks for merging @agners ! Do you know when this will hit a production release?

bryanyork:
Hiya again @agners ! Should I create a PR to move this to the rel-9 branch? Was hoping to get this into 9.4 so I can move my install to a production version.

Issue #2168, opened 2022-10-03
Changing real-time schedulding priority not possible since HAos 9 for ODROID build


jens-maus:
### Describe the issue you are experiencing

Since HAos 9.x changing the real-time scheduling priority using, e.g. the `chrt` command, seem to be not possible for the ODROID build of HomeAssistantOS. This affects certain third-party HA addons like the "RaspberryMatic CCU" addon coming with binaries wanting to change the real-time scheduling priority using kernel calls like `sched_setscheduler()` (cf. https://github.com/jens-maus/RaspberryMatic/issues/1969)

### What operating system image do you use?

odroid-n2 (Hardkernel ODROID-N2/N2+)

### What version of Home Assistant Operating System is installed?

9.0

### Did you upgrade the Operating System.

No

### Steps to reproduce the issue

1. Install HomeAssistantOS 9.0 for a ODROID-based Hardware (e.g. ODROID-N2+)
2. change to the console and login to the OS console using `ha login`
3. chose a PID of a running process using `ps aux``
4. try to change the runtime scheduling priority using the `chrt -f -p 16 <PID>` command
5. Note, the `Operation not permitted` error messages
6. Redo the same, but after changing `kernel.sched_rt_runtime_us` to `-1` via `sysctl kernel.sched_rt_runtime_us=-1`
7. Retest changing the real-time schedulding priority and note that now no `Operation not permitted` error is shown.

### Anything in the Supervisor logs that might be useful for us?

```txt
n/a
```

### Anything in the Host logs that might be useful for us?

```txt
n/a
```

### Additional information

cf. https://github.com/jens-maus/RaspberryMatic/issues/1969
cf. https://github.com/jens-maus/RaspberryMatic/issues/1968#issuecomment-1264373923

PR #2169, opened 2022-10-03
Streamline CONFIG_RT_GROUP_SCHED

pvizeli:
Fix: #2168 

With cgroup2, we can not handle realtime schedule for now. Once we can, we can undo that change, but we should still have it global and not per board to make the same experience overall


buildroot-external/kernel/docker.config
```diff
@@ -4,6 +4,8 @@ CONFIG_CFS_BANDWIDTH=y
 CONFIG_FAIR_GROUP_SCHED=y
 CONFIG_NET_SCHED=y
 
+CONFIG_RT_GROUP_SCHED=n
+
```
agners, line 8: That doesn't work with Kconfig. `# CONFIG_RT_GROUP_SCHED is not set` should though.

agners (changes requested):

agners:
Yeah this config got removed in #1329, and should be removed for those two boards too. Good catch! Hm, we probably should use minimal defconfigs for those two boards as well :thinking:

buildroot-external/kernel/docker.config
pvizeli: good point :D

pvizeli:
> Hm, we probably should use minimal defconfigs for those two boards as well 🤔

Maybe, but that would be more somethings for REL-10 I guess or at least in a new PR

PR #2175, opened 2022-10-07
Update outdated ui references in issue template

mdegat01:
Updated UI guidance in issue template to match current UI. Also updated several of the my links as the system health report has what is being asked for, instead of the about info page (which is where Info goes).

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -40,9 +40,9 @@ body:
         - tinker (ASUS Tinker Board/Tinker Board S)
         - yellow (Home Assistant Yellow)
       description: >
-        Can be found in the [Configuration panel -> Settings -> Info](https://my.home-assistant.io/redirect/info/). It is listed as the `Board` value.
+        Can be found in [Settings -> System -> Repairs -> System Information](https://my.home-assistant.io/redirect/system_health/). It is listed as the `Board` value.
 
-        [![Open your Home Assistant instance and show your Home Assistant version information.](https://my.home-assistant.io/badges/info.svg)](https://my.home-assistant.io/redirect/info/)
+        [![Open your Home Assistant instance and show health information about your system.](https://my.home-assistant.io/badges/system_health.svg)](https://my.home-assistant.io/redirect/system_health/)
   - type: input
     validations:
       required: true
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -40,17 +40,17 @@ body:
...
+        [![Open your Home Assistant instance and show health information about your system.](https://my.home-assistant.io/badges/system_health.svg)](https://my.home-assistant.io/redirect/system_health/)
   - type: input
     validations:
       required: true
     attributes:
       label: What version of Home Assistant Operating System is installed?
       placeholder: "6.6"
       description: >
-        Can be found in the [Configuration panel -> Settings -> Info](https://my.home-assistant.io/redirect/info/). It is listed as the `Host Operating System` value.
+        Can be found in [Settings -> System -> Repairs -> System Information](https://my.home-assistant.io/redirect/system_health/). It is listed as the `Host Operating System` value.
```
agners, line 53:
```suggestion
        Can be found in [Settings -> System -> Repairs -> System Information (top right menu)](https://my.home-assistant.io/redirect/system_health/). It is listed as the `Host Operating System` value.
```

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -88,16 +89,19 @@ body:
     attributes:
       label: Anything in the Host logs that might be useful for us?
       description: >
-        Supervisor Logs can be found under [Configuration panel -> Add-ons, Backup & Supervisor -> System](https://my.home-assistant.io/redirect/supervisor_logs/), then choose Log Provider `Host`.
+        Host Logs can be found in [Settings -> System -> Logs](https://my.home-assistant.io/redirect/logs/)
+        then choose `Host` in the top right.
       render: txt
   - type: textarea
     attributes:
       label: System Health information
```
agners, line 97:
```suggestion
      label: System information
```

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -88,16 +89,19 @@ body:
...
       description: >
         **Optional** Copy the full System Health in this text area.
 
-        Can be found in the [Configuration panel -> Settings -> Info](https://my.home-assistant.io/redirect/info/).
-        Use the copy icon on top right and choose `For GitHub`.
+        System Health information can be found in the top right menu in [Settings -> System -> Repairs](https://my.home-assistant.io/redirect/repairs/).
```
agners, line 101:
```suggestion
        System information can be found in [Settings -> System -> Repairs -> System Information (top right menu)](https://my.home-assistant.io/redirect/system_health/),
```

.github/ISSUE_TEMPLATE/bug_report.yml
mdegat01: Ok I'll do this but then we need to correct the my link. It's confusing that we ask for "System information" but have a button called "System Health". I was trying to split the difference here.

.github/ISSUE_TEMPLATE/bug_report.yml
mdegat01: Sure, we just have the same link twice then. I was linking to repairs and then providing directions since there was already a button and figured then people would know where it was in the future. But probably overthinking this.

PR #2209, opened 2022-11-01
Enable Intel specific CPU idle driver

agners:


agners:
This can save up to 5W on Intel NUC systems when idle.

unclehack:
This looks good.

You may also want to add this patch https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?h=v6.0.6&id=e400ad8b7e6a1b9102123c6240289a811501f7d9. AMD systems still have to use the generic CPU idle driver. This fix helps AMD systems.

agners:
> You may also want to add this patch https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?h=v6.0.6&id=e400ad8b7e6a1b9102123c6240289a811501f7d9. AMD systems still have to use the generic CPU idle driver. This fix helps AMD systems.

I wonder why it did not got back-ported into stable kernels. I see that Ubuntu added it, so I am fine adding it as well.

That said, I did not find numbers on how much that affects power though. I am guessing its rather small in the end.

agners:
Adding this patch with #2210

unclehack:
> > You may also want to add this patch https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?h=v6.0.6&id=e400ad8b7e6a1b9102123c6240289a811501f7d9. AMD systems still have to use the generic CPU idle driver. This fix helps AMD systems.
> 
> I wonder why it did not got back-ported into stable kernels. I see that Ubuntu added it, so I am fine adding it as well.
> 
> That said, I did not find numbers on how much that affects power though. I am guessing its rather small in the end.

It's most likely something which avoids erratic behavior. I'm not sure how that workaround impacted VMs running on machines with AMD processors all of these years. It's not going to yield a performance improvement or reduced CPU usage.

PR #2210, opened 2022-11-02
Add patch to avoid dummy waits on CPU idle for AMD systems

agners:


unclehack:
This looks good. It's the same patch I've tested. HA OS builds and works just fine with this patch.

unclehack:
@agners: We need `CONFIG_X86_INTEL_PSTATE=y` for all x86_64 boards. This should provide better performance and better power management. The line `inl(acpi_gbl_FADT.xpm_timer_block.address);` ends up being executed without this config option on Intel machines. This is a problem because we're not using the Intel specific driver.

agners:
@unclehack `CONFIG_X86_INTEL_PSTATE=y` seems to be enabled for `generic-x86-64`.

Issue #2222, opened 2022-11-09
Firmware files are missing for broadcom adapters


bdraco:
### Describe the issue you are experiencing

related issue https://github.com/home-assistant/core/issues/78012

related update https://forums.raspberrypi.com/viewtopic.php?t=292113

files are available here https://github.com/winterheart/broadcom-bt-firmware

documented here https://github.com/home-assistant/home-assistant.io/pull/24865

### What operating system image do you use?

odroid-n2 (Hardkernel ODROID-N2/N2+)

### What version of Home Assistant Operating System is installed?

9.3

### Did you upgrade the Operating System.

Yes

### Steps to reproduce the issue

Install broadcom adapter, notice error in journal about missing firmware

### Anything in the Supervisor logs that might be useful for us?

```txt
n/a
```

### Anything in the Host logs that might be useful for us?

```txt
see above
```

bdraco:
This PR fixes the discovery https://github.com/home-assistant/core/pull/81926

`CONFIG_FW_LOADER_USER_HELPER_FALLBACK=y` is only enabled on the Odroid n2+ platform so this issue is most relevant there.

nagi121:
> https://github.com/home-assistant/operating-system/issues/2222#issuecomment-1308959053

I tried the workaround to make my broadcom adapter work again with no success. The mount does not work:

```
[core-ssh ~]$ mount -o bind /mnt/data/firmware /lib/firmware

mount: mounting /mnt/data/firmware on /lib/firmware failed: Permission denied
```
I have a SSD attached to ODROID-N2+ and moved all data to SSD using Home Assistant [http://homeassistant.local:8123/config/storage](url)

When will there be a permanent solution to this problem?

bdraco:
> When will there be a permanent solution to this problem?

I would pick up one of the CSR adapters from the High performance list at https://www.home-assistant.io/integrations/bluetooth/ since they don't need any patch files. 

I'm using the Feasycom FSC-BP119 (CSR8510A10) US $19.99 | 100m Long range Bluetooth USB Dongle Wireless Bluetooth 4.0 CSR Adapter for Windows 10, 8, 7 XP vista
https://a.aliexpress.com/_mM1DlzG

nagi121:
The BCM firmware patch files fix a security issue.
Buying other hardware is only an intermediate solution, because in the future the new hardware may also need firmware updates to work. 

[...]

PR #2225, opened 2022-11-10
Symlink firmware update directory to Supervisor writeable location

agners:


pvizeli (commented):
You will build an firmware plugin that manage firmware on that folder? I like that 👍

pvizeli:
As the folder is inside supervisor, we can also manage the folder directory on supervisor for UI and cli support of that feature. Maybe provide urls/updates over an repository to select firmware

dbuezas:
FYI: this doesn't seem to be working as expected, report here https://github.com/home-assistant/operating-system/issues/2222#issuecomment-1509948230

Issue #2178, opened 2022-10-10
Yellow not booting from NVMe


agners:
### Describe the issue you are experiencing

rolfl reported on Discord that Yellow won't boot using a SK Hynix NVMe. The green LED quickly flashes then change to solid (green and red). Update to latest Firmware did not seem to help.

The particular NVMe:
- SK hynix BC711 128GB PCIe NVMe M.2 2230 Gen 3 x 4 SSD, 0X3K2X, HFM128GD3GX013N

### What operating system image do you use?

yellow (Home Assistant Yellow)

### What version of Home Assistant Operating System is installed?

9.0

### Did you upgrade the Operating System.

No

### Steps to reproduce the issue

1. Use SK hynix BC711 128GB PCIe NVMe
2. Flash HAOS on it
3. Try starting

### Anything in the Supervisor logs that might be useful for us?

```txt
No
```

### Anything in the Host logs that might be useful for us?

```txt
No
```

RobPijpers:
Several other user including myself experience the same issue, more info on the forum

TimWeyand:
Hi @agners,
I don't have the yellow (yet), but have the same nvme problems with the waveshare baseboard a & b. 

As the Problem seems to be u-boot (?). I tried to figure out how this is working and if I could try/debug some stuff. I have no clue how I can build a whole HAOS image.

I might have found one thing in the u-boot configuration - maybe it's nothing - but can you take a look?. They seem to have begun to split up NVME in December 21.

> CONFIG_NVME Enable NVMe device support CONFIG_NVME_PCI Enable PCIe NVMe device support CONFIG_CMD_NVME Enable basic NVMe commands

I See a lot of changes in https://github.com/u-boot/u-boot/blob/master/configs/ changing from CONFIG_NVME to CONFIG_NVME_PCI.

Could that be by any chance the problem/solution for cm4?

honkasalo:
I'm having exactly same issue with following hardware:
- CM4 4/16Gb
- SK Hynix BC711 256Gb nvme PCIe gen 3x4
- Waveshare CM4-IO-BASE-A

All the full number updates ending to zero (8.0, 9.0) so far have introduced boot problem, namely absolutely no boot from nvme. Had to revert back to 8.5 from 9.0, didn't even try 9.2 anymore before knowing this nvme gets fixed.

Boot problem is not limited to Yellow.

agners:
[...]

PR #2236, opened 2022-11-17
Bump Raspberry Pi kernel and firmware to 1.20221104 (#2178)

agners:


boesing:
Just as I am curious: does haos update firmware again? I remember it got disabled due to too many issues?

agners:
> Just as I am curious: does haos update firmware again? I remember it got disabled due to too many issues?

Currently no. Yeah might be worth try it again. Maybe something to consider for HAOS 10.

The Home Assistant OS Installer for Yellow does though (in case you are using a Yellow).

Issue #2135, opened 2022-09-16
/etc/docker/daemon.json is read-only in HA OS 9.0


unclehack:
### Describe the issue you are experiencing

/etc/docker/daemon.json used to be writable before version 9.0. This facilitated the customisation of the Docker daemon startup flags.

This is no longer possible with HA OS 9.0. Please make this possible again. HA OS is already somewhat hostile when it comes to debugging.

The alternative is to fork HA OS and keep building new images. It doesn't really make sense to maintain a fork for one or two things which need to be modified. I could just migrate to the HA core container instead.

### What operating system image do you use?

generic-x86-64 (Generic UEFI capable x86-64 systems)

### What version of Home Assistant Operating System is installed?

9.0

### Did you upgrade the Operating System.

Yes

### Steps to reproduce the issue

1. try to edit /etc/docker/daemon.json
2. the editor says that the file is read-only
...

### Anything in the Supervisor logs that might be useful for us?

```txt
n/a
```

### Anything in the Host logs that might be useful for us?

```txt
n/a
```

agners:
This change has been on purpose, see https://github.com/home-assistant/operating-system/pull/2116. The reason this folder was writeable is for Docker to store the `key.json` file. This is now redirected to a wrietable directory. I did not anticipate that people would alter configuration of the Docker daemon.

What is your use case/what do you intend to change in `daemon.json`?

> HA OS is already somewhat hostile when it comes to debugging.

It is completely unsuited for debugging, IMHO :sweat_smile: . But that is the nature of a largely stateless, read-only system. :man_shrugging: 

> The alternative is to fork HA OS and keep building new images. It doesn't really make sense to maintain a fork for one or two things which need to be modified. 

So you use this altered config not just for debugging then? I am really curious what kind of changes you need. Ideally we make them "unnecessary" or configurable through a flag.

> I could just migrate to the HA core container instead.

I have a HAOS build directory which I keep rather current. A rebuild is usually quite quick, so for me altering a system service for debugging purpose isn't that big of a deal. But I understand, that is a bit a hassle. Using a regular Debian & Supervised is probably the way to go if you need custom system service configurations which we can't ship by default.

unclehack:
> This change has been on purpose, see #2116. The reason this folder was writeable is for Docker to store the `key.json` file. This is now redirected to a wrietable directory. I did not anticipate that people would alter configuration of the Docker daemon.
> 
> What is your use case/what do you intend to change in `daemon.json`?

[...]

PR #2246, opened 2022-11-25
Decrease network size of Docker default bridge (#2135)

agners:
Instead of allocating `172.17.0.0/16` as the default network bridge use a much smaller range. This helps folks which have hosts in that network range.

Ideally, we should make the default as well as the hassio bridge configurable.

### DIFF
--- a/buildroot-external/rootfs-overlay/etc/docker/daemon.json
+++ b/buildroot-external/rootfs-overlay/etc/docker/daemon.json
@@ -7,5 +7,6 @@
         "tag": "{{.Name}}"
     },
     "data-root": "/mnt/data/docker",
-    "deprecated-key-path": "/mnt/overlay/etc/docker/key.json"
+    "deprecated-key-path": "/mnt/overlay/etc/docker/key.json",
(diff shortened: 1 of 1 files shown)


pvizeli (approved):

agners:
This relates to #2135.

SolarstormTM:
The class b private adresses can go from 172.16.0.0 - 172.31.255.255 maybe picking the a space like 172.25.0.0/16 would slim the chances of someone using that subnet.

agners:
@SolarstormTM right good point, we use `172.30.32.0/23` for the `hassio` bridge network. So I'd suggest to use `172.30.232.1/23` then, so we use the same class B network.

agners:
PR https://github.com/home-assistant/operating-system/pull/2259

dilorenzo1987:
It would be helpful to also mention the range change in the title because it broke by SSH Port Forwarding Addon :)

ThomDietrich:
@dilorenzo1987 fixed via https://github.com/ThomDietrich/home-assistant-addons/pull/11 Thanks @hnykda

PR #2250, opened 2022-11-29
Use iptables with NFT backend

agners:


agners:
@he2ss this might affect your crowdsec add-on

agners:
While this is a rather large change for rel-9 stable release, it is necessary to make the OTBR work with the Docker IPv6 extensions enabled.

We could adjust the OTBR to work with iptables in legacy mode, but that seems backwards. So we'd rather like to move forward, and start adopting iptables with nf_tables asap.

PR #2259, opened 2022-12-10
Use same class B network for Docker as Supervisor (#2246)

agners:
Use a subnet in the same class B network for the Docker default bridge is using. This avoids conflicting with more than one class B network.

### DIFF
--- a/buildroot-external/rootfs-overlay/etc/docker/daemon.json
+++ b/buildroot-external/rootfs-overlay/etc/docker/daemon.json
@@ -8,5 +8,5 @@
     },
     "data-root": "/mnt/data/docker",
     "deprecated-key-path": "/mnt/overlay/etc/docker/key.json",
-    "bip": "172.17.232.1/23"
+    "bip": "172.30.232.1/23"
 }


wtremmel:
There is no such thing as a Class-B network. I assume you mean a /16.

austwhite:
@wtremmel   Might be worth doing your research on Class A, Class  B and  Class C networks. :)

austwhite:
@wtremmel   While technically correct, it is not uncommon to use the term Class B or Class C network even if they are IP Ranges

forkwhilefork:
@austwhite in 2023 it certainly is uncommon (unless you only talk to really old people I guess). We should try to avoid archaic and incorrect terminology.

agners:
> @austwhite speaking as a network engineer, in 2023 it certainly is uncommon (unless you only talk to really old people I guess). 

Ok, we are in classless world since forever. But was the "really" really necessary? Because now I really feel really old :cry: :sweat_smile: 

Since we have all the smart people here: But the network I define is a `/23` network. So technically "same /16 network" is wrong, too. It is just the same first two octets. How do I call that now? Same /16 prefix? Same /16 CIDR block?

wtremmel:
Problem is not old - I am 56. And also a network engineer. You can call the /16 the _parent prefix_ for example.

Issue #2256, opened 2022-12-07
aarch64 image converted from img to vhdx not working on Hyper-V ARM64 (stuck at waiting for root device)


ms300:
### Describe the issue you are experiencing

When i convert aarch64 images from img to vhdx and run it from Hyper-V,
the screen stucks at **waiting for root device**
![image](https://user-images.githubusercontent.com/3821012/206240822-ed66ec46-ebe5-4502-b9d9-ebbad4ddef9f.png)

i've tried 8.4 9.4 and 9.3, none of them were working. They're all stucked at **waiting for root device**

### What operating system image do you use?

generic-aarch64 (Generic UEFI capable aarch64 systems)

### What version of Home Assistant Operating System is installed?

8.4-9.4

### Did you upgrade the Operating System.

Yes

### Steps to reproduce the issue

1.convert vmdk, img or qcow2 to vhdx  using qemu-img
2.set the vm following https://www.home-assistant.io/installation/windows/
3.press run

### Anything in the Supervisor logs that might be useful for us?

```txt
N/A
```

### Anything in the Host logs that might be useful for us?

```txt
N/A
```

### Additional information

Device: Windows Dev Kit 2023(https://www.microsoft.com/en-us/d/windows-dev-kit-2023/94k0p67w7581)
converting command: `qemu-img.exe convert haos_generic-aarch64-9.3.img -O vhdx -o subformat=dynamic haos_generic-aarch64-9.3.vhdx`

agners:
It seems the Linux kernel cannot access the disk drive. What type of disk (controller) emulation are you using?

dro159:
@ms300 Did you find a work around yet by chance? I have the exact same issue on the exact same device but didn't use qemu to convert. I used StarWind V2V Converter.

ms300:
> It seems the Linux kernel cannot access the disk drive. What type of disk (controller) emulation are you using?

SCSI controller.
IDE controller is not allowed in generation 2 VM which is the only choice for windows 11 arm64.

agners:
Do you happen to know what type of SCSI controller Hyper-V is emulating?

agners:
It seems that Hyper-V is supporting a couple of SCSI controllers, I wonder which one it's using by default:
https://learn.microsoft.com/en-us/powershell/module/virtualmachinemanager/new-scvirtualscsiadapter?view=systemcenter-ps-2022#-scsicontrollertype

ms300:
[...]

PR #2260, opened 2022-12-10
Enable Fusion driver for LSI Logic SCSI controllers (#2256)

agners:


agners:
Hyper-V virtualization seems to use Hyper-V specific storage paravirtualization driver. This driver likely won't be useful on aarch64.

PR #2304, opened 2023-01-15
WiFi Access Point / HotSpot management in NetworkManager

xonestonex:
There are currently already Add-ons, that allow the usage of a WiFi interface in HomeAssistant as access point. They use hostapd to realize this functionality and create a separate Network for the connected clients. At least the Shelly integration doesn’t work correctly with clients connected to such a WiFi network.
A stupid WiFi-ethernet bridge also can’t be realized without problems, because in this case the ethernet interface has to be removed from NetworkManager and added to a bridge interface that is not managed by NetworkManger. This results in problems, because HomeAssistant thinks it’s offline.
This two patches allow NetworkManager to configure WiFi interfaces as access point and optionally bridge them with the ethernet interface.


buildroot-external/rootfs-overlay/etc/NetworkManager/NetworkManager.conf
```diff
@@ -5,7 +5,7 @@ autoconnect-retries-default=0
 rc-manager=file
 
 [keyfile]
-unmanaged-devices=type:bridge;type:tun;driver:veth
+unmanaged-devices=type:bridge;type:tun;driver:veth;except:interface-name:bridge*
```
agners, line 8: So with this bridge named devices are allowed.

Does removing `type:bridge` cause problems for the docker bridge? Maybe doing the reverse makes more sense: Remove `type:bridge` and add the interface named `hassio` and `docker*` to the unmanaged list.

xonestonex: I don't know the reason why all bridge interfaces have been added to the unmanaged list. So I've provided the a minimal change, that shouldn't break anything.

agners: Ok, but this means you need to name the bridge something starting with `bridge..`. Examples often use `br0`, which would not work now right?

I only see Docker as the reason why it has been set as unmanaged. But I'd say its better to reverse the logic now, so:

```suggestion
unmanaged-devices=type:tun;driver:veth;interface-name:docker*;interface-name:hassio
```

xonestonex: Do you have any other objections against this PR?

PR #2358, opened 2023-02-15
Linux: Update kernel 6.1.12 for non-Raspberry Pi systems

agners:


agners:
@jens-maus I still struggle to get `generic-raw-uart` and `eq3_char_loop` to build. These things are really quite annoying to maintain, out-of-tree kernel modules, with patches on-top of them. It seems that even latest upstream doesn't build on 6.1 currently. I am inclined to just disable for now and hope someone tackles the issues before the release :wink:

agners:
It seems that most problems got introduced by `CONFIG_WERROR`. It is not quite clear why it wasn't a problem with 5.15, since CONFIG_WERROR defaults to yes since 5.15. But in any case, I added now variables which ensure that `CONFIG_WERROR` is not selected for the modules which have compiler warnings (`generic_raw_uart` and `rtl8821cu).

jens-maus:
> @jens-maus I still struggle to get `generic-raw-uart` and `eq3_char_loop` to build. These things are really quite annoying to maintain, out-of-tree kernel modules, with patches on-top of them. It seems that even latest upstream doesn't build on 6.1 currently. I am inclined to just disable for now and hope someone tackles the issues before the release :wink: 

Sorry for the delay, but I took some days off for skiing..

Actually, haven't tested generic_raw_uart with kernel 6.1 yet simply because I thought as long as kernel 6.1 isn't officially promoted as the new LTS variant it is not worth the work. But as you have merged this PR now I think I will invest some time into this myself ASAP and report back.

agners:
> Sorry for the delay, but I took some days off for skiing..

No problem.

> 
> Actually, haven't tested generic_raw_uart with kernel 6.1 yet simply because I thought as long as kernel 6.1 isn't officially promoted as the new LTS variant it is not worth the work. But as you have merged this PR now I think I will invest some time into this myself ASAP and report back.

While 6.1 was expected to be the next LTS it only got promoted a few days ago. But is listed as LTS officially now: https://kernel.org/category/releases.html

It seems the problem was mostly about compile time warnings. I guess ideally those should be addressed at one point, but there are also out-of-tree Realtek Wi-Fi drivers affected.

In general, just shows how much of a paint these out-of-tree driver business is :cry:

PR #2375, opened 2023-02-28
Increase net.core.optmem_max for OTBR

agners:
The OTBR install scripts by default increases the net.core.optmem_max ancillary buffer size to 64KiB to allow for a larger number of multicast groups. Arch Linux as well recommends this size for high speed network links.


agners:
See https://github.com/openthread/ot-br-posix/blob/d3d0797/script/_rt_tables#L51-L54

Issue #2174, opened 2022-10-07
Ethernet network device name changed after updating to OS 9.1


rccoleman:
### Describe the issue you are experiencing

I just updated to HA OS 9.1, which I run in a KVM running on Debian 11, and the network failed to come up.  From the console, it looks like the ethernet device changed, for some reason, from enp0s3 to ens3, and there was no longer an active network interface.  I added another interface using nmcli con edit, assigned the same static IP and gateway that I had before, and it's back up and running.  Now my connections look like this: 

![image](https://user-images.githubusercontent.com/860888/194476192-646eb1fb-876f-4a9b-9c69-574337ec1593.png)

I rebooted and it came up fine, but I wonder what might have happened there.  I've gone through other HA OS updates and rebooted the host machine many times without incident

The only changes since 9.0 that look interesting are these:
https://github.com/home-assistant/operating-system/pull/2138
https://github.com/home-assistant/operating-system/pull/2171

I don't know if they're related, but I suspect that it would be hard for most folks to recover from.

### What operating system image do you use?

ova (for Virtual Machines)

### What version of Home Assistant Operating System is installed?

9.1

### Did you upgrade the Operating System.

Yes

### Steps to reproduce the issue

I was running HA OS 9.0 in a KVM on Debian 11 and it was working fine.  As soon as I updated to HA OS 9.1 based on an update alert, the KVM started without a functional network interface.

### Anything in the Supervisor logs that might be useful for us?

```txt
There were lots of errors in the Supervisor log when it didn't have a functional network interface, but I doubt they'll be of interest.
```

### Anything in the Host logs that might be useful for us?

```txt
Nope, ask if needed.
```

## System Information

version | core-2022.10.1
-- | --
installation_type | Home Assistant OS
dev | false
hassio | true
docker | true
user | root
virtualenv | false
python_version | 3.10.5
os_name | Linux
os_version | 5.15.72
arch | x86_64
timezone | America/Los_Angeles
config_dir | /config

<details><summary>Home Assistant Community Store</summary>

GitHub API | ok
-- | --
GitHub Content | ok
GitHub Web | ok
GitHub API Calls Remaining | 4723
Installed Version | 1.27.2
Stage | running
Available Repositories | 1127
Downloaded Repositories | 33

</details>

<details><summary>Home Assistant Cloud</summary>

logged_in | true
-- | --
subscription_expiration | February 15, 2023 at 4:00 PM
relayer_connected | true
remote_enabled | true
[...]

PR #2380, opened 2023-03-02
Use Network device naming scheme v250

agners:
The new systemd version v252 brings a new naming scheme, in particular it seems that on device tree based systems (e.g. Raspberry Pis) the Ethernet device name changes from eth0 to end0.

This breaks a previously made configuration.

Even worse, it seems that the default NetworkManager behavior is to only configure a network device if there is no profile. But since profiles are configured on a typical installation, NetworkManager doesn't bring up any of the network interface, leaving the user stranded on an unconnected system.

Ideally, we should have a plan how to migrate from one naming scheme to the next. For now, just stick with the naming scheme HAOS 9.x has been using.

### DIFF
--- a/buildroot-external/board/arm-uefi/generic-aarch64/grub.cfg
+++ b/buildroot-external/board/arm-uefi/generic-aarch64/grub.cfg
@@ -56,7 +56,7 @@ fi
 
 save_env A_TRY A_OK B_TRY B_OK ORDER MACHINE_ID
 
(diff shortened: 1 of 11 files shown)


agners:
We've previously encountered a similar issue in https://github.com/home-assistant/operating-system/issues/2174#issuecomment-1272025622. Ideally we should find a way to deal with this type of network interface name changes.

Issue #2376, opened 2023-03-01
Hyper-V Snapshots Failing


scyto:
### Describe the issue you are experiencing

I am trying to take a backup of hassos on Hyper-V using synolgy active backup.
While i can take a non-applications aware snapshot the application aware snapshots are failing.

I followed t[his troubleshooting article](https://kb.synology.com/en-us/DSM/tutorial/What_can_I_do_when_production_checkpoint_cannot_be_created_ABB) and confirmed i cannot take a snapshot in the hyperv ui. This article implied that the services required for checkpoint creation are not installed or running.

I then looked to t[his Microsoft article](https://learn.microsoft.com/en-us/windows-server/virtualization/hyper-v/manage/manage-hyper-v-integration-services#start-and-stop-an-integration-service-from-a-linux-guest) to identify if the services were running.

running `lsmod | grep hv_utils` gives an output as follows:

```
➜  ~ lsmod | grep hv_utils
hv_utils               45056  0 
```
this is different to the example outpu in the MS article (no hv_vmbus) listed

running `ps -ef | grep hv` gives output as follows: 
<img width="476" alt="image" src="https://user-images.githubusercontent.com/11369180/222025952-85628b81-ce07-46b3-931c-023541347a49.png">

This would indicate that the vss daemon (hv_vss_daemon is not running (and the kvp daemon too as just a point of note)

I can't run compgen as recommended in the MS article as that package isn't installed in hassos

tl;dr it seems the VSS daemon is not installed or not running for some reason - i will keep digging through docs but this is the first time i have dug deep on the daemons.

### What operating system image do you use?

generic-x86-64 (Generic UEFI capable x86-64 systems)

### What version of Home Assistant Operating System is installed?

9.5

### Did you upgrade the Operating System.

Yes

### Steps to reproduce the issue

Follow the **METHOD 1** steps in this article to reproduce the error in hyper-v

https://kb.synology.com/en-us/DSM/tutorial/What_can_I_do_when_production_checkpoint_cannot_be_created_ABB

### Anything in the Supervisor logs that might be useful for us?

```txt
nothing of note
```

### Anything in the Host logs that might be useful for us?

```txt
nothing of note, all the logs seem to be about docker containers
```

## System Information

version | core-2023.2.3
-- | --
installation_type | Home Assistant OS
dev | false
hassio | true
docker | true
user | root
virtualenv | false
python_version | 3.10.7
os_name | Linux
os_version | 5.15.90
arch | x86_64
[...]

Issue #1996, opened 2022-06-28
Unable to take application consistent production checkpoint in HyperV - HAOS 8.2


Locknes:
### Describe the issue you are experiencing

I am not able to take production checkpoints of my HAOS VM, with the operation failing with the error in the image below.

### What operating system image do you use?

ova (for Virtual Machines)

### What version of Home Assistant Operating System is installed?

Home Assistant OS 8.2

### Did you upgrade the Operating System.

Yes

### Steps to reproduce the issue

1. Enable Production Checkpoint in hyperV VM configuration
2. Disable "Create Standard checkpoint if the guest does not support creation of production checkpoints" box
3. Try to take checkpoint
...
![image](https://user-images.githubusercontent.com/44587135/176275608-0d4f614c-2d2e-4b56-bc69-f1dbe8dc82d9.png)
![image](https://user-images.githubusercontent.com/44587135/176275719-6103dbe2-b902-48f6-aafd-ae1a843415af.png)

### Anything in the Supervisor logs that might be useful for us?

```txt
22-06-28 14:03:59 INFO (MainThread) [supervisor.resolution.evaluate] Starting system evaluation with state CoreState.RUNNING
22-06-28 14:03:59 INFO (MainThread) [supervisor.resolution.evaluate] System evaluation complete
22-06-28 14:03:59 INFO (MainThread) [supervisor.resolution.fixup] Starting system autofix at state CoreState.RUNNING
22-06-28 14:03:59 INFO (MainThread) [supervisor.resolution.fixup] System autofix complete
22-06-28 14:05:12 INFO (MainThread) [supervisor.updater] Fetching update data from https://version.home-assistant.io/stable.json
22-06-28 14:05:40 INFO (MainThread) [supervisor.store.git] Update add-on https://github.com/blakeblackshear/frigate-hass-addons repository
22-06-28 14:05:40 INFO (MainThread) [supervisor.store.git] Update add-on https://github.com/home-assistant/addons repository
22-06-28 14:05:40 INFO (MainThread) [supervisor.store.git] Update add-on https://github.com/hassio-addons/repository repository
22-06-28 14:05:40 WARNING (SyncWorker_0) [supervisor.addons.validate] Add-on have full device access, and selective device access in the configuration. Please report this to the maintainer of Frigate NVR (Full Access)
[...]

PR #2381, opened 2023-03-02
Enable Microsoft HyperV integration services

agners:


agners:
See also #2376 and #1996.

scyto:
wow, thanks, would you like me to test this now in any sort of way or just wait for this change to get into beta and then test?

agners:
Yes you can find dev builds here: https://os-builds.home-assistant.io/10.0.dev20230302/

You can also update an existing installation to this dev build using

```
ha su options --channel=dev
ha su reload
ha su update
ha os update
```

scyto:
I downloaded the new VHDX to my hyperv-server and spun up a new VM rather than touching my live install

I can report that production level checkpoints work perfectly now, thanks for doing this - this now means my synology backup process backs this VM p like i need it to :-)

also for your reference i can confirm that hyper-v gues services is now functionally equivalent to the experience on native debian 11 install in that the following doesn't work in hyper-v MMC UI:

1. type clipboard text doesn't work
2. capture screen returns the following

image

**to be clear i am saying if someone files a bug complaining about these two behaviours it IS NOT home assistant specific** - hopefully that might spare you sometime later if anyone reports those, lol :-)

agners:
Thanks for testing!

PR #2392, opened 2023-03-08
Enable Multi-Gen LRU

agners:
Multi-Gen LRU should improve performance under memory pressure. This is especially useful for embedded platforms where memory is scarce.


jens-maus:
You probably know it already, but for the RaspberryPi  `CONFIG_LRU_GEN` and `CONFIG_LRU_GEN_DEFAULT` is already enabled per default. See:

https://github.com/raspberrypi/linux/blob/02928eb7eb435c6ec86745dcbb6ff99992128149/arch/arm64/configs/bcm2711_defconfig#L75-L76

agners:
@jens-maus hm, no, I didn't know. In my testing, LRU_GEN actually stalled the system even longer in some cases :cry:  (my test was with our zram and setup, on OVA with 512MiB of memory). I had reports of Yellow being unresponsive, I now start to wonder if this might be a side effect of LRU_GEN :thinking: 

It seems it wasn't a very conscious decision from RPi folks, at least no reasoning when asked about LRU_GEN (see https://github.com/raspberrypi/linux/issues/5272#issuecomment-1348079922).

unclehack:
@agners: Would you be willing to provide your results to the developers who've written the multi-gen LRU? They'll certainly be interested in solving regressions and other issues. Perhaps the fixes will also be pulled into 6.1.

jens-maus:
@agners 
>As it stands now (Linux 6.1.20), Multi-Gen LRU doesn't seem to be helpful in OOM situations for HAOS.

BTW: Didn't check it yet in HAos, but do you already have a mechanism in place to adjust `oom_score_adj` to ensure critical HA services are the last ones being killed in OOM situations?

E.g. in RaspberryMatic I am adjusting `/proc/$$/oom_score_adj` before starting curial service like:

https://github.com/jens-maus/RaspberryMatic/blob/master/buildroot-external/overlay/base/etc/init.d/S62HMServer#L187-L189

This should ensure that these critical services will be the last ones being killed in case of OOM situations.

agners:
> @agners: Would you be willing to provide your results to the developers who've written the multi-gen LRU? They'll certainly be interested in solving regressions and other issues. Perhaps the fixes will also be pulled into 6.1.

The results are not really scientific, I just triggered some OOM situation using ESPHome compile task, which lead to longer hangs then I've experienced without. #2400 enables Multi-Gen LRU, but disabled by default. So you can actually reproduce with nightly OS builds (https://os-builds.home-assistant.io/) by enabling it manually:

```
echo y >/sys/kernel/mm/lru_gen/enabled
```

@yuzhaogoogle I think you are the author of the patchset. Any thoughts of what form a feedback might be helpful?

agners:
> BTW: Didn't check it yet in HAos, but do you already have a mechanism in place to adjust `oom_score_adj` to ensure critical HA services are the last ones being killed in OOM situations?

We set values in Supervisor: Supervisor, Core and plug-ins get higher values, add-ons lower. It is by no means fine tuned, but it currently helps to leave Supervisor/Core afloat when installing too many add-ons.

See: https://github.com/home-assistant/supervisor/pull/3781

yuzhaogoogle:
> > @agners: Would you be willing to provide your results to the developers who've written the multi-gen LRU? They'll certainly be interested in solving regressions and other issues. Perhaps the fixes will also be pulled into 6.1.
> 
> The results are not really scientific, I just triggered some OOM situation using ESPHome compile task, which lead to longer hangs then I've experienced without. #2400 enables Multi-Gen LRU, but disabled by default. So you can actually reproduce with nightly OS builds (https://os-builds.home-assistant.io/) by enabling it manually:
> 
> ```
> echo y >/sys/kernel/mm/lru_gen/enabled
> ```
> 
> @yuzhaogoogle I think you are the author of the patchset. Any thoughts of what form a feedback might be helpful?

What `min_ttl_ms` value did you try? **Larger** values lead to **faster** OOM kills. (There are times people misinterpreted it as the other way around.)

Also, was the swapfile full when your system became unresponsive or OOM kills happened?

agners:
> What `min_ttl_ms` value did you try? **Larger** values lead to **faster** OOM kills. (There are times people misinterpreted it as the other way around.)

I tried the documented ones (1000 and 3000). Later I tried with just 1. The Arm boards I tested with are not that fast. It seemed to me that even 1 was too "trigger happy" with a setting of 1.

> Also, was the swapfile full when your system became unresponsive or OOM kills happened?

Just to be clear: "became unresponsive" was with `min_ttl_ms=0`.

Just retested and checked free: It was full before the system became unresponsive. For this test I am using zram still (swap on a zram block device). The system gets responsive also without Multi-Gen LRU, but in this case for ~60s or so. With Multi-Gen LRU and `min_ttl_ms=0` it seems to take much longer (~8 minutes in my last test).

yuzhaogoogle:
> > What `min_ttl_ms` value did you try? **Larger** values lead to **faster** OOM kills. (There are times people misinterpreted it as the other way around.)
> 
> I tried the documented ones (1000 and 3000). Later I tried with just 1. The Arm boards I tested with are not that fast. It seemed to me that even 1 was too "trigger happy" with a setting of 1.

Thanks.

My understanding is that "trigger happy" means OOM kills happened before  your system became unresponsive. Is this correct?

And what behavior do you expect?

> > Also, was the swapfile full when the system became unresponsive or OOM kills happened?
> 
> Just to be clear: "became unresponsive" was with `min_ttl_ms=0`.

Yes, that's my understanding. (`min_ttl_ms=0` disables thrashing prevention.)

> Just retested and checked free: It was full before the system became unresponsive. For this test I am using zram still (swap on a zram block device). The system gets responsive also without Multi-Gen LRU, but in this case for ~60s or so. With Multi-Gen LRU and `min_ttl_ms=0` it seems to take much longer (~8 minutes in my last test).

Do you mean that the system first became unresponsive and then (probably after OOM kills) it became responsive again after about 60 secs (the non-MGLRU case) and 8 mins (the MGLRU case)?

A couple of points here:

1. If the swapfile was full, the system was likely thrashing the file pages. If there were only a handful of file pages, the system could easily hit the `min_ttl_ms` thrashhold.
2. Did you feel the "trigger happy" effect while using a terminal or watching videos? I suspect the former because lags are much less observable in that case. If so, please try to play a video and see whether `min_ttls_ms` works better than the baseline.

agners:
> My understanding is that "trigger happy" means OOM kills happened before your system became unresponsive. Is this correct?

Yes, it triggered well before it came unresponsive, so that is good. However, I felt it triggered OOM in cases where it shouldn't. As if it just killed because the system was a bit slow in general (testing on Raspberry Pi 3). It felt like 1 is "too high" for these type of systems.

> Do you mean that the system first became unresponsive and then (probably after OOM kills) it became responsive again after about 60 secs (the non-MGLRU case) and 8 mins (the MGLRU case)?

Correct.

Home Assistant OS is usually running headless, we don't have a graphical interface. All services are accessed through network.

yuzhaogoogle:
> > My understanding is that "trigger happy" means OOM kills happened before your system became unresponsive. Is this correct?
> 
> Yes, it triggered well before it came unresponsive, so that is good. However, I felt it triggered OOM in cases where it shouldn't. As if it just killed because the system was a bit slow in general (testing on Raspberry Pi 3). It felt like 1 is "too high" for these type of systems.

I don't think it's because `min_ttl_ms=1` is too high. It's very likely you hit the second `if`, not the first one. (It happens when you have less than 8MB reclaimable file pages when swapfile is full.)

```
	if (min_ttl) {
		int gen = lru_gen_from_seq(min_seq[LRU_GEN_FILE]);
		unsigned long birth = READ_ONCE(lruvec->lrugen.timestamps[gen]);

		if (time_is_after_jiffies(birth + min_ttl))
			return false;

		/* the size is likely too small to be helpful */
		if (!nr_to_scan && sc->priority != DEF_PRIORITY)
			return false;
	}
```

So in this case, either use a larger swapfile so that you won't hit that `if` or just delete it (which isn't a good idea): https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/mm/vmscan.c?h=v6.1#n4513

agners:
First off: Thanks for taking your time and join this discussion, this is really appriciated! :heart_eyes: 

Since we are talking about `min_ttl_ms` with non-zero values now: Does that mean that Multi-Gen LRU with `min_ttl_ms=0` is expected to trigger OOM not quicker than regular LRU?

Regarding extending swap, some background: HAOS is targeted for single-board computer type of systems running on flash memory (e.g. SD cards) with limited flash cycles. So far we have been using 25% of memory for zram swap space. Recently I moved to 33% zswap on storage with `vm.swappiness=1`. This will likely cause a bit of writes, but keeping swap space low and with the tuned swappiness typical work loads are expected to not cause a lot of additional writes. However, adding too much zswap will allow people to run much more on their systems, and will definitely cause more writes on storage as "active" memory starts to get swapped in and out. 

The lowest memory I am testing is 512MB RAM (+33% zswap). Do you think I run into situations where I have less than 8MB reclaimable file pages? How could I check that?

In general: I'd really like to enable Multi-Gen LRU and `min_ttl_ms`, as a responsive system is important for home automation. We don't typically have OOM situations, as the system is mostly running idle, but still, it would be really nice if OOM situations get resolved quickly.

What I am concerned about regressions: Essentially people claiming that things previously worked now lead to (premature) OOM situations. Is this a legitimate fear?

I did some more testing using ESPHome add-on to compile a firmware (essentially compiler/linker running in a container). Currently, with Raspberry Pi 3 (1GB RAM) + zswap seems to successfully run my test runs with and without Multi-Gen LRU `min_ttl_ms=10`, as well as up to 1000.

One thing I've noticed is that on RPi 3 `min_ttl_ms` reads back 10 even when setting 1. I guess that is due to `CONFIG_HZ=100`. Lower means less aggressive, so with a `CONFIG_HZ=100` system I have to run a more aggressive `min_ttl_ms`?

I've now run some more testing on KVM with different memory sizes, to compare behavior when running work loads right on the limit of OOM.

| Memory       | Vanilla LRU            | Multi-Gen LRU/`min_ttl_ms=1`    |
|-------------------|---------------------------|-------------------------------------------------------------|
| 500MB         | OOM at compile   | OOM at compile                                               |
| 600MB         | OOM at compile   | OOM at compile                                               |
| 760MB         | OOM at linking     | OOM at linking                                               |
| 790MB         | OOM at linking     | OOM at linking                                               |
| 800MB         | Working                 | Working                                                           |

My virtual machine the two LRUs seem to succeed or fail in the same memory configuration.

With these additional testing I am inclined to just enable it and see what the feedback is during our beta testing phase.

One thing I was wondering: There is no compile time option to set the default value for `/sys/kernel/mm/lru_gen/min_ttl_ms`. Also at runtime other mechanism such as sysctl are not available. What is the recommended way to set teh default of this value?

yuzhaogoogle:
> First off: Thanks for taking your time and join this discussion, this is really appriciated! heart_eyes
> 
> Since we are talking about `min_ttl_ms` with non-zero values now: Does that mean that Multi-Gen LRU with `min_ttl_ms=0` is expected to trigger OOM not quicker than regular LRU?

Yes, MGLRU is expected to try harder under heavy memory pressure.

> Regarding extending swap, some background: HAOS is targeted for single-board computer type of systems running on flash memory (e.g. SD cards) with limited flash cycles. So far we have been using 25% of memory for zram swap space. Recently I moved to 33% zswap on storage with `vm.swappiness=1`. This will likely cause a bit of writes, but keeping swap space low and with the tuned swappiness typical work loads are expected to not cause a lot of additional writes. However, adding too much zswap will allow people to run much more on their systems, and will definitely cause more writes on storage as "active" memory starts to get swapped in and out.

zram is a lot more popular than zswap for client devices. Android and ChromeOS both use zram. I only know two major zswap users: Google and Meta data centers.

So for HAOS, I highly recommend zram over zswap because of two concerns:

1. the SSD wearout
2. the waste of SSD space when zswap is underutilized

If HAOS plans to stick with zswap, please make sure it uses the zsmalloc backend. zsmalloc is in generally better supported and more advanced than zbud, z3fold, etc.
 
> The lowest memory I am testing is 512MB RAM (+33% zswap). Do you think I run into situations where I have less than 8MB reclaimable file pages? How could I check that?

Yes, if `/proc/swaps` shows it's nearly full. In this case, MGLRU will only consider file pages and it's likely file pages are below 8MB for small systems.

> In general: I'd really like to enable Multi-Gen LRU and `min_ttl_ms`, as a responsive system is important for home automation. We don't typically have OOM situations, as the system is mostly running idle, but still, it would be really nice if OOM situations get resolved quickly.

Agreed.

> What I am concerned about regressions: Essentially people claiming that things previously worked now lead to (premature) OOM situations. Is this a legitimate fear?

Yes, and please feel free to tag me when you get such reports.

> I did some more testing using ESPHome add-on to compile a firmware (essentially compiler/linker running in a container). Currently, with Raspberry Pi 3 (1GB RAM) + zswap seems to successfully run my test runs with and without Multi-Gen LRU `min_ttl_ms=10`, as well as up to 1000.
> 
> One thing I've noticed is that on RPi 3 `min_ttl_ms` reads back 10 even when setting 1. I guess that is due to `CONFIG_HZ=100`.

Correct.

> Lower means less aggressive, so with a `CONFIG_HZ=100` system I have to run a more aggressive `min_ttl_ms`?

Values below 500ms usually don't make much difference. Not hitting nearly full swapfile generally brings more reliable performance under heavy memory pressure.

> I've now run some more testing on KVM with different memory sizes, to compare behavior when running work loads right on the limit of OOM.
> 
> Memory	Vanilla LRU	Multi-Gen LRU/`min_ttl_ms=1`
> 500MB	OOM at compile	OOM at compile
> 600MB	OOM at compile	OOM at compile
> 760MB	OOM at linking	OOM at linking
> 790MB	OOM at linking	OOM at linking
> 800MB	Working	Working
> My virtual machine the two LRUs seem to succeed or fail in the same memory configuration.

It'd be nice to make sure the OOM kills were not caused by swapfile depletion.

> With these additional testing I am inclined to just enable it and see what the feedback is during our beta testing phase.

Sounds good. Again, please feel free to tag me if there is anything I could be of help.

> One thing I was wondering: There is no compile time option to set the default value for `/sys/kernel/mm/lru_gen/min_ttl_ms`.

Correct.

> Also at runtime other mechanism such as sysctl are not available. What is the recommended way to set teh default of this value?

initscripts :)

agners:
Thanks for your feedback!

> So for HAOS, I highly recommend zram over zswap because of two concerns:
> 
>     1. the SSD wearout
>     2. the waste of SSD space when zswap is underutilized

I've seen one other user of zswap which actually inspired me to do the switch: Ubuntu (specifically on Raspberry Pi, see [this blog post](https://ubuntu.com/blog/how-low-can-you-go-running-ubuntu-desktop-on-a-2gb-raspberry-pi-4)).

We do run a SQLite database on that same storage. My take currently is that the database likely leads to much more wear than zswap, especially when tuned to the lower end (I set it to 33% of physical memory currently, the idea being that the smaller zswap the less likely it gets that "actively used" memory is being swapped out).

We do have folks moving away from HAOS on low memory platforms because it simply wouldn't run with just physical memory + zram. With been allocated 25% of physical memory to zram so far.

> Yes, and please feel free to tag me when you get such reports.

Awesome, thanks!

> It'd be nice to make sure the OOM kills were not caused by swapfile depletion.

What do you mean? Swap in those cases was definitely pretty much completely used up:

```
# cat /proc/swaps 
Filename                                Type            Size            Used            Priority
/mnt/data/swapfile                      file            327724          326924          -2
```

agners:
> This means the swap size is the bottleneck, not the CPU or disk speed. Increasing swapfile size might help (a lot).

Uh yes, swap is the bottleneck for sure in this test case.

Home Assistant with it's add-on system is quite scalable: You can install multiple databases (e.g. InfluxDB for long time storage), alternative frontends (Graphana) and automation tools (e.g. Node RED) and tons of software to extend home automation. On a Intel NUC running 10-20 add-ons is no problem.

However, we don't prevent people from installing the same add-ons on a Raspberry Pi 3 with 1GB of RAM... And quite some just "overload" their system :sweat_smile: 

So I try to make the right trade-offs between:
- Support as much as a low end system can handle (by adding some swap on disk)
- Tune swap such that it doesn't lead to too much flash wearout (by not adding too much swap on disk)
- Handle OOM situation if they occur quickly (this attempt)

The overloaded case is what I try to simulate in my testing trying to find a good trade off for these three points.

Issue #2390, opened 2023-03-07
Yellow reboot on CM4 EMMC and powercycle on NVME disk


Antexa:
### Describe the issue you are experiencing

I got a strange behavior on my Yellow, if I go with a system reboot, Yellow boot on CM4 EMMC and if I power cycle the Yellow board it boots on NVME drive. 

this behavior appear after a factory reset (I press the two button on the board, the red and the blue according to this : 
https://yellow.home-assistant.io/faq/#can-i-reinstall-home-assistant-os-again-using-the-usb-flash-drive-with-the-home-assistant-os-installer-for-yellow

<img width="1045" alt="image" src="https://user-images.githubusercontent.com/8592053/223397814-0fb423b9-ccb1-49c4-b728-03ea39629a0a.png">

### What operating system image do you use?

yellow (Home Assistant Yellow)

### What version of Home Assistant Operating System is installed?

9.5

### Did you upgrade the Operating System.

Yes

### Steps to reproduce the issue

1. System reboot go into CM4 EMMC
2. Powercycle the yellow board go into NVME (as expected!) 

### Anything in the Supervisor logs that might be useful for us?

```txt
[32m23-03-07 09:31:23 INFO (MainThread) [supervisor.resolution.check] Starting system checks with state CoreState.RUNNING[0m
[32m23-03-07 09:31:23 INFO (MainThread) [supervisor.resolution.checks.base] Run check for IssueType.SECURITY/ContextType.CORE[0m
[32m23-03-07 09:31:23 INFO (MainThread) [supervisor.resolution.checks.base] Run check for IssueType.DNS_SERVER_FAILED/ContextType.DNS_SERVER[0m
[32m23-03-07 09:31:24 INFO (MainThread) [supervisor.resolution.checks.base] Run check for IssueType.PWNED/ContextType.ADDON[0m
[32m23-03-07 09:31:24 INFO (MainThread) [supervisor.resolution.checks.base] Run check for IssueType.FREE_SPACE/ContextType.SYSTEM[0m
[32m23-03-07 09:31:24 INFO (MainThread) [supervisor.resolution.checks.base] Run check for IssueType.IPV4_CONNECTION_PROBLEM/ContextType.SYSTEM[0m
[32m23-03-07 09:31:24 INFO (MainThread) [supervisor.resolution.checks.base] Run check for IssueType.TRUST/ContextType.SUPERVISOR[0m
[...]

PR #2410, opened 2023-03-14
Deactivate any external data disk device on first boot (#2390)

agners:


jov58:
Hi, I have the same behaviour as [here](https://github.com/home-assistant/operating-system/issues/2390). I posted my questions there but got no reaction so far. So I'm trying here.

My most important question is: can you predict what this commit will mean in my case? I'm a bit afraid of updating to V10 now because I don't know if my config will survive the update and/or if I can restore it.

As I said I have the same behaviour only the other way around: if I power-cycle, Yellow boots and uses eMMC. If I then reboot through the System Menu, it boots using NVMe. This happens most of the time, but occasionally Yellow boots with NVMe after a power-cycle or with eMMC after a restart.

So, my "normal" state running Yellow is when the system is using NVMe (faster than eMMC).

What will happen in my case, will Yellow boot with eMMC and refuse to use my NVMe? If so, how do I proceed after that and is it possible to restore my complete config (with add-ons and everything) from a full backup of the NVMe?

I also don't understand the preference for eMMC over NVMe. Why not deactivate eMMC when a NVMe is found? NVMe is much, much faster than eMMC and seems to wear out less quickly.

Isn't it possible to wipe the eMMC, flash the Home Assistant Operating System 10.0 to a NVMe and boot with that, without ever using eMMC?

Lot of questions, any help in understanding the boot procedure really appreciated.

PR #2418, opened 2023-03-17
Linux: Update kernel 6.1.20

agners:


m32w:
I do vote for this improvement !

PR #2420, opened 2023-03-17
Use zswap instead of swap in zram

agners:
This requires a swap file which will get generated automatically on startup.


jens-maus:
Interesting that you also thought about using zswap rather than zram. However, two things I wonder is:

- Is there somewhere a comparison between zswap and zram regarding performance and compression ratio, etc?!?
- Now that you implemented zswap with a swapfile I wonder if this would nit have a negative impact on wear level of sd card or ssd storage usage? Compared, zram is completely handled in RAM while zswap is using a swapfile and might have a negative impact here which is IMHO undesirable in the context of embedded systems.

agners:
> Is there somewhere a comparison between zswap and zram regarding performance and compression ratio, etc?!?

Yeah I am also not aware of a comparison. I'd guess that compression performance is quite comparable. The big difference is storing it in RAM or on disk...

> Now that you implemented zswap with a swapfile I wonder if this would nit have a negative impact on wear level of sd card or ssd storage usage? Compared, zram is completely handled in RAM while zswap is using a swapfile and might have a negative impact here which is IMHO undesirable in the context of embedded systems.

Yeah this is *very* hard to gauge. With zswap it is possible to have more memory available (since the compressed part doesn't take up memory space) and the swap space can also be larger. Swap is only really actively used when memory get scarce. If it is too scarce, the page cache gets flushed, which can cause additional writes (e.g. journal writes I'd imagine would be written out in much smaller chunks all the time in memory pressure). 

There is always some allocated memory which is really not in use. So such memory will get swapped out, and just stay there. So that is essential one write cycle per boot, and from then on the additional memory can be used for better page cache handling.

But again, in the end its very hard to gauge.

In the past HAOS/Home Assistant in general used to have issues with wearing out SD-cards very quickly. We had many write cycles with how the recorder integration (with sqlite backend) was working: It was doing a lot of commits, and since sqlite is ACID, the durability part of it caused lots of actual writes... It probably caused writes with every state change of an entity.  These days sqlite is much better tuned, batching inserts and using write-ahead logs.

jens-maus:
> > Now that you implemented zswap with a swapfile I wonder if this would nit have a negative impact on wear level of sd card or ssd storage usage? Compared, zram is completely handled in RAM while zswap is using a swapfile and might have a negative impact here which is IMHO undesirable in the context of embedded systems.
> 
> Yeah this is _very_ hard to gauge. With zswap it is possible to have more memory available (since the compressed part doesn't take up memory space) and the swap space can also be larger. Swap is only really actively used when memory get scarce. If it is too scarce, the page cache gets flushed, which can cause additional writes (e.g. journal writes I'd imagine would be written out in much smaller chunks all the time in memory pressure).
> 
> There is always some allocated memory which is really not in use. So such memory will get swapped out, and just stay there. So that is essential one write cycle per boot, and from then on the additional memory can be used for better page cache handling.
> 
> But again, in the end its very hard to gauge.
> 
> In the past HAOS/Home Assistant in general used to have issues with wearing out SD-cards very quickly. We had many write cycles with how the recorder integration (with sqlite backend) was working: It was doing a lot of commits, and since sqlite is ACID, the durability part of it caused lots of actual writes... It probably caused writes with every state change of an entity. These days sqlite is much better tuned, batching inserts and using write-ahead logs.

I actually still think that it would be better to keep on using zram to simply avoid any further regular write operation on the underlying storage disks, which could have quite severe negative effects on e.g. SD cards. In addition, putting swap into zram is IMHO quite an elegant solution since it already provides quite enough swap space without compromising the rest of the RAM too much.

However, if you still think that switching to zswap is mandatory, I would at least suggest to set `vm.swappiness` via `sysctl` to 10 (e.g. `vm.swappiness=10`) to instruct the kernel to use swap only very rarely.

pvizeli:
To understand what I did, we must know the history and go back many years.

Docker uses a lot tempfs and back in days, the devices had max 1GB memory. At this time, all distro used swapfile like ResinOS, Armbian, or Raspbian to make things working. Also, docker will not work with low memory without swap because of tempfs.

I tried to make it different and used ZRAM to compress part of the memory instead of using a swapfile. I played first with zswap but back in the days with kernel 4.3, it did not work stable on all platforms. But also then, the paging with zswap is much faster and optimized than with zram.

I think today with more memory, using zwsap with swapfile swappiness=0-10 would scale a lot better. Also the issues got mostly with backpressure on low memory which zram make it sometimes even more worse.

agners:
> Docker uses a lot tempfs and back in days, the devices had max 1GB memory. At this time, all distro used swapfile like ResinOS, Armbian, or Raspbian to make things working. Also, docker will not work with low memory without swap because of tempfs.

With this change `tmpfs` will stay in zram.

In general, JIT'ed languages are memory hungry. I can also imagine quite some of that memory is actually largely stale, but I don't have prof for that.

> I actually still think that it would be better to keep on using zram to simply avoid any further regular write operation on the underlying storage disks, which could have quite severe negative effects on e.g. SD cards.

That argument assumes that whatever lands in swap is actively used. Quite often that is actually not the case. Quite possible things are swapped out and never accessed during the whole boot time of a system.

In general, memory management is done using LRU algorithm, things which have been accessed most recently stay in main memory. And things which haven't been used since a long time get swapped out (e.g. allocated memory by a program which is not actively used). Swapping those out to main storage, and freeing up (read and write) page cache, helps certainly performance and, I'd argue, even save write cycles (or allows to combine writes into larger junks, which is also healthier for flash based storage).

Imagine memory get scarce. Almost all memory is used. The system is running and is writing logs and storing data in database files. Because the page cache can't grow (all memory is used...) writes are flushed to main storage in small chunks (100KiB at a time). 

If swap is available, that one program which allocated 2MiB of memory and never accesses it, could be swapped out to disk. That would lead to a 2MiB one time write. Then the page cache can grow, and our logs can be written in 2MiB chunks.

Which scenario is better for your SD card?

There are too many variables to exactly tell what is happening.

In the end, flash write cycles are also there to be used. Our system *does* write to storage anyways. If you take 1% of that storage life time, but have better performing system, would you "invest" these write cycles?

There are multiple reports moving away from Home Assistant OS because it crashes (likely due to OOM situations). They start using Debian 11 + Supervised on the same hardware, using swap space on main storage. People seem to be happier with that setup.

I'd say, the risk here is if we make the swap space too big: In that case, things get swapped out which are accessed somewhat frequently. If the memory gets under pressure, this will lead to swap read and writes and significantly contribute to flash storage aging. 

I'd rather prefer to *know* rather than guess... Unfortunately that is quite hard to benchmark :cry: Let's see if I can come up with some data.

> However, if you still think that switching to zswap is mandatory, I would at least suggest to set `vm.swappiness` via `sysctl` to 10 (e.g. `vm.swappiness=10`) to instruct the kernel to use swap only very rarely.

Tuning swappiness is on my todo. My idea was to first find some scenario/test case, test with default settings before starting to turn knobs. We'll see.

agners:
I did some testing, my test setup was as follows:

- Home Assistant OS 10.0.dev builds
- KVM virtual machine using 512MB RAM
- Automation to increment a counter every second (base load I/O writes)
- ESPHome build to create memory pressure

I've compared zram with zswap with different parameters: Factor of memory to zswap file size (0.25, 0.5, 1), swappiness of 60 (default), 30 and 1 (minimal, 0 would turn off swapping).

The question I am trying to answer is: How much is memory pressure affecting write cycles?

For that to answer, I need to have the system under memory pressure. My first attempt was using ESPHome compile: With 512MiB that leads to immediate OOM trigger, since compiling takes up too much memory. I've been using the following test script:

```sh
#!/bin/sh

sysctl -w vm.swappiness=60
date && vmstat -s && iostat
while [ true ]; do
	sleep 300 && date && vmstat -s && iostat
	docker start addon_5c53de3b_esphome
	docker exec -i addon_5c53de3b_esphome esphome compile config/esphome/test.yaml &
done
```

I've only been able to run builds with zram and zswap-0.25. Higher zswap file size lead to compile success, and therefor not comparable IO-Writes.

The IO-Writes by the compile attempts dwarft any difference. The graphs look very boring (values in kBytes):

image

It still is interesting graph, since it shows that other IO effects can be so much bigger.


The second attempt was using `tail /dev/zero`, which causes a simple OOM without doing any IO on its own.

```sh
#!/bin/sh

sysctl -w vm.swappiness=1
date && cat /proc/vmstat && iostat
sleep 30
date && cat /proc/vmstat && iostat
while [ true ]; do
        sleep 30 && date && cat /proc/vmstat && iostat
        tail /dev/zero
done
```

The first three data points are without memory pressure. Then data points under memory pressure follow. Values in kBytes.

image

During the idle phase the values do not differ much. As soon as memory pressure comes in, things change. Especially when there is lots of zswap available, the system starts to swap out each time memory gets scarce. `vm.swappiness` dampens the curve a bit, but not by much.

Comparing idle system shows that zswap typically doesn't lead to a big IO difference.

image


When I look at an 

My conclusion:
- A "write saving" effect of zswap wasn't really visible.
- There is some variability in the system which cause minor differences between measurement runs (sqlite database grows over time etc.). 
- zswap helps extending memory, ESPHome is actually working on a 512MB system.
- Using zswap doesn't affect IO writes by much on a typical HAOS system not not under memory pressure.
- If not much zswap is available, the only idle memory is swapped out. The zswap-0.25 are essentially straight: With that I draw the conclusion that about 512MB/4 is memory which is idle in any case, and can be swapped out without causing additional writes over time.
- If more zswap is available, swapping effects under memory pressure start to add up and cause disk writes.

My setup is quite artificial. I think `tail /dev/zero` is not a good representation: If the system would have additional add-ons running, they would also have a part of largely idle memory, which would go into zswap space for a long period of time.

In my opinion, the trade offs are worth it: IO writes by database/journalctl are in must setups much higher than what zswap will contribute. In my opinion, we can go up to 50% zswap size, but for now I'll go with a conservative 33%. Let's see how this behaves during RC phase, and take it from there.

pvizeli (approved):

prj:
Unfortunately this PR essentially breaks HA on RPI 3B+.

For RPI 3B+ users the only options are either staying on 9.5 or through a ssh workaround hack adding the zram back in on start (since zram0 is still there, but who knows for how long), preferably with a 1G size, as described here:
https://community.home-assistant.io/t/how-to-update-zram-swap-partition-size-on-pi/253032

zswap is unusably slow on an SD card. I don't get why both can not be used, with the zram having higher priority and zswap having lower priority. Probably planned obsolescence.

agners:
> Unfortunately this PR essentially breaks HA on RPI 3B+. A little time after boot you're sitting in swap and everything grinds to a halt.

Hm, interesting. In my testing, the opposite was the case:

- HAOS 9.5 zram was too small, it got slow and ended up in out of memory situation
- HAOS 10 with zswap worked wonders: Lots of "stale" memory got swapped out. The system was snappy and working well.

> zswap is unusably slow on an SD card. I don't get why both can not be used, with the zram having higher priority and zswap having lower priority. 

In general, a SD card with a somewhat decent speed definitely make sense with Home Assistant OS. Otherwise things such as add-on updates or looking up history data is absolutely unusable.

The size of zswap is tuned such that much of the memory swapped out to zswap is not actively used in typical use cases.

> Probably planned obsolescence.

Can you elaborate? We do not provide Raspberry Pi 3, so how does that sentence even make sense?

> Minimum requirement for HA goes from 1GB to 2GB ram with this change for no reason whatsoever.
You already have stuff like this:

Again, I see the oposite here. With HAOS 10 I can run start the system successfully on a system with 512MB RAM. With HAOS 9.5 I can't.

>  zstd is also better than lz4

Define "better"...

> the difference in speed is not too bad, the difference in size is huge.

LZ4 is roughly 4 times faster in decompression, that is quite massive IMHO.

Often swapped out memory compresses quite well (as in, there are zerod out arrays etc.). LZ4 performs pretty well with that, and is just faster compared to zstd.

> This more than doubles the available memory on the device. Hitting swap on a SD-Card is a big no. Both due to wear leveling and due to performance reasons.

Everything is a tradeoff: Having more memory available (thanks to swap on SD) allows you to make more use of the system. But yes, it will cost some write cycles.

But as my testing shows above, in typical use cases it is not that bad. Home Assistant's database writes state changes to disk constantly anyway, that is required to fulfill it's purpose.

Ubuntu uses zswap as well on Raspberry Pi type of devices (see [their blog post](https://ubuntu.com/blog/how-low-can-you-go-running-ubuntu-desktop-on-a-2gb-raspberry-pi-4)). 

So far experience has been quite good with zswap, I don't see a convincing reason in your write-up to move back.

PR #2434, opened 2023-03-28
Improve IPv6 NDP for non-local Thread Border routers

agners:
This change improves IPv6 Neigbor Discover Protocol (NDP) handling for non-local/third-party Thread Border routers. Specifically, with this change multiple routes to the same network are tracked, and router reachability probing is being enabled.

See also: https://github.com/home-assistant/operating-system/discussions/2333#discussioncomment-5446082


Jc2k (approved):

djandrew2005:
Devices continue to experience periodic disconnections

agners:
More likely this is not a routing issue but an mDNS issue. Ideally you try to continusly ping the device to see if it is indeed a routing issue.

This is not the place to discuss disconnects in a case by case base. Please use the discuss section at:
https://github.com/home-assistant/operating-system/discussions/

Issue #2438, opened 2023-03-29
Various Out of memory issues in HAOS 10.0.rc1/10.0.rc2


agners:
It seems that Home Assistant OS 10.0 behaves differently when it comes to memory allocation: Where previously larger allocation were allowed, that seems no longer to be the case. Noteable reports are:

https://github.com/hassio-addons/addon-node-red/issues/1562
https://github.com/hassio-addons/addon-nut/issues/289 (also https://github.com/home-assistant/operating-system/issues/2433)
https://github.com/hassio-addons/addon-emqx/issues/30

The first two instances have been resolved meanwhile, but the third is still pending.

Also, it is currently unclear what is causing the problem exactly. This issue is meant to track progress on investigation.

agners:
Further isolating differences between HAOS 9.5 and HAOS 10.0.rc2 found that docker.service was running with a `LimitNOFILE` limit of `1048576`, and now it is set to `infinity`. The change came with the update from Docker 20.10 to 23.0

- [contrib/init/systemd/docker.service in v20.10.22](https://github.com/moby/moby/blob/v20.10.22/contrib/init/systemd/docker.service)
- [contrib/init/systemd/docker.service in v23.0.1](https://github.com/moby/moby/blob/v23.0.1/contrib/init/systemd/docker.service)

This higher limits lead to different `ulimit -n` values, which these add-on seem to read and use to allocate memory.

The change was introduced a while ago, but only made it to a released Docker version now (see https://github.com/moby/moby/commit/8db61095a3d0bcb0733580734ba5d54bc27a614d). The motivation for the change was performance/overhead which setting this limit seems to have, from the commit log:

> There is a not-insignificant performance overhead for all containers (if
containerd is a child of Docker, which is the current setup) if systemd
sets rlimits on the main Docker daemon process (because the limits
propogate to all children).

agners:
We've decided to stick with the new behavior, see also https://github.com/home-assistant/operating-system/pull/2440#issuecomment-1490184260.

All three known problems have been resolved meanwhile. So I am consider this case closed for now.

PR #2440, opened 2023-03-29
Set file descriptors limit of Docker to pre-OS 10 default

agners:
With the bump to Docker 23.0 a new maximum number of file descriptors got set (LimitNOFILE=infinity in the Docker service files). This reintroduces the old default.

Related to: #2438


agners:
We've decided to try and stick with the new default, especially since the reason of the new default is potential performance advantage (see https://github.com/moby/moby/commit/8db61095a3d0bcb0733580734ba5d54bc27a614d).

I've made a PR dev build with this change. With that, if an add-on reports memory allocation issues, we have a build we can check against to see if the old default would help in that add-ons case. To test the old default with a (almost OS 10.x like) build, use

```
ha su options --channel dev
ha su reload

```

PR #2507, opened 2023-04-24
Avoid CLI restart on shutdown

agners:
This avoids continous CLI restarts on shutdown. The chosen solution is not particularly pretty. Better would be we could inform the OS (systemd) that a shutdown is indeed in progress and act accordingly. But this would need bigger changes which is not in scope for now.

### DIFF
--- a/buildroot-external/rootfs-overlay/usr/sbin/hassos-cli
+++ b/buildroot-external/rootfs-overlay/usr/sbin/hassos-cli
@@ -27,7 +27,9 @@ case $? in
   143)
     # 143 graceful termination (SIGTERM). Most likely a proper shutdown.
     # Just sleep for a while until actual systemd shutdown gets invoked.
-    sleep 15
+    echo ""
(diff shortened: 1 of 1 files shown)


buildroot-external/rootfs-overlay/usr/sbin/hassos-cli
```diff
@@ -19,7 +19,17 @@ docker container exec \
...
+case $? in
+  10)
+    # Jump to root login shell (login command)
     exec /bin/ash -l
-fi
+    ;;
+  143)
+    # 143 graceful termination (SIGTERM). Most likely a proper shutdown.
+    # Just sleep for a while until actual systemd shutdown gets invoked.
+    sleep 15
```
agners, line 30: Ideally we should write something, otherwise the CLI just looks "dead" during this sleep.

```suggestion
    echo ""
    echo "Home Assistant CLI has been terminated."
    sleep 15
```

frenck (approved):

PR #2544, opened 2023-05-11
Make it clearer that the image file needs to be renamed

c0ffeeca7:
- add information that the update was only successful if version 20230328 is visible.


Documentation/boards/hardkernel/odroid-m1.md
```diff
@@ -15,8 +15,11 @@ from [linuxfactory.or.kr][1] as follows:
 1. Download `spiupdate_odroidm1_20220304.img.xz`
 2. Use balenaEtcher or another tool to flash the updater onto an SD card
 3. Download `spiboot-20230328.img`
-4. Store the file with the file name `spiboot.img` onto the FAT partition of that same SD card.
-5. Plug-in that SD card to your ODROID-M1. Petitboot will update itself, you can verify the progress on the HDMI output.
+4. Important! Rename the `spiboot-20230328.img` file to`spiboot.img`.
```
agners, line 18:
```suggestion
4. Rename the `spiboot-20230328.img` file to`spiboot.img`.
```

PR #2575, opened 2023-06-05
Remove rng-tools/rngd since it is not required with kernel 5.15+

jens-maus:
This PR removes `BR2_PACKAGE_RNG_TOOLS` for all supported platforms since kernel 5.15+ the `rngd` (i.e. `rng-tools`) is not required anymore. In newer kernels (5.6+) `/dev/random` is not blocking anymore and the standard kernel provided entropy pool is strong enough since kernel 5.15+ so that tools like `rng-tools` or `haveged` are not required anymore and just unnecessarily consume CPU time. (cf. https://github.com/nhorman/rng-tools/issues/195#issuecomment-1520598021, https://forum.manjaro.org/t/low-entropy-on-my-system/119233).


agners:
I actually suspected that this isn't necessary any more, never came around to actively investigate! Thanks for tackling this along with the references why it isn't necessary anymore :+1: :star_struck:

agners (approved):

Issue #2535, opened 2023-05-04
Bluetooth failing after upgrade to HAOS 10 on Intel Bluetooth hardware


agners:
> Can you check the kernel log, to see if the problem starts similarly, e.g. using:
> 
> ```
> ha host logs -t kernel -n 1000000 -b 0 | grep Bluetooth | head -n 100
> ```

@agners This is the result of the command
![Captura de pantalla 2023-05-03 220333](https://user-images.githubusercontent.com/81590855/236035334-8d67f0b8-ba96-4ac2-b673-6d08ea9a96ea.jpg)

I don't know if this is related but In Home Assistant System>Logs> Host is full of messages like this:

May 03 20:09:04 homeassistant systemd[1]: run-docker-runtime\x2drunc-moby-799946efbb48db91e285d1e93f36945c7175e095d2a5a9fe838fde71f8e6ffdf-runc.jJmMvJ.mount: Deactivated successfully.
May 03 20:09:20 homeassistant systemd[1]: run-docker-runtime\x2drunc-moby-7aee883888a9e7d1b71b97ddbc2b231c2763cc47b14b2c728d6d0de7ab1c2e00-runc.RpO30d.mount: Deactivated successfully.
May 03 20:09:34 homeassistant systemd[1]: run-docker-runtime\x2drunc-moby-799946efbb48db91e285d1e93f36945c7175e095d2a5a9fe838fde71f8e6ffdf-runc.sMXYc7.mount: Deactivated successfully.
May 03 20:09:50 homeassistant systemd[1]: run-docker-runtime\x2drunc-moby-7aee883888a9e7d1b71b97ddbc2b231c2763cc47b14b2c728d6d0de7ab1c2e00-runc.QivMYX.mount: Deactivated successfully.
May 03 20:10:04 homeassistant systemd[1]: run-docker-runtime\x2drunc-moby-799946efbb48db91e285d1e93f36945c7175e095d2a5a9fe838fde71f8e6ffdf-runc.ypP5rP.mount: Deactivated successfully.
May 03 20:10:20 homeassistant systemd[1]: run-docker-runtime\x2drunc-moby-7aee883888a9e7d1b71b97ddbc2b231c2763cc47b14b2c728d6d0de7ab1c2e00-runc.r9dWM5.mount: Deactivated successfully.

_Originally posted by @SrBurner in https://github.com/home-assistant/operating-system/issues/2485#issuecomment-1533660497_

samcrang:
Before:

[...]

PR #2598, opened 2023-06-15
Add patch to fix Bluetooth LE advertisement stall (#2535)

agners:
On some platforms (it seems to be pronounced on Intel NUC systems) Bluetooth advertisements suddenly stop after a short while. Currently there are work arounds in place to restart the HCI controller to keep receiving the advertisements.

Advertisements have been received fine with Linux 5.15. This change reverts a commit which has been isolated to be the culprit.

Since the commit touches generic Bluetooth code used by all controllers, let's revert the commit for all platforms.


agners:
Thanks @sairon for finding the actual offending commit!

PR #2615, opened 2023-06-26
Add config options to enable 32-bit ARMv7 instruction support on ODROID-M1

jens-maus:
This refs https://github.com/jens-maus/RaspberryMatic/issues/2349


jens-maus:
@agners can you please initiate a dev build for this PR so that I can refer this to testers. thx.

PR #2623, opened 2023-07-01
Enable UHID to support Bluetooth input devices

ANDuff:
Enabling UHID config for ova to enable Bluetooth keyboard input to be passed to HA on any ova based install


agners:
Hm, from my perspective we can make this a global change (e.g. in `buildroot-external/kernel/hassos.config`).

PR #2628, opened 2023-07-04
Enable watchdog control in systemd

sbyx:
This is a follow-up to #2627 to enable the systemd configuration for dealing with watchdogs. As mentioned in the comments of the last PR this should also enable watchdog handling on RPi where the watchdog driver is already enabled.

### DIFF
--- a/buildroot-external/rootfs-overlay/usr/lib/systemd/system.conf.d/watchdog.conf
+++ b/buildroot-external/rootfs-overlay/usr/lib/systemd/system.conf.d/watchdog.conf
@@ -1,0 +1,3 @@
+[Manager]
+RuntimeWatchdogSec=default
+WatchdogDevice=/dev/watchdog


jens-maus:
LGTM. However, as the comment in the `system.conf` states:

```
... or by creating "drop-ins" in the system.conf.d/ subdirectory. The latter is generally recommended.
```

perhaps it would be a better approach to add such a dropin file to `system.conf.d` subdirectory instead of modifying the default `system.conf`?

sbyx:
Due to the squashfs the file will be read-only in the installation anyway so hmm, but no strong preference here anyway.

agners:
We've previously added a `logind.conf` in `/etc/system` directly, but I agree the drop in directories are the nicer approach. In particular this avoids the config file from getting stale as new configs get added in newer systemd variants.

However, I'd go for the distribution location at `/usr/lib/systemd/system.conf.d/*.conf`

agners:
Also, can you create two separate PRs for the two things? We might need to revert the enable Watchdog in case it causes problems, and in that case it would be better if its separated from the driver enable PR.

sbyx:
I changed it to a snippet in /usr/lib/systemd.

sbyx:
I removed the watchdog driver here and changed the systemd config into a snippet in /usr/lib/systemd, please check again.

agners (approved):

garethhowell:
I know this was done a while back, but what was the reason for not setting `RuntimeWatchdogUSec=infinity`. This disables the watchdog in normal running conditions: i.e. the time when it's most useful (IMHO)
```
# systemctl show | grep -i watchdog
WatchdogDevice=/dev/watchdog0
WatchdogLastPingTimestamp=Sat 2024-01-20 15:31:54 UTC
WatchdogLastPingTimestampMonotonic=1510421697
RuntimeWatchdogUSec=infinity
RuntimeWatchdogPreUSec=0
RebootWatchdogUSec=10min
KExecWatchdogUSec=0
ServiceWatchdogs=yes
```

agners:
It is set to default, which translates to infinity when reading back: `RuntimeWatchdogSec=default` → `config_parse_watchdog_sec()` at `load-fragment.c:6620` maps the literal string "default" to `USEC_INFINITY` → in `watchdog_update_timeout()` at `watchdog.c:285`, `USEC_INFINITY` means "don't call WDIOC_SETTIMEOUT, just WDIOC_GETTIMEOUT and adopt whatever the driver has".

Which is maybe also not ideal, since this means every hardware will use its own default :cold_sweat: .

PR #2637, opened 2023-07-11
ova: add support for various ethernet adapters

cydia2020:
This pull request proposes the addition of kernel configurations for various PCI-e Ethernet adapters to Home Assistant's virtual machine disk image (ova). The aim is to allow the seamless and painless passthrough of various PCIe Ethernet adapters, which is a feature available on multiple VM platforms. This allows users like me to use their motherboards slower on-board ethernet adapters for less demanding tasks like facilitating Home Assistant's connectivity needs, and avoiding the need for additional configurations and/or hardware.


agners:
We should make this reusable and support the same set of PCIe devices in the ova and generic-x86-64 image. Can you create a `buildroot-external/kernel/device-support-pcie.config` and move the added devices  in there? Remove any of those drivers from `buildroot-external/board/pc/generic-x86-64/kernel.config` and add `buildroot-external/kernel/device-support-pcie.config`  to the config symbol `BR2_LINUX_KERNEL_CONFIG_FRAGMENT_FILES` in `buildroot-external/configs/ova_defconfig` and `buildroot-external/configs/generic_x86_64_defconfig`.

PR #2659, opened 2023-07-30
Including the RTW8821ce driver module to support Wifi on the AK1 PRO micro PC

dfreimuth:
I was looking for a low-cost PC that could run Home Assistant. I looked at the Home Assistant Yellow, but it looks like the CM4 modules are non-existent. So, I started looking at other options. I found this AK1 PRO for ~$160 USD, this looked like a good Intel Celeron N5105 option 4 core 12GB ram. I think that this would be close to the same cost as buying a Yellow and adding a CM4 module, so I decided to try it. The first thing I noticed is that I had no Wifi. I did not realize that it uses Realtek radios. after some research, I found that it has RTW8821ce. The driver is part of the Linux kernel but not enabled for HAOS. This request is to enable the RTW8821ce module to support the AK1 Pro. This AK1 PRO seems to be sold under a few different company names, but the one that I am focusing on is here https://www.amazon.com/gp/product/B099KSJN8P


dfreimuth:
One add'l note, for reference and ease here is the link to the kernel config info for the 8821CE.

https://www.kernelconfig.io/CONFIG_RTW88_8821CE?kernelversion=6.1.38

agners (approved):

PR #2671, opened 2023-08-06
Delete Dockerfile

yash-naikk:


agners:
Using a Docker image is the recommended way to build HAOS, so this Dockerfile is required.

PR #2691, opened 2023-08-17
Add grub.cfg operand for headless x86_64 devices

Martes-Wigglesworth:
- updated grub.cfg with nomodeset for headless boot issue resolution.

### DIFF
--- a/buildroot-external/board/pc/grub.cfg
+++ b/buildroot-external/board/pc/grub.cfg
@@ -60,7 +60,7 @@ fi
 
 save_env A_TRY A_OK B_TRY B_OK ORDER MACHINE_ID
 
-default_cmdline="rootwait zram.enabled=1 zram.num_devices=3 net.naming-scheme=v250 systemd.machine_id=$MACHINE_ID fsck.repair=yes $boot_condition"
+default_cmdline="rootwait zram.enabled=1 zram.num_devices=3 net.naming-scheme=v250 systemd.machine_id=$MACHINE_ID fsck.repair=yes $boot_condition nomodeset"
 file_env -f ($root)/cmdline.txt cmdline
 
 # root is a full HDD/partition definition in GRUB format like hd0,gpt1


Martes-Wigglesworth:
I am still extremely new to Home Assistant so I hope I found the correct update workflow.

I am attempting to add the line as follows, to the resultant /etc/default/grub.cfg

`GRUB_CMDLINE_LINUX_DEFAULT="nomodeset"`

agners:
> I am still extremely new to Home Assistant so I hope I found the correct update workflow.

That is fine, welcome!

> 
> I am attempting to add the line as follows, to the resultant /etc/default/grub.cfg
> 

Yes, your changes indeed add `nomodeset` to the default kernel command line :+1: 


While I think it should be generally fine to disable kernel mode setting, it can have some negative consequences: Older UEFI bioses might initialize with every low resolution, so these systems will show a console with very low resolution. 

A bit more worrying could be buggy BIOSes might don't correctly hand over a framebuffer. On such systems using `nomodeset` might actually be negative.

What exact system do you experience problems with kernel mode setting?

Martes-Wigglesworth:
> > I am still extremely new to Home Assistant so I hope I found the correct update workflow.
> 
> That is fine, welcome!
> 
> > I am attempting to add the line as follows, to the resultant /etc/default/grub.cfg
> 
> Yes, your changes indeed add `nomodeset` to the default kernel command line 👍
> 
> While I think it should be generally fine to disable kernel mode setting, it can have some negative consequences: Older UEFI bioses might initialize with every low resolution, so these systems will show a console with very low resolution.
> 
> A bit more worrying could be buggy BIOSes might don't correctly hand over a framebuffer. On such systems using `nomodeset` might actually be negative.
> 
> What exact system do you experience problems with kernel mode setting?

I came across this issue with my HP EliteDesk 600 G3, after running it for about a month and then deciding to place it into my house's rack as final location.  So this is one of those "old, but not ancient" situations.

(i.e. "Its old enough not to require me to send it back to HP to pay them to replace the motherboard if I forgot the admin password on the bios, but it is new enough to have UEFI boot and core i5")

Martes-Wigglesworth:
@agners I have done some further research on possible behaviors and it appears that this setting is described as reverting back to "Bios" settings for initial booting, and then after successful boot then the parameters will be rediscovered based on drivers, such as the NVidia drivers, etc...  I do not see anything that would imply a failure state when utilizing "nomodeset" however, I have really never done any kernel development so I will have to defer to your input for further negative use-cases.  (It seems to be a failsafe when a linux boot fails due to failures of graphical detections on certain systems.)

agners:
The bootloader GRUB2 reads additional platform specific kernel command line arguments from `/mnt/boot/cmdline.txt`. You can edit that file, it will not get overwritten by the upgrade process. However, currently there is no easy way to edit that file other than editing from a live distribution or use the OS SSH access on port 22222 (see https://developers.home-assistant.io/docs/operating-system/debugging#ssh-access-to-the-host).

Can you try if that works for you?

We probably should consider a command line option  under the `ha os`  command to edit that file. This would at least allow to  solve cases like this.

Martes-Wigglesworth:
> The bootloader GRUB2 reads additional platform specific kernel command line arguments from `/mnt/boot/cmdline.txt`. You can edit that file, it will not get overwritten by the upgrade process. However, currently there is no easy way to edit that file other than editing from a live distribution or use the OS SSH access on port 22222 (see https://developers.home-assistant.io/docs/operating-system/debugging#ssh-access-to-the-host).
> 
> Can you try if that works for you?
> 
> We probably should consider a command line option under the `ha os` command to edit that file. This would at least allow to solve cases like this.

I was looking at editing that file before but figured that a PR would make sense if others are not aware.  I will test this out using the explicit grub operand and see if I can get it to work.  I tried to edit the other files under /mnt/boot and they kept getting ignored, earlier.  So I figured that something else was being read which I was unaware of so I stopped just prior to updating cmdline.txt, and opted for the PR and this dialog for it.

Martes-Wigglesworth:
@agners I am retesting now and can verify that I see no updated boot option operands after updating cmdline.txt with the above-listed GRUB operand and value set to “nomodeset”. 

I also see that this system does not boot without the monitor after these attempted updates, which would make some sort of sense if the grub2 configuration is ignoring my updates to cmdline.txt

Martes-Wigglesworth:
> > @agners I am retesting now and can verify that I see no updated boot option operands after updating cmdline.txt with the above-listed GRUB operand and value set to “nomodeset”.
> 
> The text file is directly loaded, so no GRUB specific operands are necessary. Also make sure to have everything in a single line. This works here:
> 
> ```
> # cat /proc/cmdline 
> BOOT_IMAGE=(hd0,gpt4)/bzImage root=PARTUUID=a3ec664e-32ce-4665-95ea-7ae90ce9aa20 rootwait zram.enabled=1 zram.num_devices=3 net.naming-scheme=v250 systemd.machine_id=2403f24a7b9840c0a6c63ae9357948e0 fsck.repair=yes console=tty1 nomodeset rauc.slot=B
> # cat /mnt/boot/cmdline.txt 
> console=tty1 nomodeset
> ```

Greetings @agners 

As I updated in the original dialog for the issue, for me, this HP EliteDesk does not seem to support headless behavior without a physical dummy VGA device.  

I realized that the configurations were being picked up because all of a sudden I realized I was looking at large font from BSD days, instead of the flash and then smaller legitimate font for boot messaging.  The issue was that this device's BIOS seems to have the behavior associated with, "no POST without monitor" configuration, which some older thin-clients seem to have.

Hence, for my situation, I believe an update to documentation may be a good fit for resolution.  

However, there seem to be other instances where certain headless supported bios implementations do not boot headless utilizing HAOS, however, I don't have any other items which will be testable for that use-case.  (I ended up installing the image onto a mini pc that was purposely purchased for this purpose, a while back.  I did some further testing and discovered that the bios on this version of HP EliteDesk is setup to not post properly if a VGA input is not detected.  

I would like to get something out of this effort, since others seem to have been equally confused by the lack of headless booting, however, for this instance, I believe that there are simply bios configurations which are setup to only allow for assumed "desktop" use of their device, and they do this to prevent using the device as a headless system, for whatever reason.

Do you have any further recommendations for how to proceed?  I don't see anything about this topic in the docs, however, I may have missed it.

agners:
> Hence, for my situation, I believe an update to documentation may be a good fit for resolution.

From what I understand there was no real resolution for your use case correct? What would you like to document? Just the fact that there is problem without monitor for your model?

> Do you have any further recommendations for how to proceed? I don't see anything about this topic in the docs, however, I may have missed it.

There are quite some corner cases which would benefit from documentation. Unfortunately, adding hints about various configurations also clutters the regular documentation quite a bit, which confuses people and isn't helpful for the general flow.

Ideally we probably would need some troubleshooting section for the generic-x86-64 installation guide and collect such hints there.

Currently editing cmdline.txt is also harder than necessary. Ideally I'd like to add a command to the `ha` CLI command to edit commandline options.

agners:
I am closing this PR as the solution to the problem is to use a custom kernel command line (per-instance).

PR #2714, opened 2023-08-30
Use proper fix for Bluetooth LE advertisement stall (#2598)

agners:
Instead of reverting the new hci_sync based BLE scan disable logic use the fix proposed by Luiz:
https://lore.kernel.org/linux-bluetooth/CABBYNZ+5RMqNVMyYKi+gOVaV+K6k8Z-C37KnfGa=qRUORc3dWg@mail.gmail.com/

This fix avoids BLE stalls just like the revert.


fdi8686:
Hi agners,
I applied the changes of Luiz to the source code, compiled the kernel but the issue is still there. "uname -r" shows my modified kernel so I should be running it. Is there any other way to check I applied the patch correctly? (not really expert).
Thanks

PR #2751, opened 2023-09-14
Delete Bluetooth device cache regularly

agners:
Delete stale Bluetooth devices from the BlueZ device cache every week. This makes sure that the overlay partition doesn't run out of inodes which has happened in real world scenarios where many new Bluetooth devices are discovered.

BlueZ maintains these files on a best effort base. So removing them while BlueZ is running should be safe. Their modification timestamp gets updated regularly (on each connect), so deleting cache entries older than 7 days essentially deletes all the entries of devices which haven't been used the last 7 days.

An alternative considered  was to lower BlueZ GATT caching (e.g. by using Cache=yes instead of always, to cache only paired devices). However, this would hurt performance and battery lifetime of Bluetooth devices due to additional unnecessary GATT attributes reads. This is in particular true for Bluetooth 5.1 devices which support the Database Hash charactristic. Caching has also helped reliability with intermittent connections (see
https://github.com/bluez/bluez/issues/191).

More importantly, besides the GATT attribute cache the same files are also used to cache the device names as well. This is independent of the above mentioned GATT cache configuration (see device_store_cached_name in BlueZ). So disabling the GATT caching alone wouldn't solve the particular problem we are facing.

See also: https://github.com/home-assistant/operating-system/issues/2752

### DIFF
--- a/buildroot-external/rootfs-overlay/usr/lib/systemd/system/haos-bt-cache.service
+++ b/buildroot-external/rootfs-overlay/usr/lib/systemd/system/haos-bt-cache.service
@@ -1,0 +1,15 @@
+[Unit]
+Description=Remove Bluetooth cache entries older than one week
+
+[Service]
+Type=oneshot
+ExecStart=/bin/sh -c 'find /var/lib/bluetooth/*/cache -mindepth 1 -type f -mtime +7  -exec rm {} \;'
+PrivateDevices=yes
+PrivateNetwork=yes
+PrivateUsers=no
+ProtectKernelTunables=yes
+ProtectKernelModules=yes
+ProtectControlGroups=yes
+MemoryDenyWriteExecute=yes
+SystemCallFilter=@default @file-system @basic-io @system-service
+
--- a/buildroot-external/rootfs-overlay/usr/lib/systemd/system/haos-bt-cache.service
+++ b/buildroot-external/rootfs-overlay/usr/lib/systemd/system/haos-bt-cache.service
@@ -3,7 +3,7 @@ Description=Remove Bluetooth cache entries older than one week
 
 [Service]
 Type=oneshot
-ExecStart=/bin/sh -c 'find /var/lib/bluetooth/*/cache -mindepth 1 -type f -mtime +7  -exec rm {} \;'
+ExecStart=/bin/sh -c 'find /var/lib/bluetooth/*/cache -mindepth 1 -type f -atime +7  -exec rm {} \;'
 PrivateDevices=yes
 PrivateNetwork=yes
 PrivateUsers=no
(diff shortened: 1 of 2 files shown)


buildroot-external/rootfs-overlay/usr/lib/systemd/system/haos-bt-cache.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=Remove Bluetooth cache entries older than one week
+
+[Service]
+Type=oneshot
+ExecStart=/bin/sh -c 'find /var/lib/bluetooth/*/cache -mindepth 1 -type f -mtime +7  -exec rm {} \;'
```
rexbron, line 6: Does the modifed time get changed evertime that the cached file is used? Would atime be a more accurate way to determine if the cache is stale?

agners: Yes, I validated this in my testing: All devices which are nearby have updated modification timestamps. Should have mentioned that in the original PR.

rexbron: I have 3x BTHome Temperature BLE sensors. As I mentioned above my whole cache seems like it would be cleared. Is there a concern with removing active cached files?

agners: Uh wait, you are saying your installation doesn't update modification timestamps? :thinking:

agners: Yeah that is definitely not what I am seeing here:

```
# find /var/lib/bluetooth/*/cache -mindepth 1 -type f -mtime +7 | wc -l
118
# ls -1 /var/lib/bluetooth/*/cache | wc -l
134
# 
```

However, that being said, modification time could rely on an implementation detail of how the cache is managed currently. If a cache file has been accessed, it might be used, so we should use atime.

Using atime leads to only slightly different numbers here:

```
# find /var/lib/bluetooth/*/cache -mindepth 1 -type f -atime +7 | wc -l
116
# ls -1 /var/lib/bluetooth/*/cache | wc -l
134
```

agners: > I have 3x BTHome Temperature BLE sensors. As I mentioned above my whole cache seems like it would be cleared. Is there a concern with removing active cached files?

It would lead to unnecessarily read of the GATT services/attributes. Slows down that first read and uses a bit more battery, but since that would be only once a week it wouldn't be a big deal.

That said, I think the BTHome devices' GATT service/attributes are not cached currently, at least I don't see any entries in the cache files. 

But in any case, by checking access time we would even prevent that read.

pvizeli (commented):
Why not with tmpfile?

agners:
> Why not with tmpfile?

The cache is meant to survive reboots, from what I understand (it is in /var/lib after all).

PR #2767, opened 2023-09-21
Adjust Home Assistant OS versioning to prepare for new release strategy

agners:
With OS 11 we'll create rc pre-releases which will get directly pushed to the beta channel. In contrast, release builds will get directly pushed to the stable channel.

Similar to Home Assistant Core we'll create bump commits for all stable and beta releases. This makes sure that the source code matches the built binaries for all releases.

The development build will get a generated version. To avoid issues with the new rc builds the dev build version will get injected on source level now.


.github/workflows/build.yaml
```diff
@@ -138,6 +147,13 @@ jobs:
           submodules: true
           persist-credentials: false
 
+      - name: Inject development version
```
agners, line 150:
```suggestion
      - name: Set version suffix
```

.github/workflows/build.yaml
```diff
@@ -32,32 +33,49 @@ jobs:
...
+      - name: Generate development version
+        shell: bash
+        id: version_dev
+        if: ${{ github.event_name != 'release' }}
+        run: |
+          version_dev="dev$(date --utc +'%Y%m%d')"
+          echo "Development version \"${version_dev}\""
+          echo "version_dev=${version_dev}" >> $GITHUB_OUTPUT
+
+      - name: Inject development version
```
agners, line 45:
```suggestion
      - name: Set version suffix
```

PR #2768, opened 2023-09-23
Rockchip-blobs: Add support for OP-TEE blob

darkxst:
Add support to Rockchip-blobs for handling OP-TEE binary blobs.

OPTEE is mandatory for some 32-bit SoC's such as rv1126 and others. Optee OS doesnt yet support any SoC's so its necessary to use the Rockchip vendor blobs.

Example Usage:
```
BR2_PACKAGE_ROCKCHIP_BLOBS=y
BR2_PACKAGE_ROCKCHIP_BLOBS_VERSION="b4558da0860ca48bf1a571dd33ccba580b9abe23"
BR2_PACKAGE_ROCKCHIP_BLOBS_TEE="bin/rv11/rv1126_tee_ta_v2.09.bin"
```

There will be a corresponding buildroot patch to set the TEE build envvar.


agners:
> 32-bit SoC

We rather prefer to phase out 32-bit support, especially on Arm :grimacing:  What is the draw to use the rv1126 SoC? 1.2GHz Cortex-A7 seems rather on the low end these days.

Also, the rockchip-blob package is somewhat a hack. I talked with @citruz, it is also interfering with rk3399 builds. I'd rather prefer to build ATF from sources (which I think for the rk3568 should be possible with patches).

That said, there will probably always be some Rockchip SoC which needs the binary blobs, at least at first. Maybe the incompatibility can be fixed with some more Buildroot/Kconfig changes. Ideally we'd upstream them too.

It is a small PR, so I leaning towards merging it.

jens-maus:
> > 32-bit SoC
> 
> We rather prefer to phase out 32-bit support, especially on Arm 😬

FYI: I fully agree that IMHO the 32-bit support for Arm platforms supporting both (32bit and 64bit) should be phased out. IMHO there is no reason why e.g. we still supply 64bit and 32bit RaspberryPi support since 64bit is nowadays the standard environment for the Pis. Only platforms like e.g. the Tinkerboard which only comes with 32bit SoCs should still be supported.

agners:
> FYI: I fully agree that IMHO the 32-bit support for Arm platforms supporting both (32bit and 64bit) should be phased out. IMHO there is no reason why e.g. we still supply 64bit and 32bit RaspberryPi support since 64bit is nowadays the standard environment for the Pis.

Well, there are two reasons actually:

1. 32-bit requires slightly less memory. We have some folks which migrated an installation and found that on 64-bit they run into out-of-memory situations. I don't think there is a solution to that really, other than using a system with more memory available.
2. There are at least 3500 installations out there according to analytics. Since that is quite some folks, we should at least inform folks through our channels before we stop supplying updates.

Note that also Raspberry Pi OS itself uses 32-bit by default still today :grimacing: 

While I am all for going towards phasing out 32-bit Arm, I think we still have to approach it slowly. We also still have the ASUS Tinkerboard which is a 32-bit platform. The installation base is rather small however (analytics counts 124 instances).

A first step we did a couple of months back was removing the 32-bit images from the installation documentation.

darkxst:
> What is the draw to use the rv1126 SoC?

Only that it is what Sonoff used in the [iHost Hub](https://itead.cc/product/sonoff-ihost-smart-home-hub)

Apart from the somewhat questionable choice of SoC, this is a pretty nice package with an on-board Silabs EFR32MG21 (unlocked bootloader), Wifi/BT, RGB indicator, speaker/mic etc. I've only tested on the quad core@1.5GHz version ( 4GB DDR4 RAM) but seems to handle HA fine. 

Its also mostly supported upstream/mainline in u-boot and kernel (from 6.3).

agners:
> Only that it is what Sonoff used in the [iHost Hub](https://itead.cc/product/sonoff-ihost-smart-home-hub)

Ah I see. Is that system booting off the SD card or how are you booting HAOS on it?

On second thought, I kinda dislike to add these patches and carry them downstream. Ideally we should start upstream the `rockchip-blobs` package and then incorporate your changes as well.

What is your plan with this board?

darkxst:
> Ah I see. Is that system booting off the SD card or how are you booting HAOS on it?

Yeh I am booting off SD card at the moment, ultimately would boot off eMMC, but that is only 8GB so still going to need a datadisk on sd or usb.
 
> On second thought, I kinda dislike to add these patches and carry them downstream. Ideally we should start upstream the `rockchip-blobs` package and then incorporate your changes as well.

Sure that makes sense.

> What is your plan with this board?

If I can get it working completely (currently device tree is incomplete so some peripherals not setup yet) was thinking of proposing merging it, but then the 32-bit thing might be a blocker for that?

agners:
Since we rather move to phasing out Arm 32-bit support we don't want to add this platform to Home Assistant OS.

PR #2772, opened 2023-09-25
generic-x86-64: enable backlight support

craftyguy:
This enables backlight support on these hosts, which is useful if running HASS on an old laptop or tablet and you want to (e.g.) conserve power by controlling the backlight.

I'm not really sure if this is enough to enable backlight control for i915... it's not clear how I can use normal kernel build system tools to validate this kconfig fragment (e.g. to pull in any other module/driver dependencies), and it's not clear to me how I can build/test a kernel with this config on an existing install of HAOS. I looked at kconfig from a system where backlight control was working, as a basis for these changes.


buildroot-external/board/pc/generic-x86-64/kernel.config
```diff
@@ -28,6 +28,7 @@ CONFIG_BCMA=m
 CONFIG_SYSFB=y
 
 CONFIG_FB=y
+CONFIG_FB_BACKLIGHT=m
```
agners, line 31: This is very unlikely to be necessary as it is used only for a couple of fbdev only drivers which we haven't enabled even, so:

```suggestion
```

agners:
> I'm not really sure if this is enough to enable backlight control for i915... it's not clear how I can use normal kernel build system tools to validate this kconfig fragment (e.g. to pull in any other module/driver dependencies), and it's not clear to me how I can build/test a kernel with this config on an existing install of HAOS. I looked at kconfig from a system where backlight control was working, as a basis for these changes.

You probably could use a specific buildroot target to let only config part run and then copy the config file from the Buildroot's build directory. At that point it is probably easier to just build a complete image :sweat_smile: 

From what I can tell, this is likely to do what you intend. I am fine with just merging as is, then you can test the next nightly build to verify.

agners:
@craftyguy downloading the rauc file is possible, but kinda complicated. You can install official dev builds directly by switching to the dev channel like so:

```
ha supervisor options --channel dev
ha supervisor reload
ha supervisor update
ha os info # <-- you should see the update here
ha os update
```

PR #2786, opened 2023-10-03
RAUC update: Use verity bundle format

agners:
Move from the current plain format to the new verity bundle format. This requires at least HAOS 10.4 to work. The Supervisor will make sure to update to the latest minor release of the previous major release, so updating will work in the regular use case.


sairon (approved):
Looks correct, will test once we have a dev build.

PR #2791, opened 2023-10-05
Set containerd root explicitly (#2476)

agners:
With the move to Docker 23 containerd stores its metadata no longer undernath the Docker data directory but at its default location at /var/lib/containerd. Previously Docker passed a containerd configuration toml file which explicitly set the metadata root underneath Docker's data directory.

On Home Assistant OS, the new location /var/lib/containerd is on a tmpfs file system. For unknown reasons, it seems that if containerd's root directory is on a tmpfs this leads to significantly more syscalls and hence CPU load.

Change the metadata location to be on the data partition again. Since containerd is treated separately from Docker these days, use a new root directory under /mnt/data for containerd as well. With this, the CPU load of containerd is back to normal.

### DIFF
--- a/buildroot-external/rootfs-overlay/etc/containerd/config.toml
+++ b/buildroot-external/rootfs-overlay/etc/containerd/config.toml
@@ -1,4 +1,7 @@
 version = 2
+
+root = "/mnt/data/docker/containerd/daemon"
+
 disabled_plugins = [
   "io.containerd.internal.v1.opt",
   "io.containerd.tracing.processor.v1.otlp",


sairon (changes requested):
It'd be better to stay consistent and create a bind mount for the `containerd` folder instead, like we do for [other persistent stuff](https://github.com/home-assistant/operating-system/blob/dev/buildroot-external/rootfs-overlay/usr/lib/systemd/system/var-lib-systemd.mount) in the `/var/lib` (which is historically `tmpfs`, unlike on standard systems).

However, before the change, the `containerd` files were at `/mnt/data/docker/containerd/daemon` and they're hanging there now without any use or cleaning - so it's a point we should consider as well - shouldn't we start using that path (again) instead of using an entirely new one?

agners:
> It'd be better to stay consistent and create a bind mount for the `containerd` folder instead, like we do for [other persistent stuff](https://github.com/home-assistant/operating-system/blob/dev/buildroot-external/rootfs-overlay/usr/lib/systemd/system/var-lib-systemd.mount) in the `/var/lib` (which is historically `tmpfs`, unlike on standard systems).

We are not all that consistent these days: Data root of Docker is explicitly set to `/mnt/data/docker`, while we also bind mount the default `/var/lib/docker` to the same location...

Also, I think we should consider making /var/lib writable by default. Most directories there are bind mounted to a writeable location already. And as you said, in standard systems this is a persistent location. This location not being persistent already lead to surprises in the past. I don't think that /var/lib has lots of writes in practice, and if we have an application which writes a lot there we should investigate/address that on a per application basis. But this is something for later. :sweat_smile: 


> However, before the change, the `containerd` files were at `/mnt/data/docker/containerd/daemon` and they're hanging there now without any use or cleaning - so it's a point we should consider as well - shouldn't we start using that path (again) instead of using an entirely new one?

I'd prefer a location not underneath Docker since containerd is a separate thing nowadays. But with the above change in mind I agree, let's reuse the old location for now :+1:

sairon (approved):
OK, agreed, let's revert to the previous behavior and keep the bigger changes for later.

PR #2792, opened 2023-10-05
Prepare 11.0.rc1

agners:


sairon (approved):
:tada: 

(just a nitpick - let's use a different branch name next time, so it wouldn't collide with the coming version tag)

PR #2812, opened 2023-10-13
Create foundation for Labgrid-based OS tests

sairon:
Add foundation for Labgrid-based tests of OS builds. Currently uses just the QEMU driver, which starts a virtual machine with pristine OS, and generates few log reports which are saved as build artifacts.

Workflow is currently triggered either manually by specifying an OS version, or by OS build job, which now saves an artifact of the OVA image. This allows for some modularity. If we eventually add the possibility to run builds on PRs, we could also add the workflow_call trigger and turn the workflow into a reusable one.

TBD (in future PRs): some meaningful tests and possibility to test on real hardware (either local or distributed).


.github/workflows/test.yaml
```diff
@@ -0,0 +1,92 @@
...
+          python-version: 3.12
+
+      - name: Install Python requirements
+        run:
+          pip install -r requirements.txt
+
+      - name: Download HAOS image
+        if: ${{ github.event_name == 'workflow_dispatch' }}
+        run: |
+          curl -sL https://os-artifacts.home-assistant.io/${{github.event.inputs.version}}/haos_ova-${{github.event.inputs.version}}.qcow2.xz > haos.qcow2.xz
```
agners, line 52:
```suggestion
          curl -sfL -o haos.qcow2.xz  https://os-artifacts.home-assistant.io/${{github.event.inputs.version}}/haos_ova-${{github.event.inputs.version}}.qcow2.xz
```

.github/workflows/test.yaml
```diff
@@ -0,0 +1,92 @@
...
+        if: ${{ github.event_name == 'workflow_run' }}
+        uses: dawidd6/action-download-artifact@v2
+        with:
+          workflow: build.yaml
+          workflow_conclusion: success
+          name: ova-image
+
+      - name: Extract OS image
+        run: |
+          xz -dc haos*.qcow2.xz > haos.qcow2 && rm haos*.qcow2.xz
```
agners, line 64: Since we use explicit file names above, star is not needed here anymore. Also, `unxz` does what you want:

```suggestion
          unxz haos.qcow2.xz
```

agners (commented):
This looks awesome, looking forward for the first tests to flow in :star_struck: 

You went with the Python default dir `tests/` in the root, I wonder if we should name it a bit differently to make it clear these are Labgrid specific tests :thinking: 

Or maybe add a README.md in that directory (along with a high level description how to run those?)

.github/workflows/test.yaml
sairon: D'oh, I probably had some temporary mental weakness that I forgot about `unxz`.

The asterisk is a relict from previous iterations, thanks for pointing that out.

sairon:
@agners Thanks for the review!

> Or maybe add a README.md in that directory (along with a high level description how to run those?)

I'll go this way and keep it named `tests`, since the name of the framework is just an implementation detail. I will also add a simple shell script that starts the tests and use it in the workflow, so running it locally will be easier and same as in the CI env.

PR #2813, opened 2023-10-13
generic-x86-64: Add missing dependency for #2800 to kernel.config

xtsrthm:
Add the CONFIG_IIO=m config option, which is a dependency for CONFIG_DLN2_ADC=m and CONFIG_BMP280=m, added in #2800.


sairon (approved):
Sorry it took us so long, I just changed the order in the config a bit, otherwise it's fine. Thanks for the PR!

PR #2830, opened 2023-10-17
RaspberryPi: Update kernel 6.1.54 - stable_20231004

sairon:


jens-maus:
I have also noticed that the RaspberryPi Linux GitHub repo contains this new `stable_20231004` tag (https://github.com/raspberrypi/linux/tags) and thought they might have changed their regular stable tagging scheme once more. What makes me wonder if this is really true is, however, that they don't use the same tag in the raspberrypi/firmware github repo (https://github.com/raspberrypi/firmware/tags) and still use the old `1.YYYYMMDD` tagging there.

Unfortunately I haven't found an official documentation or comment somewhere what regular stable tagging scheme they are using or if they don't do this regular. Seems to be not well defined, I am afraid.

PR #2833, opened 2023-10-17
Do not hardcode OS deployment type

sairon:
Hardcoding the deployment requires changing it back and forth when merging between the branches. Instead of doing that, fall back to the development mode and determine the version along with the channel in the CI preparation step. Deployment can be overridden anytime locally by adding the DEPLOYMENT parameter to the make command, e.g.:

`make DEPLOYMENT=production ova`

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -28,6 +28,7 @@ jobs:
       version_main: ${{ steps.version.outputs.version_main }}
       version_full: ${{ steps.version.outputs.version_full }}
       channel: ${{ steps.channel.outputs.channel }}
+      deployment: ${{ steps.channel.outputs.deployment }}
       matrix: ${{ steps.generate_matrix.outputs.result }}
       build_container_image: ghcr.io/${{ github.repository_owner }}/haos-builder@${{ steps.build_haos_builder.outputs.digest }}
     steps:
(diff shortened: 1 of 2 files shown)


agners:
Just like the version number, this is a very philosophical question: Should the build information be part of the source or should it be passed in from the build system.

> Hardcoding the deployment requires changing it back and forth when merging between the branches. 

Since we anyways bump the version number on source level in main and rc branch (even in the same file), this will lead to a commit anyways, so this is not really a problem per-se.

That said, even with this patch merged, we could still set the default in the main and rc branch to keep all build information part of the source and allow for reproducible builds.

However, I recently opted the other direction for the dev/version suffix (see https://github.com/home-assistant/operating-system/pull/2767/files). Instead of making `buildroot-external/meta` adjustable, I went the road to modify things on source level. I went that road because I felt `buildroot-external/meta`  should be a pure key/value file, which not necessarily need to get sourced.... Again, this is kinda philosophical question. But IMHO, we should use the same style for version suffix as for development.

sairon:
Understood, thanks for the insights. But IMHO it's still valid requirement to make the `DEPLOYMENT` configurable, because currently the only option you have is to patch the `meta` file if you want to make a local build of the release branch.

In that case, I think we could add some more logic to `buildroot-external/scripts/rauc.sh` which would allow for this. Or do you have a better idea?

agners:
> In that case, I think we could add some more logic to `buildroot-external/scripts/rauc.sh` which would allow for this. Or do you have a better idea?

Yes I was actually thinking that too. Technically, deployment and the signing key used are not really related. It's just that we have separate keys for those two deployment types for security reason.

So if one wants to build a image which pretends to be a production build, and sign it with a dev key, that is perfectly fine. It just won't get accepted by a image which is a regular production image (with the production keychain only)...

sairon:
I will open another PR with the alternative solution then.

PR #2842, opened 2023-10-19
Fix extraction of OVA image artifact in test step

sairon:
If the test image is obtained from an artifact instead of downloading, its name contains the version as well, in that case we still need to use wildcard expansion.


agners (changes requested):
We also use `disk-image: ./haos.qcow2` in the `tests/qemu-strategy.yaml` file. So I guess we should rather rename in the image artifact case.

agners (approved):

PR #2854, opened 2023-10-23
Add udev rule to rpi-rf-mod package 

jens-maus:
This PR adds an udev rule to set write permissions for the led device nodes for the RPI-RF-MOD hardware (HomeMatic RF module) so that newer versions of addons (e.g. raspberrymatic) using dedicated users to access these leds nodes can still modify these leds during runtime.

### DIFF
--- a/buildroot-external/package/rpi-rf-mod/82-rpi-rf-mod-leds.rules
+++ b/buildroot-external/package/rpi-rf-mod/82-rpi-rf-mod-leds.rules
@@ -1,0 +1,4 @@
+# make sure all rpi-rf-mod led nodes in /sys are generated with permissions so that addons
+# can access it (e.g. raspberrymatic)
+SUBSYSTEM=="leds", ACTION=="add", KERNEL=="rpi_rf_mod:*", RUN+="/bin/chmod -R g=u,o=u /sys%p"
+SUBSYSTEM=="leds", ACTION=="change", KERNEL=="rpi_rf_mod:*", ENV{TRIGGER}!="none", RUN+="/bin/chmod -R g=u,o=u /sys%p"
--- a/buildroot-external/package/rpi-rf-mod/rpi-rf-mod.mk
+++ b/buildroot-external/package/rpi-rf-mod/rpi-rf-mod.mk
@@ -57,6 +57,7 @@ define RPI_RF_MOD_INSTALL_TARGET_CMDS
 	if [[ -n "$(RPI_RF_MOD_DTS_FILE_ALT)" ]]; then \
 		$(INSTALL) -D -m 0644 $(@D)/buildroot-external/package/rpi-rf-mod/dts/$(RPI_RF_MOD_DTS_FILE_ALT).dtbo $(BINARIES_DIR)/; \
 	fi
+	$(INSTALL) -D -m 644 $(RPI_RF_MOD_PKGDIR)/82-rpi-rf-mod-leds.rules $(TARGET_DIR)/lib/udev/rules.d/
 endef
 
 $(eval $(generic-package))


agners:
Sorry for the late review.

So first, off, there is a related request in Supervisor, see https://github.com/home-assistant/supervisor/issues/4560. With debated a bit how to go about LEDs in general in Home Assistant, and came to the conclusion to open up direct access (see https://github.com/home-assistant/supervisor/issues/4560).

Now this made me wonder: Can you even write to that path in first place?

It turns out yes, and it requires three things to be true `SYS_ADMIN` need to be set to get the [mount capability](https://man7.org/linux/man-pages/man7/capabilities.7.html) inside the container in first place, so in add-on `config.yaml`:

```
privileged:
  - SYS_ADMIN
```

Second, a custom apparmor profile is needed, as the default prevents mount (see https://github.com/moby/moby/blob/v24.0.7/profiles/apparmor/template.go#L42).

And third, the init system needs to remount /sys with rw, something along these lines:
```
/bin/mount -o remount,rw sysfs /sys
```

And it seems that your init system is doing this on startup. This is not really all that relevant for you, but helps giving a bit of context.

Now all of the above gives you write access as root to that location in first place. But because your init system then starts processes as user, you need to have user writable LEDs.

That is where this new udev rule comes in.

Now on typical system that is the way to go about this.

But it feels a bit weird to me that we adjust permissions of device files in `/sys` for the needs of a particular add-on. All LED attribute are by default read-only `0644`.

So far we don't have a good way to support add-on running with user privileges in general. Even less so when it's about HW access :cry: 

All that said, it seems a lot of fuzz about a couple of lines in HAOS :sweat_smile:  and we have already custom HW support, so :man_shrugging:  However, before simply merging this, maybe there is another way to address this? :thinking:

jens-maus:
> Now all of the above gives you write access (at least as root) to that location in first place. But because your init system then starts processes as user, you need to have user writable LEDs.
> 
> That is where this new udev rule comes in.

That's absolutely correct. However, please allow me to provide some background info on that. In the past (and actually also currently) the process that steers this special LED of the `RPI-RF-MOD` Homematic RF-module is/was running in the `root` user context and thus it can/could access the correspondings `/sys/class/leds/..../trigger` files and write out the new LED mode without any issue.

However, to increase the security to some extend I actually introduced dedicated users for some of the internal processes of the whole RaspberryMatic/CCU ecosystem, thus introducing a dedicated user `hssled`. And the corresponding LED daemon that is about to access the LED files is supposed to be run under this particular user context. Therefore, the addon is in some need to be able to access these special rpi-rf-mod led files as this particular user.

But to be honest, because of this still open PR I am still maintaining a workaround just for the docker container use case which will still start that particular led daemon with full root permissions (cf. https://github.com/jens-maus/RaspberryMatic/blob/master/buildroot-external/overlay/base/etc/init.d/S06InitSystem#L139).

Nevertheless, I would be great to get rid of this special case and have the hssled daemon running under that sealed led user because this daemon also contains some network code which – on a very bad guess – could end up being misused due to RCE attacks. 
 
> All that said, it seems a lot of fuzz about a couple of lines in HAOS 😅 and we have already custom HW support, so 🤷‍♂️ However, before simply merging this, maybe there is another way to address this? 🤔

As said, the only potential alterantive way would be to keep the status like it is: keep the led daemon running in `root` context for that special HAos/docker use case. Nevertheless, I would really appreciate to have that particular udev rule taking care of that so that only a particular user/group can access these led nodes so that I can start that led daemon in that context. And even thought I understand your reasonings, I feel that in that particular case it should be fine to add that udev rule because it is limited to that particular LED case for that special `RPI-RF-MOD` homematic rf module case and does not apply to all general LED sysfs cases.

agners:
> However, to increase the security to some extend I actually introduced dedicated users for some of the internal processes of the whole RaspberryMatic/CCU ecosystem, thus introducing a dedicated user `hssled`. And the corresponding LED daemon that is about to access the LED files is supposed to be run under this particular user context. Therefore, the addon is in some need to be able to access these special rpi-rf-mod led files as this particular user.

Yeah I understand, and running things inside the container as user I do absolutely support. In fact I think it is quite bad that most containers just run things as root :cry: 

> But to be honest, because of this still open PR I am still maintaining a workaround just for the docker container use case which will still start that particular led daemon with full root permissions (cf. https://github.com/jens-maus/RaspberryMatic/blob/master/buildroot-external/overlay/base/etc/init.d/S06InitSystem#L139).

What you could do is start a service which just adjusts user rights. But that doesn't work when hot plugging is used I guess? :thinking: 

Missing udev support is a general architectural issue with containers. I guess the common work around here is to rely on the host's udev implementation.

> I feel that in that particular case it should be fine to add that udev rule because it is limited to that particular LED case for that special RPI-RF-MOD homematic rf module case and does not apply to all general LED sysfs cases.

Yeah I guess since we already have a custom driver anyways it is not that big departure architectural wise. 

Sorry I for being pedantic, but it's always tricky when creating such precedence. Also sometimes a better (and still simple) solution appears during discussion. But it really doesn't look to be the case here, so let's just use host OS udev :+1:

agners (approved):

PR #2855, opened 2023-10-23
Use alternative environment for release build bump

sairon:
By using a separate environment, we can postpone the bump in the version repository by adding a requirement for approval. Dev version will use default (empty string) environment which doesn't have any constraints.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -243,7 +243,7 @@ jobs:
   bump_version:
     name: Bump ${{ needs.prepare.outputs.channel }} channel version
     if: ${{ github.repository == 'home-assistant/operating-system' }}
-    environment: "${{ github.event_name == 'release' && 'version_release' || '' }}"
+    environment: "${{ github.event_name == 'release' && needs.prepare.outputs.channel || '' }}"
     needs: [ build, prepare ]
     runs-on: ubuntu-22.04
 
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -241,8 +241,9 @@ jobs:
             output/images/haos_ova*.qcow2.xz
 
   bump_version:
-    name: Bump dev channel version
+    name: Bump ${{ needs.prepare.outputs.channel }} channel version
     if: ${{ github.repository == 'home-assistant/operating-system' }}
+    environment: "${{ github.event_name == 'release' && 'version_release' || '' }}"
     needs: [ build, prepare ]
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -241,8 +241,9 @@ jobs:
             output/images/haos_ova*.qcow2.xz
 
   bump_version:
-    name: Bump dev channel version
+    name: Bump ${{ needs.prepare.outputs.channel }} channel version
     if: ${{ github.repository == 'home-assistant/operating-system' }}
+    environment: "${{ github.event_name == 'release' && 'version_release' || '' }}"
```
agners, line 246: While at it, we could have two environments directly, one for beta and one for stable. So, how about:

```
    environment: "${{ github.event_name == 'release' &&  needs.prepare.outputs.channel || '' }}"
```

sairon: Sure, updated :+1:

agners (approved):

.github/workflows/build.yaml
```diff
@@ -241,8 +241,9 @@ jobs:
             output/images/haos_ova*.qcow2.xz
 
   bump_version:
-    name: Bump dev channel version
+    name: Bump ${{ needs.prepare.outputs.channel }} channel version
     if: ${{ github.repository == 'home-assistant/operating-system' }}
+    environment: "${{ github.event_name == 'release' && needs.prepare.outputs.channel || '' }}"
```
ludeeus, line 246: Whats the reason behind providing a `""` environment here, instead of just using `needs.prepare.outputs.channel` like on core <https://github.com/home-assistant/core/blob/dev/.github/workflows/builder.yml#L53>

sairon: Originally the idea was to have a single pre-defined environment with protections created before the first run, and not to create any other environments in case they don't exist yet (which is what GH does - if you specify a non-existing `environment`, it creates it). But after changing it to channel name per Stefan's suggestion, creating a third one for dev wouldn't do any harm now...

agners: Hm, I was under the assumption that an environment always has some protection rule, but it seems that can be bypassed completely. I agree we should just use channel directly, and create a `dev` environment without protection rules. :+1: 

There is another problem in our current build pipeline: Once we release a final release, beta should receive a bump as well (see https://github.com/home-assistant/core/blob/dev/.github/workflows/builder.yml#L309-L316). 

@sairon can you add this to the pipeline as well?

PR #2864, opened 2023-10-26
RaspberryPi: Update kernel 6.1.58 - stable_20231024

sairon:


agners (changes requested):
I think we can remove `buildroot-external/board/raspberrypi/patches/linux/0001-Bluetooth-hci_sync-Fix-handling-of-HCI_QUIRK_STRICT_.patch`. This patch has been shipped with 6.1.57 stable release.

sairon:
@agners Ahhh, thanks! It's in my working directory, forgot to stage that into the commit. Of course, it'd not build without the patch removed.

agners:
> @agners Ahhh, thanks! It's in my working directory, forgot to stage that into the commit. Of course, it'd not build without the patch removed.

I was wondering if that built :sweat_smile: . Ok all clear :+1:

agners (approved):

PR #2871, opened 2023-10-27
Add checkout step to bump_version job in the build workflow

sairon:
It's required both for the artifact index and to run the RPi Imager bump local action.


agners (approved):
This also fixes the issue with the index generation for dev builds :+1: 

Thanks!

PR #2872, opened 2023-10-27
Optimize build cache for dev builds

sairon:
* Remove downloaded files cache, as it doesn't save that much time and it can't fit into the repo cache limit, randomly causing eviction of CC object cache for a single board.
* Limit saving of the object cache only to the dev branch, because of the restrictions for the cache access limit us from effectively using the cache for rc/main branches anyway.
* Adjust names of the steps a bit for clarity.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -228,6 +228,12 @@ jobs:
           path: /mnt/cache/cc
           key: haos-cc-${{ matrix.board.id }}-${{ github.run_id }}
 
+      - name: Print cache stats
+        run: |
+          echo "Cache size: $(du -sh /mnt/cache/cc)"
+          echo "Files total: $(find /mnt/cache/cc -mindepth 1 -type f | wc -l)"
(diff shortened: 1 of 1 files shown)


agners (approved):
I don't quite understand why the download cache isn't the first to get evicted. Probably some kind of race condition.

In any case, DL cache is not working currently because it's most often falls victim too cache eviction, hence just removing it is the right approach.

The current implementation is anyways not ideal as it only uploads the cache of one board, which isn't really containing all the source for other boards. Ideally we would have a per-board download cache, but this would require massively more space.

sairon:
FWIW here's an example of build where some of the object file caches were missing: https://github.com/home-assistant/operating-system/actions/runs/6376768727

The DL cache was missing there too - maybe got pushed out by another build. While I don't fully understand it either, with the changes in this PR such situation shouldn't happen anymore, since all object file caches should fit in the limit.

sairon:
Let's just add few debug prints about cache stats and see how it behaves over time.

agners (approved):

PR #2879, opened 2023-10-30
Add vyasa support

panicking:
Add vyasa support. Test was done using tinker board to dowload container. Missing point are:
- how to deploy container or reuse exists one
- how to test it connected to the cloud


agners:
This board doesn't fit into our requirements to be added as a supported platform (https://github.com/home-assistant/architecture/blob/master/adr/0017-hardware-screening-os.md).

Also, as mentioned in https://github.com/home-assistant/operating-system/pull/2768#issuecomment-1733144764, we rather plan to phase out Arm 32-bit support.

PR #2880, opened 2023-10-30
Add test suite for Supervisor tests

sairon:
Add tests of Supervisor based on scenario used in [Supervisor build](https://github.com/home-assistant/supervisor/blob/main/.github/workflows/builder.yml).

Also improved output logs hierarchy and added a fixture that starts a fresh QEMU instance before every test module.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,6 +1,21 @@
+import json
+import logging
 import os
 
+from labgrid.driver import ShellDriver
 import pytest
+
+
+logger = logging.getLogger(__name__)
+
+
+@pytest.fixture(autouse=True, scope="module")
+def restart_qemu(strategy):
+    """Use fresh QEMU instance for each module."""
+    if strategy.status.name == "shell":
+        logger.info("Restarting QEMU before %s module tests.", strategy.target.name)
+        strategy.transition("off")
+        strategy.transition("shell")
 
 
 @pytest.hookimpl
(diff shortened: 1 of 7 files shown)


tests/supervisor_test/test_supervisor.py
```diff
@@ -0,0 +1,195 @@
...
+                    )
+                    break
+            except ExecutionError:
+                pass  # avoid failure when the container is restarting
+            sleep(5)
+        else:
+            raise AssertionError("Supervisor did not update in time")
+
+
+@pytest.mark.dependency(depends=["test_check_supervisor"])
```
agners, line 90: Shouldn't this be

```suggestion
@pytest.mark.dependency(depends=["test_update_supervisor"])
```

tests/conftest.py
```diff
@@ -1,8 +1,23 @@
...
+logger = logging.getLogger(__name__)
+
+
+@pytest.fixture(autouse=True, scope="module")
+def restart_qemu(strategy):
+    """Use fresh QEMU instance for each module."""
+    if strategy.status.name == "shell":
+        logger.info("Restarting QEMU before %s module tests.", strategy.target.name)
+        strategy.transition("off")
+        strategy.transition("shell")
```
agners, line 18: This seems to be unused currently?

agners (commented):
The tests look very clean, awesome.

We probably will add more over time, e.g. check `ha resolution info`. Also, once we also make use of this HAOS/Labgrid setup in the Supervisor repository, we probably should store them there. But these things we can build up in follow up PRs.

tests/conftest.py
sairon: It is used, since it has `autouse=True, scope="module"` - so it's called before every test module. Similarly to `setup_module`, however `setup_module` doesn't have the required scope available.

agners: Ah I see :+1:

tests/supervisor_test/test_supervisor.py
sairon: Right, there would be subtle difference - theoretically `test_update_supervisor` could fail due to a timeout yet this test would pass. So it's better to have use this deterministic non-racy dependency :)

PR #2888, opened 2023-10-31
Separate artifacts index update into separate workflow

agners:


.github/workflows/build.yaml
```diff
@@ -247,6 +241,14 @@ jobs:
     needs: [ build, prepare ]
     uses: ./.github/workflows/test.yaml
 
+  update_index:
+    name: Update artifacts index
+    if: ${{ github.event_name != 'release' }}
+    needs: [ build, prepare ]
+    uses: ./.github/workflows/artifacts-index.yaml
```
sairon, line 248: I might prefer using a repository reference (first option in [docs](https://docs.github.com/en/actions/using-workflows/reusing-workflows#calling-a-reusable-workflow)), so if you change the definition of the artifact index workflow and re-run the job, the new one is used. If we keep it like this, you're stuck to what's been defined when the build job was started.

```suggestion
    uses: home-assistant/operating-system/.github/workflows/artifacts-index.yaml@dev
```

.github/workflows/artifacts-index.yaml
```diff
@@ -0,0 +1,75 @@
...
+on:
+  # Manual test of specified version
+  workflow_dispatch:
+    inputs:
+      version:
+        description: Version of HAOS to build index for
+        required: true
+        type: string
+
+ # Called by other workflows (e.g. build.yaml)
```
sairon, line 12:
```suggestion
  # Called by other workflows (e.g. build.yaml)
```

.github/workflows/artifacts-index.yaml
```diff
@@ -0,0 +1,75 @@
+name: Update artifacts index
+
+on:
+  # Manual test of specified version
```
sairon, line 4:
```suggestion
  # Manual run for specified version
```

sairon (commented):
Can't check the S3 commands work but they look fine on paper. Couple of nits in comments.

.github/workflows/build.yaml
agners: During development this makes sense I agree.

I guess an argument could be made that the version of when the job has been run needs too be used since that one has the same structure how things get uploaded to AWS (in case we move things around) :man_shrugging: 

I don't think that this will be the case, and it could be just more convinient to retrigger with the latest version of this workflow, so I am fine with your suggestion.

sairon (approved):

PR #2899, opened 2023-11-02
Add pruning of object file cache

sairon:
From build logs it seems that the `find` predicate should work fine for pruning old files, so let's do it.


sairon:
Should save some space especially in [RPi 4 cache](https://github.com/home-assistant/operating-system/actions/runs/6713139762/job/18244134879) now, since it contains about 50 % of unused files.

agners (approved):

PR #2906, opened 2023-11-02
Improve Bluetooth cache cleanup command

agners:
Use the find's delete flag to delete the files instead of spanning a shell for each file.


agners:
This should address the following error during bootup:

```
Nov 01 20:28:59 homeassistant systemd[1]: /usr/lib/systemd/system/haos-bt-cache.service:6: Ignoring unknown escape sequences: "find /var/lib/bluetooth/*/cache -mindepth 1 -type f -atime +7  -exec rm {} \;"
```

sairon (approved):

PR #2913, opened 2023-11-07
WIP: Use overlayfs instead of bind mounts for /etc

agners:
This is an attempt to use overlayfs instead of bind mounts to manage `/etc`. This comes with the downside that any service could write something into /etc, but doesn't seem to be a big down side to me, as we have all services under control essentially.

It makes the /etc setup quite a bit cleaner. The idea is to also use systemd's built-in `/etc/machine-id` support.

Currently this is missing migration support (from the current layout).


pvizeli:
We should modify the hassos-overlay to cleanup user modification in /ect/systemd/ - other than that, I think it's fine

agners:
This also did not work as intended due to ordering of things at startup: The machine-id is checked very very early during systemd startup:

```
Dec 05 08:20:56 homeassistant systemd[1]: systemd 252 running in system mode (+PAM -AUDIT -SELINUX +APPARMOR -IMA -SMACK +SECCOMP -GCRYPT -GNUTLS +OPENSSL -ACL +BLKID +CURL -ELFUTILS -FIDO2 -IDN2 -IDN +IPTC +KMOD +LIBCRYPTSETUP +LIBFDISK +PCRE2 -PWQUALITY -P11KIT -QRENCODE -TPM2 -BZIP2 -LZ4 -XZ +ZLIB -ZSTD -BPF_FRAME
Dec 05 08:20:56 homeassistant systemd[1]: Detected virtualization kvm.
Dec 05 08:20:56 homeassistant systemd[1]: Detected architecture x86-64.
Dec 05 08:20:56 homeassistant systemd[1]: Hostname set to <homeassistant>.
Dec 05 08:20:56 homeassistant systemd[1]: Installed transient /etc/machine-id file
```

So the main aim to uses systemd's machine-id support doesn't pan out with just this.

This might work out better if we'd be using a initramfs. But this obviously would be quite a departure from the current architecture.

Close this for now.

PR #2923, opened 2023-11-13
Clean up inconsistencies in kernel configs

sairon:
Clean up all kernel configs and fragments from non-existing kernel options, invalid choice values and choices that trigger warnings during kernel package configuration.

Here's an example of warnings triggered for Yellow:

```
.config:8531:warning: override: MODULE_COMPRESS_NONE changes choice state
.config:8536:warning: override: ZSWAP_COMPRESSOR_DEFAULT_LZ4 changes choice state
.config:8537:warning: override: ZSWAP_ZPOOL_DEFAULT_ZSMALLOC changes choice state
.config:8543:warning: override: CPU_FREQ_DEFAULT_GOV_ONDEMAND changes choice state
.config:8717:warning: override: reassigning to symbol CGROUP_HUGETLB
.config:8767:warning: symbol value 'm' invalid for XFRM
.config:8852:warning: symbol value 'm' invalid for MEDIA_CONTROLLER_DVB
.config:8972:warning: symbol value 'm' invalid for SND_HDA_I915
```

There were also some options that are set in our or default configs but end up patched by `KCONFIG_(DIS|EN)ABLE_OPT` in package makefiles, these options are now explicitly set in our fragments too. For example this was toggled for `generic_aarch64`:

```
CONFIG_DEFAULT_SECURITY_APPARMOR n -> y
CONFIG_DEFAULT_SECURITY_DAC y -> n
CONFIG_GCC_PLUGINS y -> n
```

The only goal of this commit is to make sure no warnings appear anymore and the resulting kernel configs remain unchanged. This will allow us to create tools for sanity checks of our kernel config overrides. There is one single change in `ova` config resulting from previously invalid `m` option for a bool value:

```diff
-# CONFIG_9P_FS_POSIX_ACL is not set
+CONFIG_9P_FS_POSIX_ACL=y
```


buildroot-external/kernel/device-support.config
```diff
@@ -67,7 +67,7 @@ CONFIG_RT2800USB_UNKNOWN=y
 # Multimedia core support
 CONFIG_MEDIA_SUPPORT=y
 CONFIG_MEDIA_CONTROLLER=y
-CONFIG_MEDIA_CONTROLLER_DVB=m
+CONFIG_MEDIA_CONTROLLER_DVB=y
```
agners, line 70: Where does that y came from? Isn't this global config applied at the end? 🤔

sairon: It's a `bool`, so it can't have `m`: https://github.com/torvalds/linux/blob/master/drivers/media/mc/Kconfig#L8

AFAIK Kconfig doesn't apply parent's (i.e. what the option `depends` on) value to children, so there is no semantics in setting it to `m`. And since the diff of all the configs before and after these changes is empty, it confirms that.

agners (approved):

PR #2943, opened 2023-11-23
Generate self-signed certificates for development

agners:
To simplify development generate a self-signed certificate on first build. Also make sure that the self-signed certificate is being added the RAUC keyring so that manual updates can be performed.

### DIFF
--- a/buildroot-external/scripts/rauc.sh
+++ b/buildroot-external/scripts/rauc.sh
@@ -2,6 +2,19 @@
 set -e
 
 
+function prepare_rauc_signing() {
+    local key="/build/key.pem"
+    local cert="/build/cert.pem"
+
+    if [ ! -f "${key}" ]; then
+        echo "Generating a self-signed certificate for development"
+        openssl req -x509 -newkey rsa:4096 -keyout "${key}" \
+                -out "${cert}" -days 3650 -nodes \
+                -subj "/O=HassOS/CN=HassOS Self-signed Development Certificate"
+    fi
+}
+
+
 function write_rauc_config() {
     mkdir -p "${TARGET_DIR}/etc/rauc"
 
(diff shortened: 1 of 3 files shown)


buildroot-external/scripts/rauc.sh
```diff
@@ -2,6 +2,19 @@
 set -e
 
 
+function prepare_rauc_signing() {
+    local key="/build/key.pem"
+    local cert="/build/cert.pem"
+
+    if [ ! -f "${key}" ]; then
```
sairon, line 9: Maybe we could add some check here, like `[ -z "$GITHUB_ACTIONS" ]` - otherwise it could lead to unexpected results for developers who fork this repository and expect they can use the CI builds without any changes. Or we can make sure the step for adding keys in GH workflow fails in case the variables are not defined.

buildroot-external/scripts/rauc.sh
```diff
@@ -19,10 +32,19 @@ function write_rauc_config() {
...
-        cp "${BR2_EXTERNAL_HASSOS_PATH}/ota/rel-ca.pem" "${TARGET_DIR}/etc/rauc/keyring.pem"
-    else
+    local cert="/build/cert.pem"
+
+    if [ "${DEPLOYMENT}" == "development" ]; then
         cp "${BR2_EXTERNAL_HASSOS_PATH}/ota/dev-ca.pem" "${TARGET_DIR}/etc/rauc/keyring.pem"
+
+        # Add local self-signed certificate (if not trusted by chain it is a
+        # self-signed certificate)
+        if ! openssl verify -CAfile "${BR2_EXTERNAL_HASSOS_PATH}/ota/dev-ca.pem" -no-CApath "${cert}"; then
```
sairon, line 42: We use `DEPLOYMENT` only to choose which CA/keyring should be deployed to the target. I'd prefer if I could do a "development build" of any commit without changing anything - now I need to edit the `meta` file if I want to have a self-signed build from the `main` or `rc` branch.

sairon (commented):
LGTM, just a few QoL suggestions.

buildroot-external/scripts/rauc.sh
agners: Technically, we could add the self-signed certificate in any case. I think the main reason I did it the way it is now was to prevent accidentally leak the self-signed certificate into a release... 

However, there is also the thought of same source -> same binary... If we change things in the image based on local environment (which certificate), the build is not really reproducible. In a way, I think it is good to fail if you can't rebuild it 100% the same way :man_shrugging:

buildroot-external/scripts/rauc.sh
agners: Yeah I prefer to solve this in GitHub actions.

buildroot-external/scripts/rauc.sh
sairon: >  to prevent accidentally leak the self-signed certificate into a release...

That's the point of the other comment. If we avoid that by not creating the certificate in the build/release pipeline, it should be fine.

>  If we change things in the image based on local environment (which certificate), the build is not really reproducible. In a way, I think it is good to fail if you can't rebuild it 100% the same way 🤷‍♂️

But why the `dev` branch should be an exception? Right now we're just making it more complicated if you want to bisect issues on other branches.

sairon: If you REALLY want to be sure it doesn't happen by accident, could we maybe add a possibility to override the behavior using an env variable, or showing and interactive prompt?

agners: > But why the `dev` branch should be an exception? Right now we're just making it more complicated if you want to bisect issues on other branches.

IMHO, a releases build and a dev/nightly build is different in the sense that for a release set/bump the version in source (to make it reproducible), whereas for dev/nightly builds the version number usually generated... So I backed this by that logic.

But yeah, it hurts bisectability. I've changed to always add the self-signed certificate, no matter the build type :+1:

PR #2947, opened 2023-11-28
Do not start hciuart.service if krnbt is used

agners:
Avoid starting (and failing to start) hciuart.service if krnbt is used. This avoid unnecessary failed services showing up.

### DIFF
--- a/buildroot-external/package/pi-bluetooth/hciuart.service
+++ b/buildroot-external/package/pi-bluetooth/hciuart.service
@@ -5,6 +5,7 @@ After=dev-serial1.device
 
 [Service]
 Type=forking
+ExecCondition=/bin/sh -c "if [ "$(cat /proc/device-tree/$(cat /proc/device-tree/aliases/bluetooth)/bluetooth/status)" == "okay" ]; then exit 1; fi"
 ExecStart=/usr/bin/btuart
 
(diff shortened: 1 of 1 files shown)


buildroot-external/package/pi-bluetooth/hciuart.service
```diff
@@ -5,6 +5,7 @@ After=dev-serial1.device
 
 [Service]
 Type=forking
+ExecCondition=/bin/sh -c "if [ "$(cat /proc/device-tree/$(cat /proc/device-tree/aliases/bluetooth)/bluetooth/status)" == "okay" ]; then exit 1; fi"
```
sairon, line 8: Wouldn't there be a problem with escaping? Also could be simplified:
```suggestion
ExecCondition=/bin/sh -c '[ "$(cat /proc/device-tree/$(cat /proc/device-tree/aliases/bluetooth)/bluetooth/status)" != "okay" ]'
```

agners: It did work on the shell but I wanted to double check. I considered single quote, but somehow I thought this would prevent `$()` from working. But that is not the case. So I prefer your solution. Thanks!

sairon (approved):

Issue #2944, opened 2023-11-24
Bluetooth adapter showing on Yellow with CM4 without Bluetooth (OS 11.2.rc1)


agners:
I'm not sure why, but as of OS 11.2.rc1 I have a bluetooth adapter showing up.
Could be related to supervisor but I don't notice any changes related to that.
It discovers hci0 but never inits on my HA Yellow (as my CM4 does not have wireless).
I don't have a big understanding of how the OS works but I assume it might be related to #2931 by @sairon.

Host logs:
```
Nov 23 21:35:19 homeassistant kernel: Bluetooth: hci0: command 0xfc18 tx timeout
Nov 23 21:35:27 homeassistant kernel: Bluetooth: hci0: BCM: failed to write update baudrate (-110)
Nov 23 21:35:27 homeassistant kernel: Bluetooth: hci0: Failed to set baudrate
Nov 23 21:35:29 homeassistant kernel: Bluetooth: hci0: command 0x0c03 tx timeout
Nov 23 21:35:37 homeassistant kernel: Bluetooth: hci0: BCM: Reset failed (-110)
...
```

Core:
```
Logger: bluetooth_auto_recovery.recover
Source: runner.py:188
First occurred: 22:35:37 (5 occurrences)
Last logged: 22:38:24

Bluetooth management socket connection lost: [Errno 22] Invalid argument
```

```
Logger: bluetooth_auto_recovery.recover
Source: components/bluetooth/util.py:78
First occurred: 22:35:16 (21 occurrences)
Last logged: 22:38:40

Could not determine the power state of the Bluetooth adapter hci0 [00:00:00:00:00:00] due to timeout after 5 seconds
Could not cycle the Bluetooth adapter hci0 [00:00:00:00:00:00]: [Errno 110] Operation timed out
Bluetooth adapter hci0 [00:00:00:00:00:00] could not be reset:
Closing Bluetooth adapter hci0 [00:00:00:00:00:00] failed: [Errno 9] Bad file descriptor
```

_Originally posted by @tjorim in https://github.com/home-assistant/operating-system/discussions/2940#discussioncomment-7655278_

agners:
I can reproduce this problem on my end. It seems that the kernel Bluetooth initialization creates an hci0 device in any case. I am not sure if we can prevent that though :thinking: 

The Bluetooth daemon seems to correctly detect that there is no controller:
```
[bluetooth]# show 
No default controller available
```

[...]

PR #2948, opened 2023-11-28
Remove kernel HCI driver if no WiFi/Bluetooth module present (#2944)

agners:
If the WiFi/Bluetooth module is not present on the SDIO bus, remove the HCI driver. This avoids hci0 interface to be present. Current Home Assistant Core versions show a Bluetooth device as soon as a hci device is present. With this change there won't be a Bluetooth device shown.

### DIFF
--- a/buildroot-external/package/pi-bluetooth/hcidisable.service
+++ b/buildroot-external/package/pi-bluetooth/hcidisable.service
@@ -4,7 +4,7 @@ ConditionPathExists=!/sys/bus/sdio/devices/mmc1:0001:1
 
 [Service]
 Type=forking
-ExecStart=/usr/sbin/rmmod hci_uart
+ExecStart=/usr/sbin/modprobe -r hci_uart
 
 [Install]
 WantedBy=hassos-hardware.target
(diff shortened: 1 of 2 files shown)


buildroot-external/package/pi-bluetooth/hcidisable.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=Remove HCI kernel driver if WiFi/Bluetooth module is not present
+ConditionPathExists=!/sys/bus/sdio/devices/mmc1:0001:1
+
+[Service]
+Type=forking
+ExecStart=/usr/sbin/rmmod hci_uart
```
sairon, line 7: `rmmod` will fail with exit code 1 and message if the module is not loaded (e.g. with `krnbt=off`), `modprobe -r` exits gracefully and also removes eventual dependencies (even though there shouldn't be any in this case). 

```suggestion
ExecStart=/usr/sbin/modprobe -r hci_uart
```

sairon (changes requested):

pelwell:
This change breaks Bluetooth in cases where the user has intentionally disabled WiFi using `dtoverlay=disable-wifi`.
It's also been unnecesary since February 2025 when the firmware started also disabling the UART on CM4s and CM5s which have no WiFi/BT module. If the matter had been raised at e.g. https://github.com/raspberrypi/linux  back when #2944 was live then the firmware would have been patched earlier.

See https://github.com/raspberrypi/linux/issues/6930 for background.

Issue #2841, opened 2023-10-18
No wifi driver for RTL8821CE


bartlebi:
### Describe the issue you are experiencing

Home Assistant OS appears to be missing wifi support for on-board wireless card RTL8821CE.
I am using Hassos in proxmox VA.
I have thrown a PCI device into the virtual machine, but HA does not see any wifi device. 
I think there is a driver missing for this chip. What can I check or do on my end to solve the problem or may you just include driver in image?

### What operating system image do you use?

generic-x86-64 (Generic UEFI capable x86-64 systems)

### What version of Home Assistant Operating System is installed?

	Home Assistant OS 11.0

### Did you upgrade the Operating System.

Yes

### Steps to reproduce the issue

Just boot

### Anything in the Supervisor logs that might be useful for us?

```txt
No
```

### Anything in the Host logs that might be useful for us?

```txt
No
```

## System Information

version | core-2023.10.3
-- | --
installation_type | Home Assistant OS
dev | false
hassio | true
docker | true
user | root
virtualenv | false
python_version | 3.11.5
os_name | Linux
os_version | 6.1.56
arch | x86_64
timezone | Europe/Moscow
config_dir | /config

<details><summary>Home Assistant Community Store</summary>

GitHub API | ok
-- | --
GitHub Content | ok
GitHub Web | ok
GitHub API Calls Remaining | 4873
Installed Version | 1.33.0
Stage | running
Available Repositories | 1313
Downloaded Repositories | 18

</details>

<details><summary>Home Assistant Cloud</summary>

logged_in | false
-- | --
can_reach_cert_server | ok
can_reach_cloud_auth | ok
can_reach_cloud | ok

</details>

<details><summary>Home Assistant Supervisor</summary>

host_os | Home Assistant OS 11.0
-- | --
update_channel | stable
supervisor_version | supervisor-2023.10.0
agent_version | 1.6.0
docker_version | 24.0.6
disk_total | 30.8 GB
disk_used | 8.3 GB
healthy | true
supported | true
board | ova
supervisor_api | ok
version_api | ok
installed_addons | Samba share (10.0.2), Terminal & SSH (9.7.1), Mosquitto broker (6.3.1), go2rtc (1.8.1), RPC Shutdown (2.4), SQLite Web (3.2.1), Zigbee2MQTT (1.33.1-1), Node-RED (14.6.2), MariaDB (2.6.1), File editor (5.6.0), go2rtc master hardware (master-hardware)

</details>

[...]

Issue #2806, opened 2023-10-12
WiFi dongle based on Mediatek mt7921au chipset


anatoly-savchenkov:
### Describe the issue you are experiencing

I run HA OS as a VM on HP MicroServer and have to use a WiFi connection. MicroServer has no built in WiFi, so I looked here https://github.com/morrownr/USB-WiFi/blob/main/home/USB_WiFi_Chipsets.md and seems today's' fastest option is a dongle based on Mediatek mt7921au chipset. So, I bought one and found that it's disabled in the HA OS flavor of Linux kernel.

Please treat this as a vote for enabling CONFIG_MT7921U (and most probably CONFIG_BT_HCIBTUSB_MTK for BT part) here https://github.com/home-assistant/operating-system/pull/2761 :)

### What operating system image do you use?

ova (for Virtual Machines)

### What version of Home Assistant Operating System is installed?

10.5

### Did you upgrade the Operating System.

Yes

### Steps to reproduce the issue

1. Use a WiFi card based on Mediatek mt7921au chipset. 
2. Find that it doesn't work
3. Realize that it's expected since many WiFi USB drivers are disabled

### Anything in the Supervisor logs that might be useful for us?

```txt
none
```

### Anything in the Host logs that might be useful for us?

```txt
none
```

sairon:
The mt7921u driver has been added in #2951, support for more cards with proprietary USB IDs in #2962.

Issue #2776, opened 2023-09-28
Missing driver for ALL mediatek wifis/devices


patryk4815:
### Describe the issue you are experiencing

Missing drivers for **ALL** mediatek devices.
For example MT7922 need `mt7921e.ko`
Please enable in kernel as module: `CONFIG_MT7921E=m`

### What operating system image do you use?

generic-x86-64 (Generic UEFI capable x86-64 systems)

### What version of Home Assistant Operating System is installed?

10.5.3

### Did you upgrade the Operating System.

Yes

### Steps to reproduce the issue

n/a

### Anything in the Supervisor logs that might be useful for us?

```txt
n/a
```

### Anything in the Host logs that might be useful for us?

```txt
n/a
```

patryk4815:
Manually recompiled kernel + added firmware:

For some reason i had to copy full director `./output/build/linux-firmware-20230515/mediatek` into image.
❗❗Because there is missing `BT_RAM_CODE_MT7922_1_1_hdr.bin` firmware file.
It is missing in buildroot, i don't see flag for enabling it - https://github.com/buildroot/buildroot/blob/master/package/linux-firmware/linux-firmware.mk

<img width="1052" alt="Zrzut ekranu 2023-09-28 o 05 38 10" src="https://github.com/home-assistant/operating-system/assets/3074260/940ded66-a536-4379-8da1-3302f8841c33">


Diff for build:
```
diff --git a/buildroot-external/board/pc/ova/kernel.config b/buildroot-external/board/pc/ova/kernel.config
index f94e70358..4844be610 100644
--- a/buildroot-external/board/pc/ova/kernel.config
+++ b/buildroot-external/board/pc/ova/kernel.config
@@ -111,3 +111,34 @@ CONFIG_IXGBEVF=m
 CONFIG_I40EVF=m
 CONFIG_MLX5_CORE=m
 CONFIG_MLX5_CORE_EN=y
+
+
+CONFIG_MT7601U=m
+CONFIG_MT76_CORE=m
+CONFIG_MT76_USB=m
+CONFIG_MT76_SDIO=m
+CONFIG_MT76x02_LIB=m
+CONFIG_MT76x02_USB=m
+CONFIG_MT76_CONNAC_LIB=m
+CONFIG_MT76x0_COMMON=m
+CONFIG_MT76x0U=m
+CONFIG_MT76x0E=m
+CONFIG_MT76x2_COMMON=m
+CONFIG_MT76x2E=m
[...]

Issue #2725, opened 2023-09-02
Missing WiFi Support for AR9462 Qualcomm Atheros Wireless Network Adapter


juliovegap:
### Describe the issue you are experiencing

Despite installing `apk add linux-firmware` WiFi Atheros module is not loaded by Alpine Linux Kernel

### What operating system image do you use?

generic-x86-64 (Generic UEFI capable x86-64 systems)

### What version of Home Assistant Operating System is installed?

6.1.45

### Did you upgrade the Operating System.

Yes

### Steps to reproduce the issue

1. lshw -C network
...

### Anything in the Supervisor logs that might be useful for us?

```txt
N/A
```

### Anything in the Host logs that might be useful for us?

```txt
N/A
```

## System Information

version | core-2023.8.4
-- | --
installation_type | Home Assistant OS
dev | false
hassio | true
docker | true
user | root
virtualenv | false
python_version | 3.11.4
os_name | Linux
os_version | 6.1.45
arch | x86_64
timezone | Europe/Madrid
config_dir | /config

<details><summary>Home Assistant Community Store</summary>

GitHub API | ok
-- | --
GitHub Content | ok
GitHub Web | ok
GitHub API Calls Remaining | 5000
Installed Version | 1.32.1
Stage | running
Available Repositories | 1277
Downloaded Repositories | 6

</details>

<details><summary>Home Assistant Cloud</summary>

logged_in | false
-- | --
can_reach_cert_server | ok
can_reach_cloud_auth | ok
can_reach_cloud | ok

</details>

<details><summary>Home Assistant Supervisor</summary>

host_os | Home Assistant OS 10.5
-- | --
update_channel | stable
supervisor_version | supervisor-2023.08.1
agent_version | 1.5.1
docker_version | 23.0.6
disk_total | 116.7 GB
disk_used | 9.2 GB
healthy | true
supported | true
board | generic-x86-64
supervisor_api | ok
version_api | ok
installed_addons | File editor (5.6.0), Log Viewer (0.15.1), Spotify Connect (0.12.4), ZeroTier One (0.17.0), Plex Media Server (3.2.4), InfluxDB (4.7.0), Grafana (9.0.3), Let's Encrypt (4.12.9), Duck DNS (1.15.0), SQLite Web (3.9.2), Advanced SSH & Web Terminal (15.0.7)

</details>

<details><summary>Dashboards</summary>

dashboards | 1
-- | --
resources | 0
mode | auto-gen

</details>

[...]

PR #2951, opened 2023-11-29
Enable more WiFi drivers for all platforms

sairon:
Preemptively enable larger set of WiFi drivers for all platforms and add more firmwares for them with the aim to harmonize WiFi device support among all boards and to have implicit support of devices that users might want to use.  Targets `generic_aarch64`, `generic_x86_64` and `ova` also include options and firmwares for cards that are using PCI/PCIe bus - support for these is in a separate config fragment.

Especially the `generic_x86_64` is currently very tight with the rootfs space, so I had to do some triaging and select only sensible drivers and firmwares - especially archaic PCMCIA devices or devices not supporting only 802.11g or lower standards were among the first that I removed during the triaging - we can consider enabling those but this time on an someone's explicit need to have them enabled.

This closes #2815 and replaces large part of #2761, also potentially addresses (at least) these: #2806, #2783, #2841, #2776, #2725, #2600


buildroot-external/kernel/device-support-wireless-pci.config
```diff
@@ -0,0 +1,73 @@
...
+CONFIG_RTL8821AE=m
+CONFIG_RTW88_8822BE=m
+CONFIG_RTW88_8822CE=m
+CONFIG_RTW88_8723DE=m
+CONFIG_RTW88_8821CE=m
+CONFIG_RTW89_8852AE=m
+CONFIG_RTW89_8852CE=m
+
+
+# TODO: REMOVE - SANITY CHECK
```
agners, line 53: Is that something we should do before merging?

agners (approved):

PR #2953, opened 2023-11-30
Fix Supervisor image corruption detection

agners:
When multiple images match the reference, multiple IDs are passed as a
single argument to docker image rm, leading to an error:
Error response from daemon: page not found

Make sure to pass the ids as separate argument to make the delete work
in any case.

### DIFF
--- a/buildroot-external/rootfs-overlay/usr/sbin/hassos-supervisor
+++ b/buildroot-external/rootfs-overlay/usr/sbin/hassos-supervisor
@@ -24,9 +24,13 @@ if [ -f "${SUPERVISOR_STARTUP_MARKER}" ]; then
     echo "[WARNING] Supervisor container did not remove the startup marker file. Assuming container image or container corruption."
     docker container rm --force hassio_supervisor || true
     SUPERVISOR_CONTAINER_ID=""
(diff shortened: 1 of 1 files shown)


agners:
Docker might return image IDs in random order, with that uniq doesn't work. While it shouldn't hurt, it can lead to confusing errors, e.g.

```
$ docker image rm --force sha256:241f0f3edb06f549f150701febf21798f179d19fc0a8b74e7694e67af8fa8433 sha256:241f0f3edb06f549f150701febf21798f179d19fc0a8b74e7694e67af8fa8433
Untagged: agners/amd64-hassio-supervisor:latest
Untagged: agners/amd64-hassio-supervisor@sha256:8ee9f1ac4fb286c36f27ba03e1264c16586f64c36aa19fde0d1089cb644ff0b5
Deleted: sha256:241f0f3edb06f549f150701febf21798f179d19fc0a8b74e7694e67af8fa8433
Error response from daemon: No such image: sha256:241f0f3edb06f549f150701febf21798f179d19fc0a8b74e7694e67af8fa8433
```

sairon (approved):

PR #2955, opened 2023-11-30
Fix RAUC certificate check

agners:
Check the right environment variable when determining if RAUC certificates through secrets are provided.


sairon (approved):
:see_no_evil:

PR #2965, opened 2023-12-04
Add image size check to hdd-image.sh

sairon:
There is no sanity check when creating OS images, so when some of the partitions gets too big, part of its data may get overwritten by the following partition, resulting in corrupted image. Add checks for the defined partition sizes and bail out if they're too big.


agners (approved):
Ultimately we really should switch to using Genimage, which likely will take care of such things. But until then, this is a very useful check! :+1:

PR #2998, opened 2023-12-18
Adjust build to allow running from forks or feature branches

sairon:
Make it possible to run build on feature branches by adding a flag that can be used to select whether the build output will be uploaded to the R2 artifacts bucket or kept only as build artifact on GH. The latter is also used for 3rd party repos, allowing builds in forked repositories.

Feature builds are using Unix timestamp as the dev version suffix. This makes them easily distiguishable, yet it makes them appear to be newer than standard daily dev version builds when compared by AwesomeVersion.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -292,7 +292,23 @@ jobs:
           path: |
             output/images/haos_${{ matrix.board.id }}-${{ needs.prepare.outputs.version_full }}.img.xz
 
-      - name: Upload OS KVM image artifact
+      - name: Upload RAUC bundle artifact
+        uses: actions/upload-artifact@v4
+        if: ${{ github.event_name != 'release' && needs.prepare.outputs.publish_build != 'true' }}
+        with:
+          name: haos_${{ matrix.board.id }}-${{ needs.prepare.outputs.version_full }}.raucb
+          path: |
+            output/images/haos_${{ matrix.board.id }}-${{ needs.prepare.outputs.version_full }}.raucb
+
+      - name: Upload Open Virtualization Format (OVA) artifact
+        uses: actions/upload-artifact@v4
+        if: ${{ github.event_name != 'release' && needs.prepare.outputs.publish_build != 'true' && matrix.board.id == 'ova' }}
+        with:
+          name: haos_${{ matrix.board.id }}-${{ needs.prepare.outputs.version_full }}.ova
+          path: |
+            output/images/haos_${{ matrix.board.id }}-${{ needs.prepare.outputs.version_full }}.ova
+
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -11,6 +11,11 @@ on:
         description: 'List of boards to build (comma separated identifiers)'
         required: false
         type: string
+      publish:
+        description: 'Publish build artifacts to R2 (not applicable to forks)'
+        required: true
+        type: boolean
+        default: true
```
agners, line 18: Could we maybe simplify this by setting this to `default: ${{ github.repository }}" == "home-assistant/operating-system" && "${{ github.event_name }}" == "release"`

.github/workflows/build.yaml
```diff
@@ -234,22 +260,54 @@ jobs:
...
+            output/images/haos_${{ matrix.board.id }}-${{ needs.prepare.outputs.version_full }}.vmdk.zip
+
+      - name: Upload RAUC update artifact
+        uses: actions/upload-artifact@v3
+        if: ${{ github.event_name != 'release' && needs.prepare.outputs.publish_build != 'true' }}
+        with:
+          name: haos_${{ matrix.board.id }}-${{ needs.prepare.outputs.version_full }}.raucb
+          path: |
+            output/images/haos_${{ matrix.board.id }}-${{ needs.prepare.outputs.version_full }}.raucb
+
```
agners, line 302: I guess there is no way using globbing or someting?

There are more images, a similar per image upload has been removed in 9f5b9f8afead535ac6c714718b2da862702d16ea.

agners (changes requested):

.github/workflows/build.yaml
sairon: Well, it would be less cluttered then, but unfortunately not. First, this is default only for `workflow_dispatch`, i.e. on-demand build from the Actions page. Second, `default` has to be boolean, you can not use `github` context here yet.

.github/workflows/build.yaml
sairon: You can use glob or multiple paths - but in that case it creates a .zip archive of all the matched files. And that's yet another hidden _drawback_  (?) of using artifacts, it'll be a .zip archive every time. But at least GH sends the correct content-disposition headers, so you know it's a zip when you download it.

agners: Yeah zip is not ideal, as it gets big and you typically just want to download one.

But I think adding `vdi.zip` `vhdx.zip` and `ova` for completeness make sense still, or any particular reason those are not part of the upload?

agners (approved):

PR #3006, opened 2023-12-20
Add script for checking resulting kernel config

sairon:
There is bunch of kernel config options that are not propagated correctly to the kernel configuration after fragments are merged and processed by Kconfig. Current Buildroot tools are not good at discovering these - while we cleaned up most inconsistencies by using linux-diff-config and output from the merge_config.sh script, there are still options that were removed or get a different value than intended because of dependencies, etc.

This commit adds a Python script that is using Kconfiglib to parse current kernel's Kconfig files and the generated .config and compare the requested values from individual kernel config fragments. The script can be used manually by running `make linux-check-dotconfig` from the buildroot directory (with path to BR2_EXTERNAL directory set) and it's called also from the CI, where it generates Github Workflow warning annotations when some of the values are not present or when set incorrectly.

The kconfiglib.py is checked-in to the repo as well, because the library is currently abandoned on PyPI and packaged version has a bug that causes errors parsing Kconfigs in newer Linux versions, fixed in outstanding pull request ulfalizer/Kconfiglib#119 - so version from this PR is used here.

If pypi/support#2526 is ever resolved, we could remove it from our repo and use pip for installing the package as a requirement during build of the build container.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -224,6 +224,15 @@ jobs:
             ${{ needs.prepare.outputs.build_container_image }} \
             make BUILDDIR=/build ${{ matrix.board.defconfig }}
 
+      - name: Check Linux config
+        run: |
+          docker run --rm --privileged -v "${GITHUB_WORKSPACE}:/build" \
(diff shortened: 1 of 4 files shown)


agners (approved):
LGTM, very nice :+1:

PR #3011, opened 2023-12-21
Remove duplicated step uploading ova QEMU image for the test job

sairon:
Instead of uploading the file twice with a fixed name, upload it in the same step that is used for unpublished builds and pass the version string to the test job.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -20,7 +20,7 @@ on:
         type: boolean
         default: true
       version:
-        description: Version of HAOS to test (used in name of the qcow2 image artifact)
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yaml
```diff
@@ -19,6 +19,10 @@ on:
         required: false
         type: boolean
         default: true
+      version:
+        description: Version of HAOS to test (used in name of the qcow2 image artifact)
```
agners, line 23:
```suggestion
        description: Version of HAOS to test (as used in the name of the qcow2 image artifact)
```

agners (approved):

PR #3012, opened 2023-12-21
Set Raspberry Pi 2 Supervisor architecture correctly

agners:
The Raspberry Pi 2 was based on Cortex-A7 in first revisions, then Cortex-A53. However, both ar armv7 architectures, and that is also how the Operating System itself is configured. Use the correct architecture for Supervisor.

The hassos-supervisor script should recreate the container automatically on first boot after update.

Since Supervisor itself is written in Python this shouldn't affect much. However, the Supervisor uses it's architecture as default architecture when building local add-on and uses it as default architecture for multi-arch container images. This is also where this error got noticed (see https://github.com/home-assistant/supervisor/issues/4402#issuecomment-1865979421).


jens-maus:
Please note that we had similar issues in the RaspberryMatic project in summer where suddently ghcr was returning with `404` errors. We could also solve that by changing our docker container generation from `armhf` to `armv7` as well and also adapted our HA addon definitions. (cf. https://github.com/jens-maus/RaspberryMatic/commit/e8728692222cc16493a2a852742b2fedec98050a, https://github.com/jens-maus/RaspberryMatic/issues/2393)

Good to see that this is now also hopefully solved in the rpi2 builds which AFAICS were the last build targets with `armhf` in the meta files.

agners:
> Please note that we had similar issues in the RaspberryMatic project in summer where suddently ghcr was returning with `404` errors. We could also solve that by changing our docker container generation from `armhf` to `armv7` as well and also adapted our HA addon definitions. (cf. [jens-maus/RaspberryMatic@e872869](https://github.com/jens-maus/RaspberryMatic/commit/e8728692222cc16493a2a852742b2fedec98050a), [jens-maus/RaspberryMatic#2393](https://github.com/jens-maus/RaspberryMatic/issues/2393))

Hm, the issue seems slightly different: The problem was at installation time, and was on the Tinkerboard. The Tinkerboard seems to use armv7 Supervisor by default, so this should have worked fine from HAOS/Supervisor perspective.

However, armhf in Debian is arm/v7, whereas Supervisor defined it to be arm/v6 (see https://github.com/home-assistant/supervisor/blob/2023.12.0/supervisor/docker/interface.py#L56). This is because we aligned more with Raspbian/Raspberry Pi OS in that perspective which defined their `armhf` architecture to be ARMv6 (see https://forums.raspberrypi.com/viewtopic.php?t=4256#p55723). I guess the problem in your case was simply that the multi-arch `arm/v6` image did not exist.

sairon (approved):

PR #3035, opened 2024-01-04
Remove rootfstype from RPi5 cmdline

sairon:
Defining rootfstype is not necessary and removing it from cmdline would be more tricky than removing it from boot.scr on platforms that use UBoot if we eventually decide to switch to a different FS.


sairon:
Closing in favor of #3037

PR #3050, opened 2024-01-07
updated grub.cfg

jscongdon:
updated grub.cfg to point loading and saving of the env to include the file specification

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,0 +1,3 @@
+# Default ignored files
+/shelf/
+/workspace.xml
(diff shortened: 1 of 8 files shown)


jscongdon:
This issue affects x86 platforms
On a fresh install of the OS, the default order for the slot-loading is correct: `A B`
When a new OS version is installed, the slot order is updated in `/EFI/BOOT/grubenv` to `B A` correctly but on boot `load_env` is not loading `/EFI/BOOT/grubenv` correctly and uses the default order of `A B` that is set in the `/EFI/BOOT/grug.cfg`
This fix points the loading/saving of grubenv directly to the `/EFI/BOOT/grubenv` file.

This fix was tested with OS version 11.1 loaded in Slot A and 11.2 loaded in Slot B. Initial manual intervention was needed in order to load 11.2 in Slot B. After this fix, Slot B was loaded on boot. When version 11.3 was released, the OS upgrade was performed via the UI, and on reboot the device booted to Slot A containing the updated version of 11.3

Issue #3079, opened 2024-01-16
rpi5: `overlay_prefix` required in `config.txt` for dtoverlay support


jens-maus:
### Describe the issue you are experiencing

Due to the changed bootloading chain in the current RaspberryPi5 support it seems loading device tree overlays specified in the `/mnt/boot/config.txt` bootloader config file seems to be broken at the moment.

Even thought the documentation on the tryboot functionality in the RaspberryPi bootloader states that `os_prefix` is added to the existing `overlay_prefix` option (default: `overlays/`), this seems to be NOT the case at the moment (cf. https://www.raspberrypi.com/documentation/computers/config_txt.html#overlay_prefix)

This has become visible in the optional support for the `RPI-RF-MOD` Homematic RF module which requires to adapt the `/mnt/boot/config.txt` file, adding a `dtoverlay=rpi-rf-mod` statement to it. However, doing so in the current `config.txt` of the rpi5 version of HomeAssistantOS does not load the `rpi-rf-mod.dtbo` file. After looking at the output of the RaspberryPi bootloader one can see that the bootloader tries to load that dtoverlay, but fails:

<img width="433" alt="Bildschirmfoto 2024-01-16 um 10 11 56" src="https://github.com/home-assistant/operating-system/assets/5887769/35a7596f-37fa-4874-b43f-6d2b9f118290">

After adding the following additional (usually not required) `overlay_prefix` line to the `config.txt` file:

```
overlay_prefix=slot-A/overlays/
```

the bootloader can correctly load the `rpi-rf-mod.dtbo` (not the number of bytes mentioned in the rpi-rf-mod line):

<img width="426" alt="Bildschirmfoto 2024-01-16 um 10 13 46" src="https://github.com/home-assistant/operating-system/assets/5887769/21238f81-dd2a-44eb-a942-8089d85f6bbc">

Thus, it seems that with the tryboot functionality in place the stated `${os_prefix}${overlay_prefix}` prefixing (cf. https://www.raspberrypi.com/documentation/computers/config_txt.html#overlay_prefix) does not seem to be in place or work correctly in the context of HomeAssistantOS.

BTW: It can also be seen in the bootloader that it is also not able to load the `vc4-kms-v3d.dtbo` with the missing `overlay_prefix` line.

### What operating system image do you use?

RaspberryPi 5

### What version of Home Assistant Operating System is installed?

11.4

### Did you upgrade the Operating System.

No

### Steps to reproduce the issue

[...]

PR #3080, opened 2024-01-16
add optional rpi-rf-mod dtoverlay block

jens-maus:
This PR adds the "usual" dtoverlay block for optionally enabling the `rpi-rf-mod.dtbo` device tree overlay to support the `RPI-RF-MOD` HAT hardware for HomeMatic(IP) support.

Please note that, while this PR will allow users to more easily enable support for this special dtoverlay, due to #3079 it is currently not correctly loaded and would require an additional `overlay_prefix` line which I intentionally omitted here until #3079 is fixed.

### DIFF
--- a/buildroot-external/board/raspberrypi/rpi5-64/config.txt
+++ b/buildroot-external/board/raspberrypi/rpi5-64/config.txt
@@ -43,6 +43,12 @@ arm_boost=1
 os_prefix=slot-A/
 cmdline=/cmdline.txt
 
+# Uncomment this to enable GPIO support for RPI-RF-MOD/HM-MOD-RPI-PCB
+#enable_uart=1
+#dtparam=i2c_arm=on
+#dtoverlay=miniuart-bt
+#dtoverlay=rpi-rf-mod
+
 [cm4]
 # Enable host mode on the 2711 built-in XHCI USB controller.
 # This line should be removed if the legacy DWC2 controller is required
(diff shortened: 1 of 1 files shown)


jens-maus:
@agners @sairon Can you please consider adding this minor PR to the upcoming 11.5 release so that users can more easily enable support for the HomeMatic RF modules on a RaspberryPi5?

agners (approved):

buildroot-external/board/raspberrypi/rpi5-64/config.txt
```diff
@@ -43,6 +43,12 @@ arm_boost=1
 os_prefix=slot-A/
 cmdline=/cmdline.txt
 
+# Uncomment this to enable GPIO support for RPI-RF-MOD/HM-MOD-RPI-PCB
+#enable_uart=1
+#dtparam=i2c_arm=on
+#dtoverlay=miniuart-bt
```
sairon, line 49: @jens-maus Is this still needed on RPi5? AFAIK it has a dedicated UART port now - but is it still multiplexing it with BT pins?

jens-maus: @sairon I just tested if `dtoverlay=miniuart-bt` is still required for the RaspberryPi5 image and in fact, it is not. Thus, removed that line from the comment. So please merge accordingly.

sairon (approved):
Thanks @jens-maus for the update!

buildroot-external/board/raspberrypi/rpi5-64/config.txt
agners: @jens-maus for a different use case, I've just used `enable_uart=1` on Raspberry Pi 5, but this did not give me any UART :thinking: Did you had to do something additional to get this going on Raspberry Pi 5?

PR #3090, opened 2024-01-19
Dockerfile: migrate to trusted.gpg.d for docker repo keys

darkxst:
Using apt-key for managing Debian repo keys has been deprecated for a while now. While it currently works ok, seems to be broken in bookworm. This migrates to storing keys in trusted.gpg.d as recommended by the warning messages when using apt-key. This method is still compatible with bullseye.


sairon (approved):
Nice catch, thank you!

PR #3091, opened 2024-01-19
Fix build_haos_builder step to fetch the correct version

tompipe:
The ```build_haos_builder``` step attempts to cache the builder to the docker repo using the value of ```steps.version_main.outputs.version_main```

https://github.com/home-assistant/operating-system/blob/c210a1a52af683b093be2c278e4c6028561d69bf/.github/workflows/build.yaml#L146-L155

However, the ```version_main``` step doesn't exist, the correct id for this step is ```version```

https://github.com/home-assistant/operating-system/blob/c210a1a52af683b093be2c278e4c6028561d69bf/.github/workflows/build.yaml#L82-L93

This is causing the builder image to not be cached correctly:

image

image

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -150,8 +150,8 @@ jobs:
           context: .
           file: Dockerfile
           tags: ghcr.io/${{ github.repository_owner }}/haos-builder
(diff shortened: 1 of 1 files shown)


sairon (approved):
Thanks, it's an obvious mistake that got introduced by some older refactoring.

PR #3098, opened 2024-01-22
Fix ARCH/SRCARCH for aarch64 in linux-check-dotconfig

sairon:
Wrong arch for arm64 produces few false check results resulting from some symbols in arm64 tree not being loaded. We can use raw $(ARCH) variable for arm, only need to translate aarch64 -> arm64.

### DIFF
--- a/buildroot-external/external.mk
+++ b/buildroot-external/external.mk
@@ -3,8 +3,8 @@ include $(sort $(wildcard $(BR2_EXTERNAL_HASSOS_PATH)/package/*/*.mk))
 .PHONY: linux-check-dotconfig
 linux-check-dotconfig: linux-check-configuration-done
 	CC=$(TARGET_CC) LD=$(TARGET_LD) srctree=$(LINUX_SRCDIR) \
-	ARCH=$(if $(BR2_x86_64),x86,$(if $(BR2_arm)$(BR2_aarch64),arm,$(ARCH))) \
-	SRCARCH=$(if $(BR2_x86_64),x86,$(if $(BR2_arm)$(BR2_aarch64),arm,$(ARCH))) \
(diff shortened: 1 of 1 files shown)


buildroot-external/external.mk
```diff
@@ -3,8 +3,8 @@ include $(sort $(wildcard $(BR2_EXTERNAL_HASSOS_PATH)/package/*/*.mk))
 .PHONY: linux-check-dotconfig
 linux-check-dotconfig: linux-check-configuration-done
 	CC=$(TARGET_CC) LD=$(TARGET_LD) srctree=$(LINUX_SRCDIR) \
-	ARCH=$(if $(BR2_x86_64),x86,$(if $(BR2_arm)$(BR2_aarch64),arm,$(ARCH))) \
-	SRCARCH=$(if $(BR2_x86_64),x86,$(if $(BR2_arm)$(BR2_aarch64),arm,$(ARCH))) \
+	ARCH=$(if $(BR2_x86_64),x86,$(if $(BR2_aarch64),arm64,$(ARCH))) \
```
agners, line 6: Nice catch! Shouldn't x86 also be `x86_64`? :thinking: 

```suggestion
	ARCH=$(if $(BR2_x86_64),x86_64,$(if $(BR2_aarch64),arm64,$(ARCH))) \
```

sairon: Nope, it is used in path to top-level arch Kconfig, x86 is in `arch/x86`.

sairon: If we were building for i386, it should have different `ARCH` and `SRCARCH`, to set same values as Linux's Makefile does:
```
ifeq ($(ARCH),i386)
        SRCARCH := x86
endif
```
But I'll rather just ignore that here :innocent:

agners: Right! :+1:

agners (approved):

PR #3099, opened 2024-01-23
add missing ARMv8 32bit kernel support configs

jens-maus:
This PR adds some missing kernel config options for the `generic-aarch64` target which are necessary to introduce compatibility for partly deprecated ARM 32bit instruction sets (mainly CP15 instructions) which are however required because some third-party addons (e.g. like RaspberryMatic addon) come with 32bit applications which might still require these instruction sets to be available. Please see #1323 for a similar PR and more information on that topic, thought.


sairon (approved):
Looks good on my end, thanks!

PR #3115, opened 2024-01-26
uboot: Use rockchip combined SPL images

darkxst:
Since Uboot 2023.10 all Rockchip boards generate a combined `u-boot-rockchip.bin` image that contains the loader, SPL and if required a FIT image containing TF-A/OPTEE binaries.

https://github.com/u-boot/u-boot/blob/e7f9e5eb584dd0b5d1b1ff82fe607d6da9940cc6/arch/arm/dts/rockchip-u-boot.dtsi#L15

Simply installed with:
```
sudo dd if=u-boot-rockchip.bin of=/dev/sda seek=64
```

Use this new combined image for Odroid-M1 and Green boards.

Tinker board would currently generate a broken `u-boot-rockchip.bin`  so not included here. Due to the modified entry point set with `CONFIG_SYS_MMCSD_RAW_MODE_U_BOOT_SECTOR` it probably needs to set the following setting to sync up with that entry point, otherwise it wont boot. I don't have a tinker board to test this though.
```
CONFIG_SPL_PAD_TO=0x3f8000
```


agners:
Cool, less complexity on our end :+1: 

> Tinker board would currently generate a broken `u-boot-rockchip.bin` so not included here. Due to the modified entry point set with `CONFIG_SYS_MMCSD_RAW_MODE_U_BOOT_SECTOR` it probably needs to set the following setting to sync up with that entry point, otherwise it wont boot. I don't have a tinker board to test this though.

I do have a Tinkerboard here. Can you create a separate PR for what you believe should work? I'll test it on my end.

buildroot-external/board/nabucasa/green/hassos-hook.sh
```diff
@@ -14,8 +14,7 @@ function hassos_pre_image() {
     # SPL
     create_spl_image
 
-    dd if="${BINARIES_DIR}/idbloader.img" of="${SPL_IMG}" conv=notrunc bs=512 seek=64
-    dd if="${BINARIES_DIR}/u-boot.itb" of="${SPL_IMG}" conv=notrunc bs=512 seek=16384
+    dd if="${BINARIES_DIR}/u-boot-rockchip.bin" of="${SPL_IMG}" conv=notrunc bs=512 seek=64
```
agners, line 17: Currently we have an extra partition marking/reserving the location of `u-boot.itb`. The second dd command used to reflect exactly this offset. But I guess the offset really is inherently tied to the  idbloader, so having them as one binary makes sense. Just wanted to note this.

agners (approved):

buildroot-external/board/nabucasa/green/hassos-hook.sh
darkxst: odroid-m1 hardcodes a partition for uboot but other boards dont? Anyway with default uboot config, it still ends up in the same place.

agners: Yes, ODROID-M1 is a bit special/the first one to do this.

Normally, partitions start at 1MB/2MB offset. Since U-Boot is at offset 8MB, it feels a bit weird having it there without the partition table knowing. So I've decided to give it an entry in the partition table. Since we use IDs everywhere it is safe to use a different layout for a particular board.

See: https://github.com/home-assistant/operating-system/blob/11.4/buildroot-external/scripts/hdd-image.sh#L165

This actually brings up a good point: We should definitely do this for ODROID-M1S too. And probably should have done it for Green too, but it feels a bit weird to change that now, since it would mean we'd end up having installations with different partition tables :thinking:

darkxst: I will make sure to do this on M1S then. 

I have noticed the offsets are a bit weird even still, with boot partition starting at 16M (from meta) +1M offset :confused:. But can't really change that now for existing boards!

PR #3118, opened 2024-01-27
Use rockchip combined SPL images for Tinkerboard

darkxst:
Following on from #3115 

Tinker uses a modified u-boot entry point to suit its partition layout, so needs additional config to correctly generate the combined `u-boot-rockchip.bin` file matching this entry point.

`SPL Pad Offset -> [0x2000 - 0x40 ] * 0x200 (sector size) =0x3f8000`

Thus the resulting settings are:
```
CONFIG_SYS_MMCSD_RAW_MODE_U_BOOT_SECTOR=0x2000
CONFIG_SPL_PAD_TO=0x3f8000
```
Also move the existing setting out of the patch and into uboot.config so both settings are together in the same place.

Confirmed resulting offset is correct per below, but please boot test this before merging.
```
$ binwalk spl.img 

DECIMAL       HEXADECIMAL     DESCRIPTION
--------------------------------------------------------------------------------
98324         0x18014         Flattened device tree, size: 5707 bytes, version: 17
4194304       0x400000        uImage header, header size: 64 bytes, header CRC: 0x2DCD38E, created: 2024-01-27 00:39:57, image size: 501060 bytes, Data Address: 0x1000000, Entry Point: 0x1000000, data CRC: 0xC97AED3C, OS: Firmware, CPU: ARM, image type: Firmware Image, compression type: none, image name: "U-Boot 2024.01 for tinker_rk3288"
```


agners:
I've tested a build on my ASUS Tinkerboard S, and it came up just fine. Unfortunately, my ASUS Tinkerboard non-S died a couple of weeks ago, so I can't test that. But it really shouldn't matter.

agners (approved):

PR #3120, opened 2024-01-28
Remove folder with only Amlogic patches from Odroid M1 defconfig

darkxst:
All of the patches in ./board/hardkernel/patches are specific to Amlogic SoC's. Drop this patch path when building Rockchip based Odroid-m1 boards.


buildroot-external/configs/odroid_m1_defconfig
```diff
@@ -137,7 +137,7 @@ BR2_TARGET_UBOOT_CUSTOM_GIT=y
 BR2_TARGET_UBOOT_CUSTOM_REPO_URL="https://github.com/u-boot/u-boot.git"
 BR2_TARGET_UBOOT_CUSTOM_REPO_VERSION="v2024.01"
 BR2_TARGET_UBOOT_BOARD_DEFCONFIG="odroid-m1-rk3568"
-BR2_TARGET_UBOOT_CONFIG_FRAGMENT_FILES="$(BR2_EXTERNAL_HASSOS_PATH)/bootloader/uboot.config"
+BR2_TARGET_UBOOT_CONFIG_FRAGMENT_FILES="$(BR2_EXTERNAL_HASSOS_PATH)/bootloader/uboot.config $(BR2_EXTERNAL_HASSOS_PATH)/board/hardkernel/odroid-m1/uboot.config"
```
agners, line 140: Adding it now disables/enables some configs which I am not sure we actually want/need to change. I think rather dropping that config would probably be the better option :thinking: 

FWIW, I just noticed this too: I am seeing Ethernet issues with U-Boot 2024.01 here, and work on a patch which disables Ethernet altogether. For this I need a boards specific config fragment, that is where I've noticed it :sweat_smile:

I suggest we leave this config as is in this PR, and I'll fix it up with the upcoming Ethernet fix.

buildroot-external/configs/odroid_m1_defconfig
```diff
@@ -5,7 +5,7 @@ BR2_CCACHE=y
 BR2_CCACHE_DIR="/cache/cc"
 BR2_OPTIMIZE_2=y
 BR2_ENABLE_LTO=y
-BR2_GLOBAL_PATCH_DIR="$(BR2_EXTERNAL_HASSOS_PATH)/patches $(BR2_EXTERNAL_HASSOS_PATH)/board/hardkernel/patches $(BR2_EXTERNAL_HASSOS_PATH)/board/hardkernel/odroid-m1/patches"
+BR2_GLOBAL_PATCH_DIR="$(BR2_EXTERNAL_HASSOS_PATH)/patches $(BR2_EXTERNAL_HASSOS_PATH)/board/hardkernel/odroid-m1/patches"
```
agners, line 8: There are patches in this directory, but I see these are really not needed for M1. Nice catch!

agners (changes requested):

buildroot-external/configs/odroid_m1_defconfig
darkxst: ok I have dropped uboot fragment patch from PR

buildroot-external/configs/odroid_m1_defconfig
sairon: On the other hand, we normally pick all the patches from top to the bottom, so this will be an exception now. Maybe we should consider using a different folder for platform-specific patchsets?

darkxst: Most others dont even have $(vendor)/patches directory, except for Raspberry!


All of the patches in `hardkernel/patches/linux`  seem to be board specific DT patches for C2/C4/N2, there is nothing there I can see that is platform specific, perhaps they could just be moved to relevant board folders? I guess the uboot patch probably applies to any amlogic board though.

I suppose there would be a case for a shared `rockchip-patches` folder given Green/M1/M1S likely have some common patches, for example the RNG patcheset which seems to have died upstream. And on that note perhaps also a shared `kernel-rockchip.config` given they both identical

darkxst: Sure, I will shuffle some things around.

sairon (approved):
I'll merge this PR as-is. Let's leave eventual refactoring for later, so it wouldn't get in conflict with Linux 6.6 update.

PR #3164, opened 2024-02-10
Update grub.cfg to blank console

patrickbohren:
To safe energy an allow usage of Surface tablets with integrated screen, switch off the backlight after 120sec

### DIFF
--- a/buildroot-external/board/pc/grub.cfg
+++ b/buildroot-external/board/pc/grub.cfg
@@ -60,7 +60,7 @@ fi
 
 save_env A_TRY A_OK B_TRY B_OK ORDER MACHINE_ID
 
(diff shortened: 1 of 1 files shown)


sairon (changes requested):
IMHO your use case does not justify such change for all deployments, moreover specific only to a single platform. You're always free to edit `cmdline.txt` in your boot partition to add anything extra, which is more appropriate in this case.

Leaving this open though for @agners to give him another look at it.

agners:
From what I can tell the Linux default is 600s. I'd rather not overwrite that by default. I agree with @sairon, you can edit cmdline.txt on your particular platform (which will be preserved on update) to customize your kernel command line.

sairon:
OK, since we have an agreement here, I'm closing the PR. Thanks for the effort anyway, @patrickbohren 🙏

PR #3165, opened 2024-02-11
Factor out common Rockchip config

darkxst:
Further cleanup for common Rockchip parts before introducing the Odroid-m1s board.

- Move to common kernel-arm64-rockchip.config for all 64-bit Rockchip boards, this config is very comprehensive and likely would work for any other 64-bit Rockchip boards added in the future. Clear the CONFIG_RTC_DRV_RK808 option for Green, this was the only difference between the two original configs. Named as such to make it clear its not good for 32-bit.

- Move RNG patchset to common patch folder, technically this is specific to rk356x SoCs, which for now covers Green and both Odroid-M1 variants.


sairon (approved):
LGTM, thanks! My only concern is that the top-level `patches-rockchip` folder is introducing a new convention that hasn't been used until now, however I don't have any better idea how to do that to reach the same goal. Other option would be applying the patches always, since they're basically not affecting anything else if the config option is not enabled. And the best option in mid-term would be pushing the RNG driver upstream :)

darkxst:
> And the best option in mid-term would be pushing the RNG driver upstream :)

RNG patches upstream seem to have been abandoned after v2, I suppose they could be picked up again though.

https://patchwork.kernel.org/project/linux-rockchip/patch/20221128184718.1963353-2-aurelien@aurel32.net/

PR #3169, opened 2024-02-12
Update issue template with better links to logs, add CLI instructions

sairon:
Legacy supervisor_logs target (which is currently kind of broken) replaced with standard logs with provider specified. Added instructions how to get logs in HA CLI.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -82,7 +82,7 @@ body:
       label: Anything in the Supervisor logs that might be useful for us?
       description: >
         Supervisor Logs can be found in [Settings -> System -> Logs](https://my.home-assistant.io/redirect/logs/?provider=supervisor)
-        then choose `Supervisor` in the top right. Alternatively enter `supervisor logs` in the Home Assistant CLI.
+        then choose `Supervisor` in the top right. Alternatively enter `ha supervisor logs` in the Home Assistant CLI.
 
         [![Open your Home Assistant instance and show your Supervisor system logs.](https://my.home-assistant.io/badges/supervisor_logs.svg)](https://my.home-assistant.io/redirect/logs/?provider=supervisor)
       render: txt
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -81,19 +81,19 @@ body:
     attributes:
       label: Anything in the Supervisor logs that might be useful for us?
       description: >
-        Supervisor Logs can be found in [Settings -> System -> Logs](https://my.home-assistant.io/redirect/logs/)
-        then choose `Supervisor` in the top right.
+        Supervisor Logs can be found in [Settings -> System -> Logs](https://my.home-assistant.io/redirect/logs/?provider=supervisor)
+        then choose `Supervisor` in the top right. Alternatively enter `supervisor logs` in the Home Assistant CLI.
```
agners, line 85: When running from SSH/web terminal add-on, you have to use `ha` in front. I'd prefer if we always prepend the commands with `ha`.

```suggestion
        then choose `Supervisor` in the top right. Alternatively enter `ha supervisor logs` in the Home Assistant CLI.
```

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -81,19 +81,19 @@ body:
...
   - type: textarea
     validations:
       required: true
     attributes:
       label: Anything in the Host logs that might be useful for us?
       description: >
-        Host Logs can be found in [Settings -> System -> Logs](https://my.home-assistant.io/redirect/logs/)
-        then choose `Host` in the top right.
+        Host Logs can be found in [Settings -> System -> Logs](https://my.home-assistant.io/redirect/logs/?provider=host)
+        then choose `Host` in the top right. Alternatively enter `host logs` in the Home Assistant CLI.
```
agners, line 96:
```suggestion
        then choose `Host` in the top right. Alternatively enter `ha host logs` in the Home Assistant CLI.
```

agners (changes requested):

.github/ISSUE_TEMPLATE/bug_report.yml
sairon: I tried to address it by explicitly saying "from Home Assistant CLI". But you are right it may still be confusing to beginners, and using `ha` in HA CLI just raises a warning, so it doesn't do much harm.

agners (approved):

PR #3175, opened 2024-02-14
Odroid UMS gadgets

darkxst:
UMS gadget patches split out from initial Odroid M1S PR. As discussed in that PR needs some further work, but creating draft PR for visibility and discussion.


darkxst:
Hardkernels USB UMS tool is just a custom u-boot image with UMS enabled. 
https://docs.u-boot.org/en/latest/usage/cmd/ums.html

Perhaps we could have a recovery command in HAOS that reboots into u-boot UMS mode, using `fw_setenv` to trigger UMS instead of the default boot flow?

PR #3187, opened 2024-02-20
Clarify M1S flashing and recovery docs

sairon:
Remove the mentions of petitboot, as it's not normally used on M1S. Document possibility to use SD card boot and alternative method of reflashing the eMMC from an OS running from an SD card.

### DIFF
--- a/Documentation/boards/hardkernel/odroid-m1s.md
+++ b/Documentation/boards/hardkernel/odroid-m1s.md
@@ -4,19 +4,26 @@ Home Assistant OS 12 and newer support the ODROID-M1S board.
 
 ## SD-card
 
-SD-card boot is supported only if you have erased Petitboot from eMMC.
+ODROID-M1S can boot HAOS directly from an SD card, as it has higher priority than the system on the eMMC. Simply flash the image to the SD card using your favorite tool and insert it to the micro SD slot on the board. This works even when the eMMC is wiped, or when it contains the factory-default U-Boot SPL loader, which is still able to load U-Boot provided in the HAOS image. In the second case, however, if the SD card fails to probe (e.g. due to a hardware failure), the system on the eMMC may be booted instead of HAOS.
 
 ## eMMC
 
-To install to eMMC via on-board Petitboot bootloader:
+HAOS can be installed directly to the eMMC using a special boot image, to do that:
 
 1. Download [`ODROID-M1S_EMMC2UMS.img`][1]
 2. Use balenaEtcher or another tool to flash the UMS utility onto an SD card.
 3. Plug-in that SD card to your ODROID-M1S and boot it. Connect your PC to the Micro USB OTG port.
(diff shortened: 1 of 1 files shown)


Documentation/boards/hardkernel/odroid-m1s.md
```diff
@@ -4,27 +4,34 @@ Home Assistant OS 12 and newer support the ODROID-M1S board.
...
 4. The eMMC will show as a drive on your PC and you can directly flash the HAOS image with balenaEther.
 
-Installing HAOS erases Petitboot, if you wish to return to any official Hardkernel images then short the maskRom pads near
-the 40pin connector while booting from a [Hardkernel recovery image][2].
+Installing HAOS replaces the firmware and SPL on the eMMC with the mainline version provided by HAOS. As a result, it is not possible to use the SD card with the EMMC2UMS image anymore, because the mainline SPL is not compatible with U-Boot in the EMMC2UMS image at this time (February 2024). This does not pose any problem for standard use, just makes it more complicated in case you want to return to the Hardkernel-provided OS.
+
+A reliable way of reflashing the eMMC in this case is to use HAOS booted from an SD card. To do that, insert the SD card with HAOS to the micro SD slot and plug the board in. Once the device boots to the HA CLI, enter `login` to enter the root shell and use `curl` to download an image and `dd` it to the eMMC block device:
+
+```sh
+curl https://dn.odroid.com/RK3566/ODROID-M1S/Installer/ODROID-M1S_EMMC2UMS.img | dd of=/dev/mmcblk0
```
agners, line 23: Hm, I see, this works since the image is only 3MB and would not disturb any of the mounted partitions.

agners (approved):

Documentation/boards/hardkernel/odroid-m1s.md
sairon: Actually none of the partitions on the eMMC should be mounted when you boot from an SD card. But you're right, you can actually modify the boot segment even from a running OS, since it should never be mounted as a partition and is R/W.

darkxst (approved):
This looks good to me.

This recovery method is certainly better than using maskrom pins ;)

Documentation/boards/hardkernel/odroid-m1s.md
darkxst: > Actually none of the partitions on the eMMC should be mounted when you boot from an SD card.

Not relevant to these docs as not touching the partitions, but is that actually true? I seem to recall issues when booting HAOS from an sd card when HAOS is also installed on eMMC. HAOS has trouble choosing the correct data partition. Thus could end up with the eMMC data partition mounted even though rest of system booted from SD card.

sairon: Hmm, good point, will check what it does in that case. Let's hope no one sane is going to do that though :sweat_smile:

PR #3241, opened 2024-03-12
Update Buildroot to 2024.02

sairon:
Update Buildroot to latest LTS release. Builds fine, basic testing also found no issues. In general some library changes pulled in some more dependencies and increased the bloat a little (like udisks or nfs-utils) but it's still bearable. For details of changes required for the update see description of individual commits.


buildroot-external/configs/generic_aarch64_defconfig
```diff
@@ -3,7 +3,6 @@ BR2_TOOLCHAIN_BUILDROOT_CXX=y
 BR2_DL_DIR="/cache/dl"
 BR2_CCACHE=y
 BR2_CCACHE_DIR="/cache/cc"
-BR2_OPTIMIZE_2=y
```
agners, line 6: I see this is the new default, hence config is no longer needed :+1:

agners (approved):
Generally LGTM. There are some package checks failing because upstream tags are missing. From my point of view we can also disable that particular check, if possible.

buildroot-external/configs/generic_aarch64_defconfig
sairon: Yes, basically all the changes here shouldn't change anything, they just fell out of defconfig because they became dependencies or defaults.

sairon: > Generally LGTM. There are some package checks failing because upstream tags are missing. From my point of view we can also disable that particular check, if possible.

Okay, let's do that for now, and I'll make a note to re-enable it later - it might be actually useful for us too.

PR #3248, opened 2024-03-14
kernel: enable MPTCP support

matttbe:
First, thank you for all the work around the Home-Assistant, it is a really useful project that we can control

The official description says:

> Multipath TCP (MPTCP) connections send and receive data over multiple
> subflows in order to utilize multiple network paths. Each subflow uses
> the TCP protocol, and TCP options carry header information for MPTCP.

Thanks to MPTCP, being able to use multiple paths in parallel or simultaneously brings new use-cases:

- Seamless handovers: switching from one path to another while preserving established connections -- Apple is using it for this reason since 2013.

- Best network selection: using the "best" available path (latency, losses, cost, bandwidth) -- one path can be used as a "backup" one.

- Network aggregation: using multiple paths at the same time to have a higher throughput -- e.g. to combine a fixed an mobile network to send files faster.

For example, for HA, it is possible to keep a SSH connection alive when switching from one network to another (e.g. while travelling).

To be able to use MPTCP, both ends need to support it. An application has to request it, by creating an MPTCP socket instead of a TCP one. The rest in unchanged. An alternative is to use `mptcpize` tool, which relies on `LD_PRELOAD` to create an MPTCP socket instead of a TCP one.

Note that a MPTCP-enabled server continues to accept regular TCP connections that do not use the Multipath TCP extension without any performance impact. When a connection request is received, and is linked to a listening socket with MPTCP support, the kernel will simply check if MPTCP options are present. If not, the accepted socket will be a "plain" TCP one, with the same impact as before.

To use multiple paths at the same time, additional IP addresses need to be configured, e.g. via the `ip` tool (IPRoute2).

MPTCP in the kernel is light, and enabled in most main Linux distributions (Debian, Ubuntu, RedHat, Fedora, etc.), but in more specific ones like Raspbian. It is available in the Linux kernel since v5.6.

It would be nice to have it supported in hassio, to let apps using it.


buildroot-external/kernel/v6.1.y/hassos.config
```diff
@@ -196,3 +196,6 @@ CONFIG_NLS_CODEPAGE_437=y
 CONFIG_NLS_ASCII=y
 CONFIG_NLS_ISO8859_1=y
 CONFIG_NLS_UTF8=y
+
+CONFIG_MPTCP=y
+CONFIG_MPTCP_IPV6=y
```
agners, line 201: It doesn't really matter all that much, but can you move it up a bit to the other network/IP related configs?

agners (commented):
Thanks for your contribution. I agree, enabling MPTCP seems sensible and useful, given it is enabled in most general purpose Linux distros too.

buildroot-external/kernel/v6.1.y/hassos.config
matttbe: Sure, good idea. I just put them after IPv6, and before Netfilter options, similar to what we have in the Linux kernel.

agners (approved):

PR #3270, opened 2024-03-27
Linux: Update kernel to 6.6.23

sairon:
* https://cdn.kernel.org/pub/linux/kernel/v6.x/ChangeLog-6.6.23


agners (approved):

sairon:
It contains a potential fix for the issue we saw with RTL8821CU, so #3225 may no longer be necessary:
```
commit 40243da588a0f337b8e5460426192e9ed74f2a48
Author: Bitterblue Smith <rtl8821cerfe2@gmail.com>
Date:   Fri Mar 1 00:32:45 2024 +0200

    wifi: rtw88: 8821cu: Fix firmware upload fail
    
    [ Upstream commit 41a7acb7dde8395f52a707bbba7712a898dfafb0 ]
    
    RTL8822CU, RTL8822BU, and RTL8821CU need an extra register write after
    reading and writing certain addresses.
    
    Without this, the firmware upload fails approximately more than 50% of
    the time.
    
    Tested with RTL8811CU (Tenda U9 V2.0) which is the same as RTL8821CU
    but without Bluetooth.
    
    Fixes: a82dfd33d123 ("wifi: rtw88: Add common USB chip support")
    Signed-off-by: Bitterblue Smith <rtl8821cerfe2@gmail.com>
    Acked-by: Ping-Ke Shih <pkshih@realtek.com>
    Signed-off-by: Kalle Valo <kvalo@kernel.org>
    Link: https://msgid.link/f12ed39d-28e8-4b8b-8d22-447bcf295afc@gmail.com
    Signed-off-by: Sasha Levin <sashal@kernel.org>
```

agners:
> It contains a potential fix for the issue we saw with RTL8821CU, so #3225 may no longer be necessary:

Hm, interesting! I guess we could revert #3225 then again.

Or maybe wait until this trickles down to the Raspberry Pi kernel and then revert? :thinking:

sairon:
We always used the OOT driver for RPi because the in-tree it was added in 6.2 or so, so even if we revert it (partially, as we don't want to remove it from RPi2 and Yellow), we'll still use the OOT on RPi platforms. I'll test it for a couple of hours and see.

Issue #3280, opened 2024-03-31
USB SSD ADATA SD600Q fails to boot when connected to USB3 port on RPI4


agglrx:
### Describe the issue you are experiencing

Booting from the ADATA SD600Q SSD when connected to a USB 3.0 port on a Raspberry Pi 4 fails with `uas` errors. 

![uas-errors](https://github.com/home-assistant/operating-system/assets/32034497/46522d72-8294-4247-be57-9075564ae7a1)

Adding `125f:a88a:u` to `usb-storage.quirks` in cmdline.txt resolves the issue.

### What operating system image do you use?

rpi4-64  (Raspberry Pi 4/400 64-bit OS)

### What version of Home Assistant Operating System is installed?

12.1

### Did you upgrade the Operating System.

Yes

### Steps to reproduce the issue

1. Connect ADATA SD600Q to USB 3.0 port of RPI4
2. Boot 

### Anything in the Supervisor logs that might be useful for us?

```txt
see picture
```

### Anything in the Host logs that might be useful for us?

```txt
see picture
```

### System information

n/a

PR #3281, opened 2024-03-31
Disable UAS via usb-storage.quirks on RPi for ADATA SD600Q

agglrx:
Booting from a ADATA SD600Q fails when connected to a USB 3.0 port on RPi4. Adding it to the quirks list resolves the issue.

Fixes https://github.com/home-assistant/operating-system/issues/3280


sairon (approved):
Thank you!

PR #3285, opened 2024-04-03
Ignore non-existing CP15 barrier instructions emulation

agners:
CP15 barrier instruction emulation only exists on arm64 architecture. Avoid sysctl writing an error to the journal when the setting doesn't exist by prepending a dash.


sairon (approved):

jens-maus:
@agners Sorry, to bring back this already merged PR, but after I applied this change (add '-' to mitigate warning) my users are reporting that the `abi.cp15_barrier` setting was not set effective, thus not applied. So are you really sure that adding `-` to the config items in `sysctl.conf` is a valid approach? (cf. https://github.com/jens-maus/RaspberryMatic/commit/7bc55f22cccb4ac4d076fa59d3203a9dc6e76240)

agners:
@jens-maus I remember I tested it, but was now a bit perplex when i realized that Busybox doesn't support that syntax (see https://github.com/mirror/busybox/blob/master/procps/sysctl.c).

However, it seems that at least HAOS uses the full sysctl, so for HAOS, this works.

PR #3297, opened 2024-04-09
Update Raspberry Pi Kernel to v6.6.20

sairon:
Update Raspberry Pi kernel to latest version used by RPi OS (v6.6.20 from git commit 6f16847710cc0502450788b9f12f0a14d3429668). This allows for unifying linux patch directory again and using common config fragments. Only minor adjustments in Yellow DTS were necessary.


buildroot-external/patches/linux/6.6.25/0001-ipv6-add-option-to-explicitly-enable-reachability-te.patch
```diff
@@ -1,52 +0,0 @@
-From 76591e4075194cf717dc085b8285912f706bcd46 Mon Sep 17 00:00:00 2001
-From: Stefan Agner <stefan@agner.ch>
-Date: Tue, 28 Mar 2023 12:02:10 +0200
-Subject: [PATCH] ipv6: add option to explicitly enable reachability test
```
agners, line 4: This patch seems to be missing now? :thinking:

agners (approved):

buildroot-external/patches/linux/6.6.25/0001-ipv6-add-option-to-explicitly-enable-reachability-te.patch
sairon: Not really, now it's been replaced by the patch all versions in `buildroot-external/patches/linux/0001-ipv6-add-option-to-explicitly-enable-reachability-te.patch`

Issue #3298, opened 2024-04-10
Raspberry CM3+


fabfix:
### Describe the issue you are experiencing

Boot 32bit only, 64bit doesn't boot!

### What operating system image do you use?

rpi3-64 (Raspberry Pi 3 64-bit OS)

### What version of Home Assistant Operating System is installed?

12.1

### Did you upgrade the Operating System.

No

### Steps to reproduce the issue

flash a CM3+ with v12.1 64bit and reset. it doesn't boot.
flash a CM3+ with v12.1 32bit and reset. It boot.

### Anything in the Supervisor logs that might be useful for us?

```txt
NA
```

### Anything in the Host logs that might be useful for us?

```txt
NA
```

### System information

CM3+ 1GB RAM and storage 32GB

### Additional information

NA

PR #3301, opened 2024-04-11
Update Raspberry Pi 3 device tree list

agners:
Make sure all Raspberry Pi devices using the BCM2837 SoC (or the SiP version RP3A0 of it) are deployed. This especially makes sure that we only deploy the downstream device trees (named bcm2710*) and deploy device trees for the CM3 as well as the Zero 2 and Zero 2 W consistently for 32-bit and 64-bit.

Fixes: #3298


sairon (changes requested):
This needs rebase after RPi kernel bump to 6.6 :grimacing:

sairon (approved):
Looks good. Seems that the `broadcom/bcm2837-rpi-3-b` wasn't needed, it's the upstream DT with some downstream parts missing. Unless it was picked by FW in some cases (which IMHO doesn't make much sense) it should be safe to remove it.

agners:
> Looks good. Seems that the `broadcom/bcm2837-rpi-3-b` wasn't needed, it's the upstream DT with some downstream parts missing. Unless it was picked by FW in some cases (which IMHO doesn't make much sense) it should be safe to remove it.

I've just checked on a Raspberry Pi 3 (non-plus), and it doesn't seem to get picked today (`/proc/device-tree/wifi-pwrseq` which is specific to the upstream device tree is not present).

fabfix:
Hi @agners, thank you soo much for the fast resolution! 
Do you know in witch release this fix will available?

PR #3317, opened 2024-04-17
Update instructions for ODROID M1 Petitboot update

salim-b:
Refer to latest versions and fix minor details.


sairon (approved):
Thanks, did just a minor adjustment there.

salim-b:
> Thanks, did just a minor adjustment there.

The two trailing whitespaces after the first sentence in point 7 were actually intentional – it's a [hard line break](https://github.github.com/gfm/#hard-line-breaks) in Markdown/GFM. But it's fine either way, I guess.

sairon:
Oh, TIL! Let's use the other syntax then, some editors remove trailing whitespace automatically so the line break might be lost unintentionally there.

PR #3329, opened 2024-04-26
Enable KVM for running tests in GH Actions

sairon:
It was not possible on free runners previously, but it should be now: https://github.com/actions/runner-images/discussions/7191#discussioncomment-9018826


agners (approved):
That is excellent news! LGTM, thanks!

PR #3384, opened 2024-05-20
Linux: Update kernel to 6.6.31

sairon:
* https://cdn.kernel.org/pub/linux/kernel/v6.x/ChangeLog-6.6.31


jens-maus:
@sairon Please note, that the [0002-Revert-USB-core-changes-causing-issues-with-Z-Wave.m.patch](https://github.com/home-assistant/operating-system/blob/dev/buildroot-external/patches/linux/0002-Revert-USB-core-changes-causing-issues-with-Z-Wave.m.patch) under "patches/linux" is currently not compatible with 6.6.31 since there are changes in 6.6.31 which contflicts with the patch. In addition, modifying the patch in that location is currently not possible since the 6.6 kernel for the RaspberryPi platform is currently not on 6.6.31 thus the patchset can then afterwards not be applied to the RaspberryPi kernel anymore.

sairon:
@jens-maus I have noticed in the local build already, but thanks for the quick heads-up! :)

> In addition, modifying the patch in that location is currently not possible since e.g. the 6.6 kernel for the RaspberryPi platform is currently not on 6.6.31 thus the patchset can then afterwards not be applied to the RaspberryPi kernel anymore.

Right, but it's not a big issue - we had the same situation in the past before we updated RPi kernel to 6.6.y.

sairon:
The diff is so subtle that Git thinks there are no changes in the "moved" patch, yet there is, most notably this:
```diff
@@ -400,10 +400,9 @@
 -	}
 -	if (usb_endpoint_maxp(&udev->ep0.desc) == i) {
 -		;	/* Initial ep0 maxpacket guess is right */
--	} else if (((udev->speed == USB_SPEED_FULL ||
+-	} else if ((udev->speed == USB_SPEED_FULL ||
 -				udev->speed == USB_SPEED_HIGH) &&
--			(i == 8 || i == 16 || i == 32 || i == 64)) ||
--			(udev->speed >= USB_SPEED_SUPER && i > 0)) {
+-			(i == 8 || i == 16 || i == 32 || i == 64)) {
 -		/* Initial guess is wrong; use the descriptor's value */
 +	if ((udev->speed >= USB_SPEED_SUPER) &&
 +			(le16_to_cpu(udev->descriptor.bcdUSB) < 0x0300)) {
```

agners (approved):

PR #3388, opened 2024-05-22
Use Genimage for declarative image layout

sairon:


buildroot-external/genimage/hdimage-mbr.cfg
```diff
@@ -0,0 +1,6 @@
+hdimage {
+	partition-table-type = "mbr"
+	align = 1M
+	disk-signature = 0x48617373
```
agners, line 4: I guess to make it binary same as we have today this makes sense. But maybe it would be better to be random? :thinking:

agners (commented):
The whole PR looks very good :star_struck: 

There are probably simplification opportunities, especially in the SPL/boot part of things.

But transitioning what we have to genimage first is a good approach. We can then still start to optimize/change things in future steps :+1:

buildroot-external/genimage/hdimage-mbr.cfg
sairon: Not sure, we'll have to find out if we can use a different approach from hardcoded partition ID in `mbr-part.rules` and U-Boot scripts (just grep this number in the repo for reference).

buildroot-external/board/hardkernel/odroid-c2/partition-spl-spl.cfg
```diff
@@ -0,0 +1,8 @@
+partition spl {
+	size = ${BOOT_SPL_SIZE}
+	image = "spl.img"
+	in-partition-table = "no"
+	offset = 0
+	holes = {"(440; 512)"}
```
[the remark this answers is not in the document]
sairon: Same as [here](https://github.com/home-assistant/operating-system/pull/3388#discussion_r1611235350).

buildroot-external/board/hardkernel/odroid-n2/partition-spl-spl.cfg
```diff
@@ -0,0 +1,7 @@
+partition spl {
+	size = ${BOOT_SPL_SIZE}
+	image = "spl.img"
+	in-partition-table = "no"
+	offset = 0
+	holes = {"(0; 32k)"}
```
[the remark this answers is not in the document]
sairon: Correct, thanks, it is sufficient for MBR layout. Oversight in copy-paste again I think :)

buildroot-external/board/hardkernel/odroid-xu4/partition-spl-spl.cfg
```diff
@@ -0,0 +1,6 @@
+partition spl {
+	size = ${BOOT_SPL_SIZE}
+	image = "spl.bin"
+	in-partition-table = "no"
+	offset = 32K
```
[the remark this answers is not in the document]
sairon: Absolutely, I also see that XU4 and M1/M1S (where I forgot to stage SPL configs) failed in my test build, so this is probably the cause.

buildroot-external/board/khadas/vim3/partition-spl-spl.cfg
```diff
@@ -0,0 +1,7 @@
+partition spl {
+	size = ${BOOT_SPL_SIZE}
+	image = "spl.img"
+	in-partition-table = "no"
+	offset = 0
+	holes = {"(440; 512)"}  # TODO: why was 444?
```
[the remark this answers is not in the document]
agners: Uh right, I see the "Structure of a modern standard MBR" is different.

> In summary, we can keep the signature part as a writable hole for all boards, like in upstream's [odroidc2 genimage.cfg](https://github.com/home-assistant/buildroot/blob/2024.02.x-haos/board/hardkernel/odroidc2/genimage.cfg#L21).

Yes, agreed! So 440 is good then :+1:

buildroot-external/board/hardkernel/odroid-m1/hassos-hook.sh
```diff
@@ -3,18 +3,12 @@
...
     cp "${BINARIES_DIR}/boot.scr" "${BOOT_DATA}/boot.scr"
     cp "${BINARIES_DIR}"/*.dtb "${BOOT_DATA}/"
 
     cp "${BOARD_DIR}/boot-env.txt" "${BOOT_DATA}/haos-config.txt"
     cp "${BOARD_DIR}/cmdline.txt" "${BOOT_DATA}/cmdline.txt"
-
-    # SPL
-    create_spl_image
-
-    dd if="${BINARIES_DIR}/u-boot-rockchip.bin" of="${SPL_IMG}" conv=notrunc bs=512 seek=64
```
agners, line 17: It seems `image-spl-spl.cfg` and `partition-spl-spl.cfg` are missing for m1/m1s?

buildroot-external/board/hardkernel/odroid-xu4/partition-spl-spl.cfg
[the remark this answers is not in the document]
sairon: Ahh, I see what's wrong, the build failed because I have `spl.bin` here. It should be `spl.img`, but since I ran a non-clean build, I still had the other file from previous tests and didn't notice.

But the hole is not necessary, because 0-32k part of the image is untouched, this SPL is written to the 32k offset.

buildroot-external/board/hardkernel/odroid-m1/hassos-hook.sh
sairon: Yep, noticed it too shortly before you: https://github.com/home-assistant/operating-system/pull/3388#discussion_r1611245763

buildroot-external/genimage/hdimage-mbr.cfg
agners: Ah I see. Ok, yeah then we definitely need to keep it as-is for now!

agners (approved):
This LGTM, I guess we need to make sure that we check and test boot all boards.

sairon:
Okay, I think this is ready for the final review. I ran a [test build](https://github.com/home-assistant/operating-system/actions/runs/9386174970) with modified job definition to compare partition tables and layout of the generated image when an SPL is present. There are couple of differences but IMO none of them should break anything, or they even straighten some things. In summary:

1. All images now have the "first usable sector" pointing to the first sector of the first partition. This shouldn't be a problem, since the previous one (LBA 34) wasn't 1M-aligned and in the cases where the image also contains an SPL, the sectors with the SPL shouldn't be usable. But in general it has no impact on function anyway.
2. Images using the hybrid partition table and SPL (ODROID M1, M1S, Green and Tinker) now have the first partition moved by 2k sectors ahead - i.e. there's no 1M gap anymore. This gap was intentionally preserved [here](https://github.com/home-assistant/operating-system/commit/9ce0766353ddebfb4a361fc1fbce6bced8e2c146) but I don't think we really need to do that, and it will just complicate the config, since an arbitrary and calculated offset would have to be added to the boot partition for those boards.
3. Boards with MBR layout now use the `0x0F` extended type instead of the `0x05`. The type is hardcoded in Genimage (and on the other hand, `sfdisk` always creates extended partitions `0x05` even if you specify `0x0F`). The difference _should_ be the latter uses the CHS addressing, however any modern systems now use LBA, so this difference is just formal.
4. MBR layout doesn't have gap of 1M unused space anymore between the end of the extended partition and the next primary one. There is no reason to have it, that space was just wasted, and doing that just for the sake of having the images identical would complicate things again.
5. Boot partition on all hybrid table boards has changed from [MSR](https://en.wikipedia.org/wiki/Microsoft_Reserved_Partition) to more appropriate [ESP](https://en.wikipedia.org/wiki/EFI_system_partition). I can't find proper reasoning for using this type ([here](https://www.github.com/home-assistant/operating-system/pull/852)), it even [fails to boot](https://www.github.com/home-assistant/operating-system/pull/847#issuecomment-1416212525) on some historic RPi bootloader and it makes it harder to mount and edit the files in the boot partition.

tl;dr - we don't have binary-identical images but functionally there shouldn't be differences. We still should test all the boards (or at least those that have some specifics) - but I have tried on those I have available and things are looking good.

buildroot-external/scripts/rauc.sh
```diff
@@ -50,13 +50,14 @@ function install_rauc_certs() {
...
+        # shellcheck disable=SC1117
         echo -e "/dev/disk/by-partlabel/hassos-bootstate\t0x0000\t${BOOT_ENV_SIZE}" > "${TARGET_DIR}/etc/fw_env.config"
     fi
 
     # Fix MBR
-    if [ "${BOOT_SYS}" == "mbr" ]; then
+    if [ "${PARTITION_TABLE_TYPE}" == "mbr" ]; then
         mkdir -p "${TARGET_DIR}/usr/lib/udev/rules.d"
-	    cp -f "${BR2_EXTERNAL_HASSOS_PATH}/bootloader/mbr-part.rules" "${TARGET_DIR}/usr/lib/udev/rules.d/"
+        # FIXME: partition numbers are different with current genimage layout
```
agners, line 60: I guess that is no longer true now?

buildroot-external/scripts/hdd-image.sh
```diff
@@ -1,326 +1,53 @@
 #!/bin/bash
 # shellcheck disable=SC2155
```
agners, line 2: Can we maybe drop this? :thinking:

buildroot-external/scripts/rauc.sh
sairon: Correct!

buildroot-external/scripts/hdd-image.sh
sairon: Removed, after the cleanup there are not that many lines that use it anymore.

agners (approved):
LGTM! Thanks :+1:

PR #3389, opened 2024-05-23
Reintroduce IPv6 reachability probe patch for RPi lost after refactoring

sairon:
In #3384 we moved the patches around, which results in version-specific patches not applied for RPi linux-custom build. Copy the missing IPv6 reachability probe patch to the RPi patches directory.


agners (approved):

sairon:
PTAL, this one is less complicated and does the thing as well.

agners (approved):

PR #3400, opened 2024-06-04
RaspberryPi: Update kernel to 6.6.31 - stable_20240529

sairon:


sairon:
Related Buildroot PR: https://github.com/home-assistant/buildroot/pull/44

sairon:
Right, we had to wait exactly for 6.6.31 because that's where the USB/Z-Wave.me patch started failing to apply. Now that RPi is on this version, we can revert back to the previous structure.

Issue #3351, opened 2024-05-08
Upgrade Odroid N2+ to 12.3, now unable to boot.


csimonosian:
### Describe the issue you are experiencing

Upgraded my Odroid to HAOS 12.3, now HA is not accessible.

### What operating system image do you use?

odroid-n2 (Hardkernel ODROID-N2/N2+)

### What version of Home Assistant Operating System is installed?

12.3

### Did you upgrade the Operating System.

Yes

### Steps to reproduce the issue

1. Accept the installation from HA's prompt that a new release is available.
2. That's it. The instance becomes unreachable.

### Anything in the Supervisor logs that might be useful for us?

```txt
Cannot get to supervisor logs, the system is down.
```

### Anything in the Host logs that might be useful for us?

```txt
No.
```

### System information

Cannot get to System Information because the system is unreachable.

csimonosian:
Booting with a keyboard and monitor attached produces this
![IMG_20240508_213120448_HDR](https://github.com/home-assistant/operating-system/assets/72629997/e426e001-fb06-41d1-810d-2677040d1afe)

agners:
Ugh :cry: 

What eMMC size/type do you have?

tengels:
I have the "standard" [64GB eMCC from Hard Kernel](https://www.hardkernel.com/shop/64gb-emmc-module-n2-linux/)

csimonosian:
Mine was the 32GB eMMC from the UK distributor. https://www.odroid.co.uk/odroid-accessories/odroid-os?product_id=1118.  I have been preparing myself for a clean install of 12.2!

aumuelle:
very same problem here on ODROID N2+ 4GB and eMMC 5.1, 64 GB from ALLNET (https://www.reichelt.at/at/de/entwicklerboards-emmc-5-1-64-gb-erweiterungsspeicher-debo-emmc5-0-64-p248578.html?search=emmc&&r=1)

macrouch:
I'm having the same problem. I don't understand much of this. How do I undo what happened in #3319? All I have is an `HAOS>` prompt and I can't control anything in my house now 😠

stleng08:
Same issue.  Wife mad.  Awaiting a fix or way back to get out of the dog house.

aumuelle:
For those who need a "way back now":  If you don't have a recent backup (?!?) .... 1 out of 10 or 20 boot attempts seems to be successfully here on my side. So perhaps you are lucky too. 
[...]

PR #3409, opened 2024-06-11
Temporary revert ODROID-N2 U-Boot to 2023.01 to fix eMMC issues

sairon:
As we don't have proper solution for #3319 and #3351 yet, revert to previous U-Boot which was proven working. This is intended as a workaround but as there's nothing in the latest U-Boot that will be really missed on N2, we can stay on the older version for the time being.

This also means reverting the "40 MHz hack" back to the 24 MHz one. Since this patch only applies to N2 (meson gx), it can stay along the common hardkernel uboot patches.


agners (approved):
Generally LGTM.

Was just wondering, is the other HAOS specific patch (to add fileenv) really apply cleanly to 2023 and 2024? :thinking:

sairon:
Yep, it does. It passed a build & run test on N2 without errors.

PR #3412, opened 2024-06-12
Fix writing to eMMC in U-Boot on Green

sairon:
When Green starts, there is an error indicating the MMC write failed when saving the bootstate:

```
storing env...

MMC write: dev # 0, block # 1214464, count 64 ... mmc write failed 0 blocks written: ERROR
```

This results in the boot count not being updated properly if the boot fails.

Seems to be a known issue for this platform, disabling the DDR52 mode (which is the same what upstream does for other RK356x boards [1]) fixes the issues and the bootstate is updated correctly.

[1] https://patchwork.ozlabs.org/project/uboot/patch/20240204205312.2342868-2-jonas@kwiboo.se/


agners (approved):
Nice find!

PR #3416, opened 2024-06-13
Improve resiliency of skopeo operations in hassio package

sairon:
We still face occasional build errors when fetching from the Docker registry fails and is not retried with the Skopeo's built-in retry mechanism that was enabled in #1866. This happens on some network failures, or when premature EOF is returned when fetching the HTTP data. Seems we're not the only ones having such issues [1].

To workaround this, add a generic retry shell function that simply retries when the command ends with a non-zero status, no matter what was the actual cause of the error.

[1] https://www.github.com/containers/common/issues/654

### DIFF
--- a/buildroot-external/package/hassio/fetch-container-image.sh
+++ b/buildroot-external/package/hassio/fetch-container-image.sh
@@ -11,6 +11,26 @@ image_json_name=$4
 dl_dir=$5
 dst_dir=$6
 
+retry() {
+	local retries="$1"
+	local cmd=$2
+
+	local output
+	output=$(eval "$cmd")
+	local rc=$?
+
+	# shellcheck disable=SC2086
+	if [ $rc -ne 0 ] && [ $retries -gt 0 ]; then
+		echo "Retrying \"$cmd\" $retries more times..." >&2
+		sleep 3s
+		# shellcheck disable=SC2004
+		retry $(($retries - 1)) "$cmd"
+	else
+		echo "$output"
+		return $rc
+	fi
+}
+
 image_name=$(jq -e -r --arg image_json_name "${image_json_name}" \
 	--arg arch "${arch}" --arg machine "${machine}" \
 	'.images[$image_json_name] | sub("{arch}"; $arch) | sub("{machine}"; $machine)' \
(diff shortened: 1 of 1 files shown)


buildroot-external/package/hassio/fetch-container-image.sh
```diff
@@ -11,6 +11,26 @@ image_json_name=$4
...
+	local retries="$1"
+	local cmd=$2
+
+	local output
+	output=$(eval "$cmd")
+	local rc=$?
+
+	# shellcheck disable=SC2086
+	if [ $rc -ne 0 ] && [ $retries -gt 0 ]; then
+		echo "Retrying \"$cmd\" $retries more times..." >&2
```
sairon, line 24: Currently `retry 3 ...` actually does N **retries**, not sure whether it shouldn't do N-1, so the actual number of tries would be what was passed as the first argument. Let me know what you think.

buildroot-external/package/hassio/fetch-container-image.sh
```diff
@@ -19,7 +39,7 @@ image_tag=$(jq -e -r --arg image_json_name "${image_json_name}" \
 	'.[$image_json_name]' < "${version_json}")
 full_image_name="${image_name}:${image_tag}"
 
-image_digest=$(skopeo inspect --retry-times=5 "docker://${full_image_name}" | jq -r '.Digest')
```
agners, line 22: Is the skopeo retry completely useless? Or maybe worth having it retry twice or so? :thinking:

buildroot-external/package/hassio/fetch-container-image.sh
agners: Hm, yeah N-1 seems what I would have expected. But then, it's retry logic, the number is a bit arbitrary anyways. I don't care that much.

buildroot-external/package/hassio/fetch-container-image.sh
sairon: Not sure, I haven't checked the logs in detail to verify if it fails somewhere and skopeo retries. But IMO it makes more sense to have only a single retry mechanism, instead of having two leading up to M*N retries.

agners: I guess the only case where it might make sense if the `skopeo` retry is somewhat smarter, e.g. retries individual layers.

In any case, a complete retry is fine by me too. It should be an exception anyways.

PR #3436, opened 2024-06-20
Generate version information for RAUC when rauc.db is empty

sairon:
RAUC currently doesn't know the version of the booted slot when booted for the first time or after wiping the data partition. As a result `ha os info` is missing this information too.

As there's no built-in mechanism for generating these data by RAUC itself, add a oneshot service that checks if the boot slot information is contained in the rauc.db and if not, then generate it.

RAUC seems to cope quite well even with bogus data contained in rauc.db but in any case, a test has been added to check that everything works as expected.

### DIFF
--- a/buildroot-external/rootfs-overlay/usr/lib/systemd/system/raucdb-update.service
+++ b/buildroot-external/rootfs-overlay/usr/lib/systemd/system/raucdb-update.service
@@ -1,0 +1,12 @@
+[Unit]
+Description=Ensure rauc.db contains version information
+After=rauc.service
+Before=hassos-supervisor.service
+RequiresMountsFor=/mnt/data
+
+[Service]
+Type=oneshot
+ExecStart=/usr/libexec/raucdb-update
+
+[Install]
+WantedBy=multi-user.target
--- a/buildroot-external/rootfs-overlay/usr/libexec/raucdb-update
+++ b/buildroot-external/rootfs-overlay/usr/libexec/raucdb-update
@@ -1,4 +1,5 @@
 #!/bin/sh
+# shellcheck disable=SC1091
 
 set -e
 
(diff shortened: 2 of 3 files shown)


buildroot-external/rootfs-overlay/usr/lib/systemd/system/raucdb-update.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Ensure rauc.db contains version information
+After=rauc.service
+Before=hassos-supervisor.service
+RequiresMountsFor=/mnt/data
```
agners, line 5: Can we maybe use

```
ConditionFirstBoot=yes
```

here? :thinking:  Ideally, I guess we need to check if it gets triggered after a device wipe too :thinking:

buildroot-external/rootfs-overlay/usr/libexec/raucdb-update
```diff
@@ -0,0 +1,38 @@
+#!/bin/sh
+# shellcheck disable=SC1091
+
+set -e
+
+eval "$(rauc status --output-format=shell)"
+. /etc/os-release
```
agners, line 7: Nit: We could move that sourcing after the rauc.db check to avoid an unnecessary sourcing.

buildroot-external/rootfs-overlay/usr/lib/systemd/system/raucdb-update.service
sairon: I'd say you answered yourself, we'll also need to do that after the device wipe (or at least it's currently what I aimed for and makes more sense). And since we currently determine first boot condition from the presence of MACHINE_ID, it would IMHO only bring more complexity than how it's done now - i.e. ensure the version info is there on every boot.

I've reordered the script a bit more based on your feedback, so the overhead should be minimal now.

agners: We probably should clear the MACHINE_ID on device wipe :thinking: But this is a different discussion.

PR #3456, opened 2024-07-03
Use EROFS for the root filesystem

sairon:
Switch to EROFS for the root filesystem, which promises better performance than SquashFS and with proper options set, it even achieves better compression.

Some hard numbers proving it's better choice are yet to be gathered but so far it looks fine.


agners (approved):
LGTM, thanks! Looking forward to that one :star_struck: .

ThomDietrich:
Hey @sairon,
changes like this one make me wonder: These improvements do not make it into an existing setup. At which point would it make sense for users to consider a new install. Each individual improvement might not justify that but the combination of all could. I know this is not an easy question, there is probably not even a good answer for it.

First question that comes to my mind is whether a user can check with which OS version (and on which date) the initial setup was carried out. @agners is that visible anywhere in Supervisor, and should it?

If this line of question makes sense to you I'd be happy to raise it in a more appropriate place, e.g. under discussions. Apologies if a discussion thread on this already exists.

sairon:
Hi @ThomDietrich :wave: 

> These improvements do not make it into an existing setup.

You're not quite right. This changes the whole rootfs partition which is read-only and gets changed also for existing installations (you can check `mount` output - directly on the host though, not in a terminal/SSH container). The situation would have been different if we changed partition type or flags of any of the overlay or data partitions, but in that case the OTA hook will likely be adjusted to handle this situation.

On the other hand, sometimes there are changes that do not make its way to existing installations. Generally it'd be changes of the partitioning, or changes in `config.txt` or `cmdline.txt` in the boot partition. In that case we consider if a migration mechanism is needed too, or we just ignore the change, because it's likely important for new installs only (for example when adding USB quirks to disable misbehaving UAS driver on RPi).

PR #3469, opened 2024-07-09
Linux: Update kernel to 6.6.38

sairon:
* https://cdn.kernel.org/pub/linux/kernel/v6.x/ChangeLog-6.6.38


sairon:
FTR this is very small release mainly fixing powerpc crashes and arm64 build errors that don't affect us, but let's stay on the latest patch release as always.

frenck (approved):

PR #3508, opened 2024-08-07
Bump ODROID N2's U-Boot to v2024.01 again

sairon:
We previously reverted the bump because we were unsure where the eMMC issues are coming from. Now we know some of them were caused by incompatible eMMCs then never worked from the beggining, and attempt to fix them (by changing the frequency) caused some other side effects. Bump U-Boot back to the version used generally and continue from there.


agners (approved):

smarthomefamilyverrips:
@sairon will this bump affect or break Home Assistant Blue Devices? Or where HA Blue Devices never sold with incompatible eMMC's?

Issue #2723, opened 2023-09-01
NTFS, exfat and xfs modules are not compiled in x86-64 arch


interkelstar:
### Describe the issue you are experiencing

When launching a SambaNAS addon I can see in the logs 

```
modprobe: module ntfs3 not found in modules.dep
[12:33:37] WARNING: ntfs3 module not available!
modprobe: module exfat not found in modules.dep
[12:33:37] WARNING: exfat module not available!
modprobe: module xfs not found in modules.dep
[12:33:37] WARNING: xfs module not available!
```

I am using Home Assistant OS in generic x86-64 device (smth like Intel NUC) and according to this comment from add-on repo
https://github.com/dianlight/hassio-addons/issues/160#issuecomment-1612717416
As well as some others https://github.com/dianlight/hassio-addons/issues/160#issuecomment-1612717416

The problem is that x86-64 version of Hassos doesn't have this modules compiled, but what surprises me that some other arch version has. So is there any particular reason for such a discrepancy? Will it be reasonable to compile these modules into the next release version of Hassos x86-64m even if it's not the main idea of the OS, but if aarch64 can have it, why not us?

### What operating system image do you use?

generic-x86-64 (Generic UEFI capable x86-64 systems)

### What version of Home Assistant Operating System is installed?

Home Assistant OS 10.5

### Did you upgrade the Operating System.

Yes

### Steps to reproduce the issue

1. Install SambaNAS addon from https://github.com/dianlight/hassio-addons
2. Check the logs

### Anything in the Supervisor logs that might be useful for us?

```txt
No
```

### Anything in the Host logs that might be useful for us?

```txt
No
```

agners:
> The problem is that x86-64 version of Hassos doesn't have this modules compiled, but what surprises me that some other arch version has. So is there any particular reason for such a discrepancy?

No particular reason. This is because we typically base board specific kernel default configurations of some existing configuration (e.g. from the Raspberry Pi team). 

I have on my todo to align the kernel configurations a bit more. Essentially move all "generic" kernel modules to the operating system kernel configuration.

> Will it be reasonable to compile these modules into the next release version of Hassos x86-64m even if it's not the main idea of the OS, but if aarch64 can have it, why not us?

I'll consider it. While I rather prefer to not support "wild" usage of HAOS, these modules are upstream and rather maintenance from HAOS point of view.

codyrocco:
[...]

PR #3516, opened 2024-08-11
Enable NTFS and exFAT drivers for all targets

lamauny:


buildroot-external/kernel/v6.6.y/hassos.config
```diff
@@ -43,6 +43,11 @@ CONFIG_SQUASHFS=y
 CONFIG_SQUASHFS_XATTR=y
 CONFIG_SQUASHFS_LZ4=y
 CONFIG_BTRFS_FS=m
+CONFIG_NTFS_FS=m
+CONFIG_NTFS3_FS=m
```
sairon, line 47:
```suggestion
CONFIG_NTFS3_FS=m
```

`NTFS_FS` is the legacy driver and is fully replaced by `NTFS3_FS` in Linux 6.9+. It's rather pointless to have both options in HAOS, so I'll keep just the modern one.

buildroot-external/kernel/v6.6.y/hassos.config
```diff
@@ -43,6 +43,11 @@ CONFIG_SQUASHFS=y
...
+CONFIG_EXFAT_FS=m
+CONFIG_FUSE_FS=m
+CONFIG_XFS_FS=m
```
sairon, line 51:
```suggestion
```

It's very unlikely that FUSE will be needed for the use case that this PR aims to address - i.e. mounting a disk with less common (in the Linux world) filesystem. 

Similarly XFS, with its deployment rather in the enterprise segment. Which I'd prefer removing for one another reason, the module binary is quite big and space becomes a problem especially on the x86-64 platform.

sairon (changes requested):
I'd prefer to remove some options for the reasons said below. I understand the value of NTFS or exFAT, especially for people who want to share the removable drive with Windows machines, but some of those are rather improbable to be seen and enabling them would just increase the bloat in the system partition.

buildroot-external/kernel/v6.6.y/hassos.config
lamauny: Ok, that sounds ok for me, thanks

sairon (approved):

trevornorcross:
> The release notes mentions this and it's a little catchy.
> 
> I really wonder what is the meaning of this in real life.
> 
> Does it mean we can do something we couldn't do before? Can we now plug NTFS drives in it and have them being recognized?

I am also trying to understand this.

codyrocco:
great!

BasPost:
> The release notes mentions this and it's a little catchy.
> 
> I really wonder what is the meaning of this in real life.
> 
> Does it mean we can do something we couldn't do before? Can we now plug NTFS drives in it and have them being recognized?

+1 Would like also to know. I just reinstalled my RPi setup tot Supervisor last weekend to have an external NTFS USB disk mounted.

sairon:
> Does it mean we can do something we couldn't do before? Can we now plug NTFS drives in it and have them being recognized?

See #2723. Not all targets had those drivers enabled in the kernel configs, now it's ensured they would be available consistently.

PR #3523, opened 2024-08-13
Implement device wipe using the hardware button on Green

sairon:
Unlike Yellow, Green doesn't have a way to easily wipe the device, e.g. if the user forgets the password - in that case the only option is to use a microSD card and reflash the system. Fortunately, Green has a hardware button wired to the PMIC chip which exposes the button state in one of the registers. Read this value in U-Boot and decide if cmdline flag for device wipe should be set - same as we do on Yellow.

Also enable LED driver and command in U-Boot. In the current implementation, if the button is held for ~5 seconds when plugging in the device (this time includes DDR training, SPL, etc.), the yellow LED turns solid to indicate wipe is about the start. When the Linux kernel starts, the kernel LED driver takes over and starts blinking in heartbeat pattern. Because it takes a while to load the kernel, the LED stays solid for 2-3 seconds, which should be enough to recognize it was acknowledged.


buildroot-external/board/nabucasa/green/uboot-boot.ush
```diff
@@ -27,6 +27,30 @@ test -n "${MACHINE_ID}" || setenv BOOT_CONDITION "systemd.condition-first-boot=t
 # HassOS bootargs
 setenv bootargs_hassos "zram.enabled=1 zram.num_devices=3 rootwait systemd.machine_id=${MACHINE_ID} fsck.repair=yes ${BOOT_CONDITION}"
 
+# Check if button is held for device wipe
+setenv counter 0
+i2c dev 0
+while test ${counter} -lt 2; do
+  i2c read 0x20 0xf0.1 1 ${loadaddr}
+  setexpr PWRON_STS *${loadaddr} \& 0x80
```
agners, line 35: We could make this a U-Boot environment, which we can run, e.g. `run check_pwron_sts`. 

That way we could just nest tests, get rid of the somewhat awkward while while still avoiding a lot of duplication.

buildroot-external/board/nabucasa/green/uboot-boot.ush
```diff
@@ -27,6 +27,30 @@ test -n "${MACHINE_ID}" || setenv BOOT_CONDITION "systemd.condition-first-boot=t
...
+
+  if test ${PWRON_STS} -eq 0x0; then
+    echo "Reset button pressed for ${counter} seconds"
+    setexpr counter ${counter} + 1
+
+    if test ${counter} -eq 2; then
+      echo "Proceeding with device wipe"
+      led user on
+      setenv bootargs_hassos "${bootargs_hassos} haos.wipe=1"
```
agners, line 44: Probably unlikely, but one of my worry was that this happens accidentally. Maybe far-fetched, but if there is a Green out there which is pushed against the wall (probably unlikely as there is Ethernet) or the case is pressing the button somehow, it will cause a reset on OS update :cold_sweat: 

To solve this, we could test for a falling edge, if the button is pressed, so in pseudocode:

```
if check button pressed
    sleep 1
    if check button pressed
        led user on
        echo "Waiting for button release
        while check button pressed
                sleep 0.5
        done
        echo "Proceeding with device wipe"
        ...
    end
end
```

This would make it "hang" if the button is pressed forever, but I think that is acceptable.

Docs then would state to release the button once the Yellow LED is on to trigger a device wipe.

Wdyt?

Maybe I am overthinking this, we can also leave as is, your call.

buildroot-external/board/nabucasa/green/uboot-boot.ush
sairon: I just ignored it when there was just one occurrence, but since I like the falling edge idea from the other command, it will actually be useful :+1:

buildroot-external/board/nabucasa/green/uboot-boot.ush
sairon: I like it, I will adjust it a bit to make it working as you suggest.

agners (approved):
Nice feature, thanks! :muscle:

PR #3536, opened 2024-08-16
Add LED indication for button handling in Yellow's U-Boot

sairon:
With #3523 as inspiration, it might be useful to wait for buttons to be released, e.g. in case when they become stuck. Also indicate the button operation (wipe, boot files removal, UMS) has been handled by turning on the yellow LED.


agners (approved):
LGTM

PR #3568, opened 2024-08-29
Purge Cloudflare cache for all files after uploading R2 artifacts

sairon:
If an attempt to access R2 artifacts is made before the files exist, the 404 reply gets cached and it's not possible to access the file after it's been created without purging the cache, essentially doing a cache poisoning for future build artifacts. To avoid it, list all files that have been created by the build and call the purge cache API.

As there's a limit for number of files that can be purged in a single API call [1], the GNU split utility is used to split intermediary list of files to chunks of 30 URLs, which is then converted to a JSON array and passed to the curl command.

[1] https://developers.cloudflare.com/api/operations/zone-purge#purge-cached-content-by-url

### DIFF
--- a/.github/workflows/artifacts-index.yaml
+++ b/.github/workflows/artifacts-index.yaml
@@ -83,11 +83,20 @@ jobs:
 
       - name: Flush CloudFlare cache
         run: |
-          curl --silent --show-error --fail -X POST \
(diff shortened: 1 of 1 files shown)


agners (approved):
LGTM!

PR #2051, opened 2022-08-09
Support IPv6 NAT

pvizeli:
Need: https://github.com/home-assistant/supervisor/pull/3780

PR #3576, opened 2024-09-02
Bump Docker to v27.2.0

sairon:
Update Docker and containerd to latest supported version. Also update DinD image used in data partition setup.


sairon:
Buildroot PR: https://github.com/home-assistant/buildroot/pull/55

agners (approved):
I was a bit worried about this change because I've recently realized that Docker touches IPv6 tables quite substantially: Since Docker 27.x `ip6tables` setting is enabled by default (see https://github.com/moby/moby/issues/48365). This could influence our OpenThread Border Router. However, I realized that we've enable the `ip6tables` option already since quite a while (specifically #2051). So this new default should not change things for HAOS essentially. In theory, we could drop the explicit setting, but I'd rather keep it.

In tests, OTBR also works fine with this change. So LGTM, thanks!

PR #3581, opened 2024-09-10
Enable debug logging for GRUB rescue shell options

sairon:
If the system fails to boot, some kernel messages may not be shown before the system fully boots. Enable the debug option for rescue shell options to have an easy way to see those without modifying GRUB options. This will increase log verbosity and turn on debug logging for Systemd as well [1].

[1] https://www.freedesktop.org/software/systemd/man/latest/systemd.html#debug

## Summary by CodeRabbit

- **New Features**
	- Added a `debug` parameter to the boot configuration for both "Slot A" and "Slot B" rescue shell options to enhance troubleshooting capabilities during system initialization.


agners (approved):
LGTM, thanks!

PR #3582, opened 2024-09-11
Relocate HAOS Systemd drop-ins to /usr/lib/systemd

sairon:
With some exceptions, Systemd drop-ins overriding default unit configuration have been placed to `/etc/systemd/system`. This is meant for user overrides of those, or per `man 5 systemd.unit` for "system unites created by the administrator". Relocate all of these to `/usr/lib/systemd` which should be used as path for units "installed by the distribution package manager" which is
closer to what we're trying to achieve.

This will make it easier to detect changes to unit files once we enable the possibility to edit the content of /etc.


buildroot-external/patches/systemd/0003-systemd-timesyncd-delay-start-after-network-online.t.patch
```diff
@@ -0,0 +1,41 @@
...
+Date: Wed, 11 Sep 2024 17:25:00 +0200
+Subject: [PATCH] systemd-timesyncd: delay start after network-online.target
+
+As explained in [1], it's desired for the systemd-timesyncd to run after
+the network is deemed online, otherwise the connectivity (at least on
+HAOS with NetworkManager) is not operational when the NTP sync is
+attempted and it fails, delaying the boot and leading to other problems.
+Because it's not possible to remove dependencies of existing units using
+drop-ins, patch the service template file for systemd-timesyncd unit
+instead, avoiding the need for complete unit file override as in [2].
```
agners, line 12: Shouldn't this include a delete of `systemd-timesyncd.service` then? :thinking:

sairon: Yes, it absolutely should! 🤦🏻‍♂️I will amend the last commit to remove it. Thanks!

agners (approved):

PR #3583, opened 2024-09-11
Add test for no failed Systemd units

sairon:
Test that systemctl reports no failed units at the end of the basic test.

## Summary by CodeRabbit

- **Tests**
	- Introduced a new test to verify that there are no failed system units, enhancing the robustness of the testing framework.

### DIFF
--- a/tests/smoke_test/test_basic.py
+++ b/tests/smoke_test/test_basic.py
@@ -71,3 +71,8 @@ def test_supervisor_logs(shell):
 def test_systemctl_status(shell):
     output = shell.run_check("systemctl --no-pager -l status -a || true")
     _LOGGER.info("%s", "\n".join(output))
+
+@pytest.mark.dependency(depends=["test_init"])
+def test_systemctl_check_no_failed(shell):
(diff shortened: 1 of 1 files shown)


Impact123:
Checking the exit code of `systemctl --failed --quiet` or the line count of `systemctl --failed --legend=false` might be more reliable in the long-term in case the text is ever changed in the future.

tests/smoke_test/test_basic.py
```diff
@@ -71,3 +71,8 @@ def test_supervisor_logs(shell):
 def test_systemctl_status(shell):
     output = shell.run_check("systemctl --no-pager -l status -a || true")
     _LOGGER.info("%s", "\n".join(output))
+
+@pytest.mark.dependency(depends=["test_init"])
+def test_systemctl_check_no_failed(shell):
+    output = shell.run_check("systemctl --no-pager -l list-units --state=failed")
```
agners, line 77: Threre is `--output=json`, but not sure if it makes for a better test :thinking: Either way works for me.

agners (approved):

PR #3589, opened 2024-09-17
Update buildroot branch in .gitmodules to 2024.02.x-haos

adeepn:
Set 2024.02-haos as the default remote tracking branch. This should not
influence regular submodule updates/inits as they reference the git
hash tracked by the operating-system repository directly (#2288)

## Summary by CodeRabbit

- **New Features**
	- Updated the buildroot submodule to a newer version, enhancing compatibility and functionality.
  
- **Bug Fixes**
	- Incorporated improvements and fixes from the latest buildroot branch.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "buildroot"]
 	path = buildroot
 	url = https://github.com/home-assistant/buildroot.git
-	branch = 2022.02.x-haos
+	branch = 2024.02.x-haos


sairon (approved):
I think it only affects `git submodule update --remote`, which we usually don't do, as it doesn't fit our workflow. But it's still good idea to keep the expected remote branch up to date in the gitmodules file. I will try to keep it in mind next time we update Buildroot :) Thanks!

agners (approved):

Issue #3250, opened 2024-03-14
ODROID-N2 doesn't boot from Kingston eMMC modules


piercy:
### Describe the issue you are experiencing

I recently purchased an odroid n2+ and 32gb emmc module.  After flashing `haos_odroid-n2-12.1.img.xz`, first with a usb emmc device and then trying to do it via USB connection with the odroid n2+ device.  The flash was successful in both cases, but once the device was powered up I couldnt access HAOS via IP or homeassistant.local.   I knew the IP was correct as I could compare the mac address.

There was no blue light on the board, just a red one.  After some trial and error I went back to HAOS 12.0 and it flashed and worked first time. I had a blue light and I could immediately access it through a browser.

### What operating system image do you use?

odroid-n2 (Hardkernel ODROID-N2/N2+)

### What version of Home Assistant Operating System is installed?

12.1

### Did you upgrade the Operating System.

No

### Steps to reproduce the issue

1. Flash `haos_odroid-n2-12.1.img.xz` to an odroid device
2. Power it on
3. observe no blue light and home assistant is inaccessible
...

### Anything in the Supervisor logs that might be useful for us?

```txt
Couldn't get to logs as it wouldn't boot
```

### Anything in the Host logs that might be useful for us?

```txt
Couldn't get to logs as it wouldn't boot
```

scoobydrvr:
Just jumping in to say I have a very similar issue with my new n2+ as well...

https://github.com/home-assistant/operating-system/issues/3227

piercy:
I have been struggling with this again today, seems to not be based on the version. neither 12.0 or 12.1 work for me.  However, if i flash ubuntu the device boots fine. So i dont think it can be a device or emmc issue.  

No idea what to do any more but it seems like I cant use HAOS with my odroid n2+

scoobydrvr:
Just FYI - I had to buy a USB to serial converter to get any indication of what's going on. I've been informed they changed the bootloader between 11.4 and 11.5. 11.4 boots fine but 11.5 gives me a message over serial saying that it's unable to find the eMMC card (see my thread above for specifics).

piercy:
> Just FYI - I had to buy a USB to serial converter to get any indication of what's going on. I've been informed they changed the bootloader between 11.4 and 11.5. 11.4 boots fine but 11.5 gives me a message over serial saying that it's unable to find the eMMC card (see my thread above for specifics).

[...]

PR #3619, opened 2024-10-10
Add patch to fix Kingston eMMC compatibility with ODROID-N2

sairon:
Testing shows that the 24MHz cap is problematic for Kingston. Bumping this cap to 40 MHz in #3319 fixed Kingston but reintroduced issues with some other eMMCs that the frequency cap was supposed to fix.

This commit adds another patch that selects a different clock for eMMC. While it's still unclear why Kingston fails with the frequencies of slightly below 24 MHz (see patch message why it's below), using 24 MHz from xtal seems to be fine and hopefully shouldn't introduce regressions for old eMMCs.

Alternative approach would be using the same clock source but setting the frequency cap to 25 MHz. This results in exact 25 MHz from PLL which is even the defined frequency for some eMMC modes, Kingston eMMC works fine with it and the difference from the previously used value wouldn't be too high, but with 24 MHz we're even closer.

refs #3250, refs #3227

## Summary by CodeRabbit

- **Bug Fixes**
	- Resolved communication issues with Kingston eMMCs by ensuring precise clock rates at 24 MHz.
	- Improved compatibility with other eMMCs by adjusting the clock source configuration.


agners (approved):
LGTM!

The 24MHz directly from XTAL potentially is the better clock source indeed, so let's try how this works out in practice on some eMMC modules!

sairon:
Unfortunately, this [doesn't work as good](https://github.com/home-assistant/operating-system/issues/3250#issuecomment-2406085832) as I initially thought :cry: I will try few more things and run some automated test to test the patch works reliably the next time.

PR #3642, opened 2024-10-22
Add kernel config to support external i2c devices

xtsrthm:
This pull request adds support for usb to i2c adapters, the i2c chardev and the bme280  family environment sensors for the ova board.

The same changes were already made to the kernel config for the generic-x86-64 board in #2800 and #2813.

Now i would like to move Home Assistant from a bare metal install to a virtual machine.

## Summary by CodeRabbit

- **New Features**
	- Enhanced support for I2C and GPIO devices, improving compatibility with various hardware interfaces.
	- Added support for new sensors, including BMP280, enhancing sensor integration capabilities.


sairon (approved):
Have you had an opportunity to test it? Otherwise it looks good to me, thanks!

0x3fiona:
I built HAOS locally and tried loading the relevant kernel modules in a VM, but haven't tested this with actual hardware, cause that requires physical access, if you want i can do that though

xtsrthm:
I built HAOS locally and tried loading the relevant kernel modules in a VM, but haven't tested this with actual hardware, cause that requires physical access, if you want i can do that though

PR #3645, opened 2024-10-23
Update test_offline.py

gaurimakhija:
Problem: This loop can run indefinitely if the network or container never becomes available, leading to a hang in the test execution. It would be better to add a timeout or a maximum retry limit.
Fix: Add a retry count or a timeout to prevent infinite looping

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved error handling in offline tests by adding a timeout assertion for supervisor network creation.


sairon (commented):
"Fix" for a problem that doesn't exist. Also, in `while True: ... else` construct, the `else` branch would be never evaluated.

PR #3646, opened 2024-10-23
Update test_offline.py

gaurimakhija:
Problem: The string "is alive!" is unlikely to appear in the output of a ping command. Typically, the output will contain something like "bytes from ..." or "time=" in successful pings.
Fix: Check for actual ping response patterns, such as "bytes from" or "time="


sairon (commented):
HAOS uses Busybox ping which doesn't return the output assumed to be returned by this PR.

PR #3647, opened 2024-10-23
Update test_supervisor.py

gaurimakhija:
The while loops in test_start_supervisor and other tests do not have timeouts, which may cause the tests to hang indefinitely if something goes wrong (e.g., the supervisor never starts or the API never returns the expected result).

Fix: Add a maximum retry count or timeout to prevent infinite looping:

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved the reliability of the supervisor startup tests with a new retry mechanism for container readiness.
	- Enhanced error handling to better manage timeout scenarios during container initialization.

### DIFF
--- a/tests/supervisor_test/test_supervisor.py
+++ b/tests/supervisor_test/test_supervisor.py
@@ -32,24 +32,36 @@ def test_start_supervisor(shell, shell_json):
         shell.run_check("docker inspect --format='{{.NetworkSettings.IPAddress}}' hassio_supervisor")
     )
 
-    while True:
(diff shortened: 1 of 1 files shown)


sairon (commented):
Overly complex solution for a problem that doesn't exist. The timeouts are handled by the `pytest.mark.timeout` annotation. The only part that is relevant is that the timeout might be too long (it had to be very long in the past where GH Actions runners didn't support KVM acceleration) but that's quite irrelevant here.

PR #3648, opened 2024-10-23
Update conftest.py

gaurimakhija:
The shell.run_check() method likely returns a list of strings (lines from the command output). However, the current implementation assumes it's a single string, using "\n".join(shell.run_check(...)). This could lead to unintended behavior if the output is already a single string or not formatted as expected.

Fix: You can simplify this by directly parsing the result without joining the lines, as json.loads can handle strings with line breaks. If the output is indeed a list, join them properly.


sairon (commented):
We can rely on the `run_check` to return a list, so this is really unnecessary.

PR #3665, opened 2024-11-07
Fall back to max eMMC clock to fix Kingston eMMC compat with ODROID-N2

sairon:
Replace the original patch that limits the frequency to 24 MHz with a more sophisticated version that first tries to initialize the eMMC at this limited frequency, and if it fails, then tries disabling the limit to use one defined in the upstream driver.

While still hacky, this seems to finally fix the Kingston eMMC issues, and with the current implementation, the initialization should appear the same as on older HAOS versions, as 24 MHz is used first. Better solution that addresses the root cause on some eMMCs (128 GB Microns?) would be still nice though...

refs #3250, refs #3227

## Summary by CodeRabbit

- **New Features**
	- Introduced a maximum clock frequency limit of 24 MHz for eMMC cards on the Meson GX platform to enhance compatibility during initialization.

- **Bug Fixes**
	- Updated driver behavior to disable the frequency cap if initialization fails, allowing for reinitialization at higher clock rates when necessary.


sairon:
I ran an automated test where the ODROID cold-booted over 100 times in a row without issues. Now I'm running another one where I'm leaving it to boot to the system and run a `reboot` afterwards, now at 30 attempts at counting. Looking good on my end.

frenck (approved):

camarillla:
@sairon, I bought a odroid N2 with pre-installed Hassio and run into problems which this patch could resolve. I am complete noob on how to use this patch file and update the driver file. Is it possible to add a step-by-step approach. Many thanks in advance :)

PR #3672, opened 2024-11-14
Remove unnecessary kernel post-install hook for GRUB-based targets

sairon:
The code for this hook was removed in #3457 but it wasn't removed from the manifest. Remove it to avoid unnecessary execution of the hook.

## Summary by CodeRabbit

- **Chores**
	- Updated conditions for bootloader and image component hooks to streamline execution logic.
	- Simplified the condition for the `[image.kernel]` section to check only for "tryboot".

### DIFF
--- a/buildroot-external/ota/manifest.raucm.gtpl
+++ b/buildroot-external/ota/manifest.raucm.gtpl
@@ -19,8 +19,7 @@ hooks=install;
 
 [image.kernel]
 filename=kernel.img
-{{- $bootloader := (env "BOOTLOADER") }}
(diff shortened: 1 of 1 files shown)


agners:
The hook is still needed so `check_grubenv` gets called no? :thinking:

sairon:
I don't think so, `check_grubenv` is called from `install-check` hook enabled in global `[hooks]` section.

agners:
> I don't think so, `check_grubenv` is called from `install-check` hook enabled in global `[hooks]` section.

Oh right, this is about `post_install_kernel`, which we only use for RPi 5/tryboot now. Make sense!

agners (approved):

PR #3700, opened 2024-11-28
Reduce time of mmc0 interface random hangs on Yellow+CM5 boot

sairon:
For yet unknown root cause, the eMMC interface sometimes fails to initialize properly, delaying boot for up to 130 seconds. This can be reduced by ~100s by disabling SD and SDIO modes on the sdio1 interface used for mmc0 before a better patch is found.

## Summary by CodeRabbit

- **New Features**
	- Introduced a patch to disable SD and SDIO modes on the eMMC interface for improved performance.
  
- **Bug Fixes**
	- Addressed significant delays in eMMC initialization, reducing total reset time by approximately 100 seconds.

### DIFF
--- a/buildroot-external/board/raspberrypi/yellow/patches/linux/0018-ARM-dts-bcm2712-yellow-Disable-SD-SDIO-modes-on-eMMC.patch
+++ b/buildroot-external/board/raspberrypi/yellow/patches/linux/0018-ARM-dts-bcm2712-yellow-Disable-SD-SDIO-modes-on-eMMC.patch
@@ -1,0 +1,44 @@
+From 9639a7a0f20cbd4a7b9968f6a3ede4727d96d297 Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Jan=20=C4=8Cerm=C3=A1k?= <sairon@sairon.cz>
+Date: Thu, 28 Nov 2024 18:05:55 +0100
+Subject: [PATCH] ARM: dts: bcm2712: yellow: Disable SD/SDIO modes on eMMC
+ interface
+MIME-Version: 1.0
+Content-Type: text/plain; charset=UTF-8
+Content-Transfer-Encoding: 8bit
+
+If the eMMC card is not initialized properly (probably a side-effect of its
+usage in U-Boot), every command sent to it takes over 10 seconds before
+timeout. With SD and SDIO modes not disabled in the device tree, it takes up
+over 130 seconds before the card is reset. Disabling these two modes reduces
+this by 100 seconds.
+
+While this is added as a partial workaround for the issue, disabling these mode
+should make no harm anyway.
+
+Signed-off-by: Jan Čermák <sairon@sairon.cz>
+---
+ .../boot/dts/broadcom/bcm2712-rpi-cm5-ha-yellow.dts    | 10 ++++++++++
+ 1 file changed, 10 insertions(+)
+
+diff --git a/arch/arm64/boot/dts/broadcom/bcm2712-rpi-cm5-ha-yellow.dts b/arch/arm64/boot/dts/broadcom/bcm2712-rpi-cm5-ha-yellow.dts
+index c85d608da2043..189c17fe2028e 100644
+--- a/arch/arm64/boot/dts/broadcom/bcm2712-rpi-cm5-ha-yellow.dts
++++ b/arch/arm64/boot/dts/broadcom/bcm2712-rpi-cm5-ha-yellow.dts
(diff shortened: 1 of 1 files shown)


agners (approved):

sairon:
Revisiting this after couple of years - deeper investigation shows that origin of the issue indeed lies in U-Boot setting some registers incorrectly. The lockup even with this patch results in some boots delayed like this (boot log with `CONFIG_DYNAMIC_DEBUG=y`:

```
[    0.882100] mmc0: starting CMD0 arg 00000000 flags 000000c0
[   10.973134] mmc0: Timeout waiting for hardware cmd interrupt.
[   11.086966] mmc0: req done (CMD0): -110: 00000000 00000000 00000000 00000000
[   11.095273] mmc0: clock 400000Hz busmode 2 powermode 2 cs 0 Vdd 21 width 1 timing 0
[   11.104207] mmc0: clock 400000Hz busmode 1 powermode 2 cs 0 Vdd 21 width 1 timing 0
[   11.111890] mmc0: starting CMD1 arg 00000000 flags 000000e1
[   21.213134] mmc0: Timeout waiting for hardware cmd interrupt.
[   21.326946] mmc0: req done (CMD1): -110: 00000000 00000000 00000000 00000000
[   21.334002] sdhci-brcmstb 1000fff000.mmc: sdhci_brcmstb_hs400es(): Setting HS400-Enhanced-Strobe mode
[   21.343217] mmc0: clock 0Hz busmode 2 powermode 0 cs 0 Vdd 0 width 1 timing 0
[   21.351622] mmc0: mmc_rescan_try_freq: trying to init card at 300000 Hz
[   21.358239] sdhci-brcmstb 1000fff000.mmc: sdhci_brcmstb_hs400es(): Setting HS400-Enhanced-Strobe mode
```

Some more debug prints and reboots later, it became obvious that `POWER_CTRL` register in the failed cases is `0x0e` (3.3V, POWER_ON bit clear) despite `sdhci_brcmstb_set_power` [should set it](https://github.com/raspberrypi/linux/blob/ae4d75fb36deba0fe1a986d9bfae65755e082dd0/drivers/mmc/host/sdhci-brcmstb.c#L214) to `0x0f` (POWER_ON | 3.3V). Retrying to set the register fixed the problem but it wasn't obviously the right solution.

Since the issue only manifested when kernel was loaded by U-Boot but not by the EEPROM bootloader, I focused on the differences between other registers, and `SD_PIN_SEL` caught my eye - it was always set to `0x02` (SD mode) while EEPROM loader leaves it at `0x00`.

Turns out U-Boot's check in [`sdhci_brcmstb_init_2712()`](https://github.com/u-boot/u-boot/blob/v2026.04/drivers/mmc/bcmstb_sdhci.c#L68-L69) tests `host->mmc->host_caps` to pick between the non-removable branch (force card-detect via `SDIO_CFG_CTRL`) and the removable branch (route the CD line via `SDIO_CFG_SD_PIN_SEL`). But that field is still zero at the time `init_2712` runs from `sdhci_bcmstb_probe()` - `host->mmc->host_caps` is populated later by the MMC uclass, *after* the driver's probe returns. The DT-parsed caps live in `plat->cfg.host_caps` at this point (filled in by the earlier `mmc_of_parse()` call), and that's what the check should be reading. U-Boot's own generic `sdhci.c` gets this right one level up, but `bcmstb_sdhci.c` doesn't.

So every BCM2712 SDHCI instance unconditionally takes the *else* branch and writes `SDIO_CFG_SD_PIN_SEL = SDIO_CFG_SD_PIN_SEL_CARD` (0x02), including the non-removable eMMC. The `SDIO_CFG` block lies outside the SDHCI core's reset scope, so this value persists across `SDHCI_RESET_ALL` straight into Linux's probe. On BCM2712, having `SD_PIN_SEL = SD` when Linux's mmc-core does its first `set_power(MMC_POWER_UP)` write apparently disturbs whatever gates `POWER_ON` - the voltage bits in the same byte stick, only bit 0 drops, and the controller never drives the bus for the first `CMD0`. The EEPROM loader leaves the register at its hardware default (0x00) because it doesn't run this driver init, which is why EEPROM boots never showed the symptom.

A patch fixing the check in U-Boot is on its way upstream (replace `host->mmc->host_caps` with `plat->cfg.host_caps` in `init_2712`). With that, `SD_PIN_SEL` stays at 0x00 for the eMMC and the Linux race disappears entirely - tested on Yellow/CM5 over 100 boots, no lock-up with the patch, without it, it failed in ~50% of cases. With the patch in, we could also drop this DTS hack to keep the device tree simple and closer to RPi's CM5 DTS.

PR #3721, opened 2024-12-06
Fix Hailo-8 firmware path expected since driver v4.19.0

sairon:
The new driver now expects the firmware file to contain version number, adjust the path and remove unnecessary makefile step.

## Summary by CodeRabbit

- **New Features**
	- Streamlined firmware build process for Hailo-8 by removing unnecessary commands.
	- Installation command now directly references the source filename for improved consistency.

- **Bug Fixes**
	- Corrected installation command to ensure proper firmware installation without referencing outdated filenames.

PR #3732, opened 2024-12-10
Compile hailo_pci module from original sources on RPi 5

sairon:
Instead of using in-tree module on RPi 5, build it as a module from the original sources. This will give us better control over the version used and will also allow us for easier way to add the module to other platforms.

This also makes 017d172 unnecessary anymore.

## Summary by CodeRabbit

- **New Features**
  - Added support for the Hailo PCI package, enabling integration into the build configuration.
  - Introduced configuration options for the Hailo PCIe driver, allowing users to enable or disable it.
  - Updated firmware build process for Hailo-8, standardizing firmware binary naming.

- **Bug Fixes**
  - Clarified help text for Hailo-8 firmware configuration to specify dependency on the Hailo PCI kernel module.

- **Chores**
  - Added SHA256 hash entries for file integrity verification in the Hailo PCI package.
  - Enhanced kernel configuration for Raspberry Pi, enabling additional LED triggers and setting local version identifier.


buildroot-external/package/hailo8-firmware/hailo8-firmware.mk
```diff
@@ -13,9 +13,13 @@ define HAILO8_FIRMWARE_EXTRACT_CMDS
 	cp $(HAILO8_FIRMWARE_DL_DIR)/$(HAILO8_FIRMWARE_SOURCE) $(@D)
 endef
 
+define HAILO8_FIRMWARE_BUILD_CMDS
+	cp $(@D)/$(HAILO8_FIRMWARE_SOURCE) $(@D)/hailo8_fw.bin
+endef
+
```
agners, line 19: Is that directly really related to the driver code move in some ways? If not, I think a separate PR would be better.

sairon: Yes, already pointed out here: https://github.com/home-assistant/operating-system/pull/3721#issuecomment-2523179200

Without this change the PR will break the upstream driver.

agners: Ah I see. Make sense then to have this in this PR!

agners (approved):

PR #3737, opened 2024-12-12
Remove existing rauc.db from a data disk on the first boot

sairon:
If data disk is adopted on Yellow using the mechanism added in #3686, it contains RAUC version information that is very likely invalid. In such case, remove the file on first boot and have it recreated by the raucdb-update service.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new systemd service for one-time operations during the first boot of the system.
  
- **Bug Fixes**
	- Ensures the removal of the specified file on first boot if it exists.

### DIFF
--- a/buildroot-external/rootfs-overlay/usr/lib/systemd/system/raucdb-first-boot.service
+++ b/buildroot-external/rootfs-overlay/usr/lib/systemd/system/raucdb-first-boot.service
@@ -1,0 +1,14 @@
+[Unit]
+Description=Remove adopted rauc.db on first boot
+After=rauc.service
+Before=raucdb-update.service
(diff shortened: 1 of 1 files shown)


buildroot-external/rootfs-overlay/usr/lib/systemd/system/raucdb-first-boot.service
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=Remove adopted rauc.db on first boot
+After=rauc.service
```
agners, line 3: Why is this actually `After`? I see that this is the case for `raucdb-update.service` as well, but I am wondering? :thinking:

agners (approved):

PR #3767, opened 2024-12-28
Update generic-x86-64 Linux kernel to 6.12

nijave:
6.6 was missing drivers for the Beelink EQ14 Intel N150

Not sure I got everything/applied patches correctly here

Requires https://github.com/home-assistant/buildroot/pull/63

For package updates, I did not thoroughly review the changelogs or perform any package-specific testing. I only checked that the build completed successfully

## Release Notes

- **Kernel Updates**
  - Upgraded Linux kernel to version 6.12.6 for x86-64 platforms
  - Enhanced memory management with ZRAM and ZSWAP support
  - Improved networking capabilities with IPv6 and MPTCP features
  - Added AppArmor security support

- **Package Updates**
  - Updated various driver packages for kernel 6.12 compatibility
  - Refreshed RTL88X2BU and Gasket package versions
  - Modified USB core to address Z-Wave device compatibility issues

- **Configuration Changes**
  - Updated sound system configurations for Intel platforms
  - Refined kernel configuration for improved hardware support

## TODO

~- finish zram updates so only lz4 is compiled in~
~- finish upstream buildroot & home-assistant/buildroot changes~
~- create busybox patch for Linux >= 6.8 -- needs https://github.com/home-assistant/buildroot/pull/68~


mmorrisontx:
Build fails on a fresh clone because the `buildroot-external/board/pc/patches` directory is missing now that the last file was removed from git.

buildroot-external/rootfs-overlay/usr/libexec/hassos-zram
```diff
@@ -57,7 +57,7 @@ if [ "$TYPE" = "swap" ] && [ "$SIZE" -eq "0" ]; then
 fi
 
 # Init device
-zramctl "$DEVICE" -s "$SIZE" -a lz4
+zramctl "$DEVICE" -s "$SIZE"
```
nijave, line 60: zram has new config options to specify which compression algorithms are included. It defaults to `lzo` unless `lz4` is explicitly enabled. It looks like lz4 was set 5+ years ago--not sure if it was intentionally to use this algorithm or an outdated default

However, I also enabled `lz4` compression in the kernel config options. I don't think both changes are needed--not sure best path
- use default zram compression
- keep as existing lz4

buildroot-external/board/pc/patches/linux/0001-iwlwifi-Make-missed-beacon-timeout-configurable.patch
```diff
@@ -1,94 +0,0 @@
-From 5712a20822709d1cd744a435eba5d9c8ef91a834 Mon Sep 17 00:00:00 2001
```
nijave, line 1: It sounds like this should have been resolved in 6.10 https://bugzilla.kernel.org/show_bug.cgi?id=203709

buildroot-external/busybox.config
```diff
@@ -951,7 +951,7 @@ CONFIG_PING6=y
 # CONFIG_ROUTE is not set
 # CONFIG_SLATTACH is not set
 # CONFIG_SSL_CLIENT is not set
-CONFIG_TC=y
+CONFIG_TC=n
```
nijave, line 954: Can't remember the error this was causing but I don't think it's needed

>TURBOchannel is a DEC (now Compaq (now HP)) bus for Alpha and MIPS
processors. TURBOchannel programming specifications are available

jens-maus: BTW: There is no possibility to specify`=n`for excluding any `CONFIG_XXXX` config option. If you need to ensure `CONFIG_TC` is not set you have to explicitly add a `# CONFIG_TC is not set` line instead.

buildroot-external/package/generic_raw_uart/generic_raw_uart.hash
```diff
@@ -1,3 +1,3 @@
 # Locally computed
 sha256  b40930bbcf80744c86c46a12bc9da056641d722716c378f5659b9e555ef833e1  LICENSE
-sha256  f1741dfe9ea6ea37213c35c735aa899f11e4c2961bccd6d8198d4051b3c42c5d  generic_raw_uart-f691101c32435921a70c96075233b6d5e5eeab7d.tar.gz
+sha256  f8326439b6cc917d4c8ccc6fa670eb5a0ca50d43e08ca02140d9314bbc2429ac  generic_raw_uart-26b6aa161d69bd8dc6900f144b74b4ac884e9dab.tar.gz
```
jens-maus, line 3: No need to bump generic_raw_uart to commit id `26b6aa1` as `[f691101](https://github.com/alexreinert/piVCCU/commit/f691101c32435921a70c96075233b6d5e5eeab7d)` is the last changed commit id that is relevant for the generic_raw_uart package. So please keep it at `f691101` because that's how we (in fact I) ensure/maintain the correct generic_raw_uart sources are used.

buildroot-external/package/generic_raw_uart/generic_raw_uart.mk
```diff
@@ -13,7 +13,7 @@
 #
 ################################################################################
 
-GENERIC_RAW_UART_VERSION = f691101c32435921a70c96075233b6d5e5eeab7d
+GENERIC_RAW_UART_VERSION = 26b6aa161d69bd8dc6900f144b74b4ac884e9dab
```
jens-maus, line 16: same here. Please keep it at `f691101`

jens-maus (changes requested):

buildroot-external/package/generic_raw_uart/0003-Linux-6.12-gpio-data.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/kernel/hb_rf_eth.c b/kernel/hb_rf_eth.c
+index 9d949c003..c8d5f15bf 100644
+--- a/kernel/hb_rf_eth.c
++++ b/kernel/hb_rf_eth.c
+@@ -715,7 +715,7 @@ static int __init hb_rf_eth_init(void)
+   gc.base = -1;
+   gc.can_sleep = false;
+ 
+-  err = gpiochip_add(&gc);
++  err = gpiochip_add_data(&gc, NULL);
```
jens-maus, line 10: And what about a similar `#if LINUX_VERSION_CODE < KERNEL_VERSION(6, 0, 0)` preprocessor statement here to make sure the sources still compile for other kernel versions. And also please check at which kernel version `gpiochip_add_data()` was added and then restrict the `#if LINUX_VERSION_CODE` version check to that specific kernel version.

nijave:
@jens-maus I think I addressed all the comments

buildroot-external/rootfs-overlay/usr/libexec/hassos-zram
nijave: It looks like maybe lz4 is a better option since it's lower CPU usage and lzo-rle (default) doesn't have much better compression

ref https://www.reddit.com/r/Fedora/comments/mzun99/new_zram_tuning_benchmarks/

mmorrisontx:
I've been running this branch on my beelink eq14. Seems to be working great, and the igpu is now usable in plex.

buildroot-external/busybox.config
```diff
@@ -951,7 +951,7 @@ CONFIG_PING6=y
 # CONFIG_ROUTE is not set
 # CONFIG_SLATTACH is not set
 # CONFIG_SSL_CLIENT is not set
-CONFIG_TC=y
+# CONFIG_TC is not set
```
sairon, line 954: You were not correct [here](https://github.com/home-assistant/operating-system/pull/3767#discussion_r1899152223), `CONFIG_TC` is part of the Busybox config and enables the `tc` (traffic control) command. I'm not exactly sure where/if it's used but definitely this needs to be looked into.

sairon (changes requested):
There are some issues I pointed out in the comments but in general very nice job, thanks! :+1:

Once the issues are addressed (especially the RPi configs), we can run a build of all targets to see if anything else emerges there. Maybe `linux-check-dotconfig` will find some removed kernel options.

buildroot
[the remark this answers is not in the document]
sairon: 333ff47f55a97c4127ea8509822be0b768de4b7b from https://github.com/home-assistant/buildroot/pull/67 can be used.

buildroot-external/busybox.config
nijave: Looks like busybox needs a patch
https://github.com/mirror/busybox/pull/96
https://lists.openembedded.org/g/openembedded-core/message/201449

buildroot-external/package/generic_raw_uart/0002-Linux-6.12-llseek.patch
```diff
@@ -0,0 +1,16 @@
+Remove llseek NULL'd in 6.0 (868941b) removed in 6.12 (cb787f4)
+
+Signed-off-by: Nick Venenga <nick@venenga.com>
```
jens-maus, line 3: Please compare to the other *.patch files again. You patch files are still missing an `Upstream:` statement which otherwise would end up in the buildroot linter check to fail.

buildroot-external/package/generic_raw_uart/0003-Linux-6.11-gpio-data.patch
```diff
@@ -0,0 +1,18 @@
+Replace `gpiochip_add` removed in 6.11 (3ff1180) with `gpiochip_add_data` added in 4.5 (b08ea35)
+
+Signed-off-by: Nick Venenga <nick@venenga.com>
```
jens-maus, line 3: Also here `Upstream:´ statement is missing.

buildroot-external/package/eq3_char_loop/0004-Fix-eq3_char_loop-driver-build-on-Linux-6.12.patch
```diff
@@ -0,0 +1,16 @@
+Remove llseek NULL'd in 6.0 (868941b) removed in 6.12 (cb787f4)
+
+Signed-off-by: Nick Venenga <nick@venenga.com>
```
jens-maus, line 3: `Upstream:` statement missing here.

buildroot-external/package/gasket/0001-linux-6.x.patch
```diff
@@ -0,0 +1,34 @@
...
+simply defined to be NULL, and a NULL llseek means seeking is
+unsupported.
+
+refs:
+- https://github.com/torvalds/linux/commit/cb787f4ac0c2e439ea8d7e6387b925f74576bdf8
+- https://github.com/torvalds/linux/commit/868941b14441282ba08761b770fc6cad69d5bdb7
+
+no_llseek has been hard dropped in linux-6.12
+
+Signed-off-by: Rudi Heitbaum <rudi@heitbaum.com>
```
jens-maus, line 16: Also here `Upstream:` statement is missing.

jens-maus (changes requested):

buildroot-external/busybox.config
nijave: https://github.com/home-assistant/buildroot/pull/68

buildroot-external/package/generic_raw_uart/0002-Linux-6.12-llseek.patch
nijave: I assume you mean like `board/hardkernel/odroid-m1/patches/uboot/0001-odroid-m1-Disable-eMMC-DDR52-mode-enable-HS200.patch` described in https://buildroot.org/downloads/manual/patch-policy.txt#:~:text=Additional%20patch%20documentation

I see 119 patches but only 3 have `Upstream:`

Additionally, is there a way to run this linter locally to verify all requirements are satisfied?

sairon: @jens-maus Actually @nijave is kind of correct here - HAOS has many patches without `Upstream` tag and its checking is disabled in CI: <https://github.com/home-assistant/operating-system/blob/dev/.github/workflows/pr-checks.yml#L34>

nijave:
Thanks for the feedback guys. I believe all issues have been resolved, todos finished, and upstream issues/PRs opened

nijave:
I can also squash--not sure if there's a preference in this repo

sairon (changes requested):
Thanks for the changes, looks _almost_ perfect. Just like with 6.6.y patches, we need to preserve 6.6.y kernel config fragments for RPi boards. 

Don't worry with squashing, it will be squash-merged in the end, but we prefer having more granular history at least in the PRs.

nijave:
> Thanks for the changes, looks _almost_ perfect. Just like with 6.6.y patches, we need to preserve 6.6.y kernel config fragments for RPi boards.
> 
> Don't worry with squashing, it will be squash-merged in the end, but we prefer having more granular history at least in the PRs.

this right? 029d556fdee3861ee9d5d6ea33527cea02072bda

sairon:
Thanks, I ran the [build for all platforms](https://github.com/home-assistant/operating-system/actions/runs/12686717600), and unfortunately, generic-x86-64 is actually the only one that compiles without errors :see_no_evil: Also some warnings for missing (or renamed) kernel config options have emerged, as I predicted previously. 

Anyway, I can take it from here, depends if you want to invest your time into it. Or we can simply scope this PR to x86 only (which would need revert of configs for most of the boards) and I'll address remaining platforms in another PR.

jens-maus (dismissed):
In fact, after having implemented similar linux 6.12 changes to my RaspberryMatic side-project, I identified one more issue with the `generic_raw_uart` and Linux kernel 6.11+ compatibility. Please see here:

https://github.com/jens-maus/RaspberryMatic/blob/master/buildroot-external/package/generic_raw_uart/0004-Linux-6.11-uart_remove.patch

Thus, @nijave it would be great if you could get all patches in that https://github.com/jens-maus/RaspberryMatic/blob/master/buildroot-external/package/generic_raw_uart/ patch as well as in https://github.com/jens-maus/RaspberryMatic/blob/master/buildroot-external/package/eq3_char_loop in sync with the ones in this PR. Because this should allow to get at least the tinkerboard platform compiled again.

sairon (commented):
Okay, since we'd like to move forward with merging 6.12, let's do it the second way I suggested, i.e. target this PR at x86-64 only and I'll take care of the rest.

I'll remove the remaining platforms from the PR, rebase it and apply suggestions from Jens. Then we can merge it and use it as a base for complete transition to 6.12. Thanks for your outstanding work, @nijave, though! I'll be able to handle it from here.

nijave:
thanks guys! got sidetracked with other projects

agners (approved):
LGTM!

I'd suggest to also bump Linux firmware to latest git. Usually a new kernel has drivers which do make use of newer firmware files. For Intel WiFi drivers we'd probably have to check what API level the new kernel supports and then go with the latest one supported.

sairon:
> I'd suggest to also bump Linux firmware to latest git. Usually a new kernel has drivers which do make use of newer firmware files. For Intel WiFi drivers we'd probably have to check what API level the new kernel supports and then go with the latest one supported.

Good point, I'll do that in subsequent PR.

PR #3793, opened 2025-01-14
Disable CQE on mmc0 to fix I/O freezes on CM5

sairon:
Similarly to #3705, enabled CQE triggers I/O freezes usually on the first boot when the swapfile is being created. While we disabled it for Yellow, with #3782 the issue started to appear on generic CM5 targets with the rpi5-64 image.

In the meantime it was discovered that there seems to be some relation with the ext4lazyinit task, which happens as a result of data partition resize, yet it's still unclear if the pattern of the access triggered by the concurrent FS initialization is somehow responsible, or if another factor comes in play. Disabling CQE yet makes the issue go away and serves as an acceptable workaround.

## Summary by CodeRabbit

- **Bug Fixes**
	- Resolved eMMC initialization issues on Raspberry Pi Compute Module 5 by disabling Command Queue Engine (CQE) feature.
	- Improved system stability during first boot of Home Assistant Operating System.


agners (approved):

P33M:
FYI this disable should no longer be required if you are sourcing latest tagged rpi-6.6.y  ([stable_20250127](https://github.com/raspberrypi/linux/releases/tag/stable_20250127)) from our downstream Linux repository.

b1rkizil (approved):

PR #3810, opened 2025-01-20
Move rauc.db to boot partition

sairon:
The RAUC metadata file contains information that is tightly related to the system and kernel partitions. With the possibility to migrate data disk, the rauc.db can contain bogus information when moved to a different system. Removal of the file on "device wipe" is also not desirable, because the information about slot status is lost.

Relocate the rauc.db to the boot partition after a system upgrade (as this can't be handled by RAUC hooks, because it needs to be executed after all slots and metadata is written) and adjust the script for recreating it. The downside is that its content in /mnt/data would be recreated if the boot slot is changed or system downgraded but this should be handled quite gracefully.

Also remove the raucdb-first-boot service which is no longer necessary with the file not present in the data partition.

## Summary by CodeRabbit

- **New Features**
	- Added migration support for RAUC database file from `/mnt/data` to `/mnt/boot`
	- Introduced new systemd service for database file migration

- **Bug Fixes**
	- Updated service dependencies to ensure proper file system mounting
	- Improved handling of database file location during system updates

- **Chores**
	- Removed legacy first-boot service related to database file management
	- Updated system configuration to reflect new database file path


agners:
> The downside is that its content in /mnt/data would be recreated if the boot slot is changed or system downgraded but this should be handled quite gracefully.

You meant to say `/mnt/boot` here no?

Could/should we maybe exclude it when boot partition is written? Afaik boot partition upgrade is quite custom already :shrug:

sairon:
> > The downside is that its content in /mnt/data would be recreated if the boot slot is changed or system downgraded but this should be handled quite gracefully.
> 
> You meant to say `/mnt/boot` here no?
> 
> Could/should we maybe exclude it when boot partition is written? Afaik boot partition upgrade is quite custom already 🤷

No, I was really meaning `/mnt/data` here - it's the case when you upgrade to OS 15, the migration is performed and then you revert back to 14.x. In that case the file gets recreated by `raucdb-update` in the data partition again. It's not a problem if you swap back and forth just by changing the slots (you just lose some of the information about the other slot), but we also need to consider the case the OS is upgraded again from the other slot to 15.x - in that case the existing /mnt/boot/rauc.db from previous migration can't be trusted, so that's why the migration is forced on every 14->15 transition.

If I understand what you meant - there is no need to exclude it, the vanilla boot partition does not contain it, it's created on the first boot like it was before, and then it's preserved in the boot partition, untouched by the `install_boot` function of our update hook.

agners (approved):

PR #3812, opened 2025-01-20
Enable USB-SD convertor on AMD/Xilinx Kria KD240 platform

michalsimek:
Kria KD240 board is using SD card but SD is connected via onboard USB HUB. USB controller is DWC3 with Xilinx glue logic. Both of these options are enabled but board is using slg7xl45106 for driving usb-hub reset (PCA9570 driver) and USB3.0 requires initialization via PHY_XILINX_ZYNQMP driver.

All options should be enabled (=y) and can't be kernel modules because provide access to rootfs.

## Summary by CodeRabbit

- **New Features**
	- Added kernel support for PCA9570 GPIO expander
	- Added kernel support for Xilinx Zynq UltraScale+ MPSoC PHY drivers
	- Enabled support for onboard USB hubs


sairon (approved):
LGTM, I just added a note (mainly for self) for the 6.12 migration.

PR #3818, opened 2025-01-21
Fix kernel tainted test

sairon:
I have only tested that it fails for unreleased 6.6.72 kernel but haven't tested the happy path and missed that it also failed because the types were different. Stupid me.

## Summary by CodeRabbit

- **Tests**
	- Updated `test_kernel_not_tainted` function assertions in multiple test files to handle newline characters more precisely when checking kernel tainted status.

### DIFF
--- a/tests/smoke_test/test_basic.py
+++ b/tests/smoke_test/test_basic.py
@@ -83,4 +83,4 @@ def test_kernel_not_tainted(shell):
     """Check if the kernel is not tainted - do it at the end of the
     test suite to increase the chance of catching issues."""
     output = shell.run_check("cat /proc/sys/kernel/tainted")
-    assert output == "0\n", f"Kernel tainted: {output}"
+    assert "\n".join(output) == "0", f"Kernel tainted: {output}"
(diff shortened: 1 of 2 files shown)


tests/smoke_test/test_basic.py
```diff
@@ -83,4 +83,4 @@ def test_kernel_not_tainted(shell):
     """Check if the kernel is not tainted - do it at the end of the
     test suite to increase the chance of catching issues."""
     output = shell.run_check("cat /proc/sys/kernel/tainted")
-    assert output == "0\n", f"Kernel tainted: {output}"
+    assert "\n".join(output) == "0", f"Kernel tainted: {output}"
```
agners, line 86: Isn't that fixing it on both sides? :thinking: 

How about:
```suggestion
    assert output.strip() == "0", f"Kernel tainted: {output}"
```

sairon: That won't work, the thing is the return type is a list of strings. This time I actually checked the test doesn't fail :see_no_evil:

agners: Hm, I see, and `"\n".join(output)` won't add a newline if there is only one element. Yeah make sense.

agners (approved):

PR #3821, opened 2025-01-22
Change cadence i2c driver from module to built-in driver

michalsimek:
On Kria KD240 slg7xl45106 device is handling reset for USB hub which is providing access to SD card (USB/SD converter). Access to this device is done via i2c which needs to be also enabled in the kernel as built-in driver not as module when rootfs is mounted.

I have tested 14.2.rc2 version and this one is also missing. I pointed via grub to DT without usb hub configuration and it boots properly. 

## Summary by CodeRabbit

- **New Features**
	- Added support for I2C Cadence driver in the kernel configuration, enabling enhanced I2C device compatibility.
	- Enabled GPIO support for Zynq devices in the kernel configuration.


sairon (commented):
I will consider pulling it in to final 14.2 release, however, have you verified this is really all it's needed for that platform? It'd be pointless without that.

michalsimek:
> I will consider pulling it in to final 14.2 release, however, have you verified this is really all it's needed for that platform? It'd be pointless without that.

I have removed reset part inside usb DWC3 node and try to boot that system and things booted properly. 
I have no idea how you are building kernel with all that infrastructure around, but if you tell me how to get 6.6.73 kernel source code out of buildroot (or whatever way you use) I can build it myself and try to be 100% confident.
I have already .config (which I got from kv260 platform via /proc/config.gz) and then load it externally via grub is easy.

sairon:
>  I have no idea how you are building kernel with all that infrastructure around, but if you tell me how to get 6.6.73 kernel source code out of buildroot (or whatever way you use) I can build it myself and try to be 100% confident.

I think there's not an easy way to only build the kernel without building the toolchain, even if you wanted just to generate the kernel's dotconfig in the same way as Buildroot does. On the other hand, the build environment is self-contained in the helper environment, so to build the kernel, run:
```
./scripts/enter.sh make -C buildroot O=/build/output_generic_aarch64/ BR2_EXTERNAL=/build/buildroot-external generic_aarch64_defconfig
./scripts/enter.sh make -C buildroot O=/build/output_generic_aarch64/ BR2_EXTERNAL=/build/buildroot-external linux
```
Alternatively, I can run an on-demand build here in the CI, or send you any of the files built/generated locally on my machine (which would be a bit faster).

michalsimek:
[kd240-bootlog.txt](https://github.com/user-attachments/files/18510219/kd240-bootlog.txt)

Just in case if you are interested how that bootlog looks like on v6.6 (with origin kernel modules on rootfs and v6.12 (without kernel modules on rootfs).

PR #3854, opened 2025-02-03
Remove USB stack patches working around obsoleted Z-Wave devices issues

sairon:
In #3224 we've introduced a patch reverting some changes in the USB stack that was supposed to work around issues with some USB devices. Later discussions revealed these devices are obsoleted by the manufacturer and there is no official way of fixing those in newer Linux kernels. However, carrying the patches makes us diverge from upstream and can eventually trigger other problems not present upstream we'll have to handle.

Drop these patches now to be part of the upcoming OS 15 release, rather than needing to drop them later in any of the patch revisions later.

## Summary by CodeRabbit

- **Bug Fixes**
  - Reverted recent USB subsystem modifications to restore reliable device recognition and initialization.
  - Improved high-speed USB operation for more consistent connectivity.
  - Enhanced support for connected devices, ensuring smoother performance (including for systems using the Z-Wave.me USB stick).


agners (approved):

HomeAssistant-Steve:
What defines a "Obsulete z-wave device"? which ones are they, is there a list somewhere?

PR #3862, opened 2025-02-05
Fix order of patches for Yellow

sairon:
After #3856, we need to change the patch order also for Yellow.


agners (approved):
Hm, I see this makes sure that the Raspberry Pi patches get applied first, which is important to make sure the IPv6 patches get applied in the right order. Make sense :+1: .

PR #3865, opened 2025-02-08
add dnsmasq for NetworkManager shared network

jkt628:
required by https://github.com/home-assistant/supervisor/pull/5577


agners:
Hm, ok so there is more than it than just Supervisor changes :cry:. It should be fairly low-risk to add `dnsmasq`. What is a bit concern is bloating. We already now struggle to fit things for `generic-x86-64` :cry: .

We certainly should check that there is no init scripts or systemd units deployed. In a quick look the `dhcpd` package excludes init scripts when NetworkManager is selected, but I don't see this for `dnsmasq` :thinking:

jkt628:
> Hm, ok so there is more than it than just Supervisor changes 😢. It should be fairly low-risk to add `dnsmasq`. What is a bit concern is bloating. We already now struggle to fit things for `generic-x86-64` 😢 .

there are also https://github.com/home-assistant/cli/pull/533 and https://github.com/home-assistant/home-assistant.io/pull/37140.  (i started on frontend yesterday but, even though i've coded for 50 years, i have never done a front end so i'm having to learn some new tricks.  at least retirement is not being wasted!)

can you provide some parameters around "fit" for `generic_x86_64`?  i have a system arriving today for testing [custom HAOS](https://github.com/jkt628/operating-system/blob/BUILD-LOCAL/BUILD-LOCAL.md) so i can hopefully address this concern.

> We certainly should check that there is no init scripts or systemd units deployed. In a quick look the `dhcpd` package excludes init scripts when NetworkManager is selected, but I don't see this for `dnsmasq` 🤔

on `rpi4_64` the only references to dnsmasq are apparmor, dbus and NetworkManager.  it runs only by NetworkManager, this taken from `ps axf`:

```text
    506 ?        Ssl    0:59 /usr/sbin/NetworkManager --no-daemon
   9279 ?        S      0:00  \_ /usr/sbin/dnsmasq --conf-file=/dev/null --no-hosts --keep-in-foreground --bind-interfac
```

i should know more about `generic_x86_64` soon.

buildroot-external/configs/generic_x86_64_defconfig
```diff
@@ -126,6 +126,8 @@ BR2_PACKAGE_IPTABLES_NFTABLES=y
 BR2_PACKAGE_IPTABLES_NFTABLES_DEFAULT=y
 BR2_PACKAGE_NETWORK_MANAGER=y
 BR2_PACKAGE_NETWORK_MANAGER_CLI=y
+BR2_PACKAGE_DNSMASQ=y
```
sairon, line 129: I see this was added manually, ideally `make ... savedefconfig` result of the modified config (e.g. using `menuconfig`) should be committed to the repo. The correctly formatted defconfig would have the changes around these lines:

```diff
diff --git a/buildroot-external/configs/generic_x86_64_defconfig b/buildroot-external/configs/generic_x86_64_defconfig
index f67c710da..49c0380e9 100644
--- a/buildroot-external/configs/generic_x86_64_defconfig
+++ b/buildroot-external/configs/generic_x86_64_defconfig
@@ -119,6 +119,8 @@ BR2_PACKAGE_BLUEZ5_UTILS_PLUGINS_AUDIO=y
 BR2_PACKAGE_BLUEZ5_UTILS_PLUGINS_HID=y
 BR2_PACKAGE_DHCP=y
 BR2_PACKAGE_DHCP_CLIENT=y
+BR2_PACKAGE_DNSMASQ=y
+# BR2_PACKAGE_DNSMASQ_TFTP is not set
 BR2_PACKAGE_DROPBEAR=y
 # BR2_PACKAGE_IFUPDOWN_SCRIPTS is not set
 BR2_PACKAGE_IPROUTE2=y
```

buildroot-external/configs/generic_x86_64_defconfig
```diff
@@ -126,6 +126,8 @@ BR2_PACKAGE_IPTABLES_NFTABLES=y
...
+BR2_PACKAGE_DNSMASQ_TFTP=n
```
sairon, line 130: This is not correct, as you can see above, disabling should be done through the `# ... is not set` stanza.

sairon (changes requested):
I guess you added only the platforms you were able to test, however, the options should be added to all the board configs pre-emptively, as there's AFAIK no mechanism that enables the functionality depending on the board in other PRs.

Regarding the concerns @agners raised - the binary itself is ~400 KiB on x86 before FS compression - not great, not terrible. Install seems to be sane and install only the binary and harmless stuff:

```

>>> dnsmasq 2.90 Installing to target
GIT_DIR=. PATH="/build/output_generic_x86_64/host/bin:/build/output_generic_x86_64/host/sbin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin" CC="/build/output_generic_x86_64/host/bin/x86_64-buildroot-linux-gnu-gcc" /usr/bin/make -j33 -C /build/output_generic_x86_64/build/dnsmasq-2.90 COPTS="-DNO_TFTP -DHAVE_DBUS" PREFIX=/usr CFLAGS="-D_LARGEFILE_SOURCE -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64  -O2 -g0 -D_FORTIFY_SOURCE=1" DESTDIR=/build/output_generic_x86_64/target LDFLAGS="" LIBS+= install
make[1]: Entering directory '/build/output_generic_x86_64/build/dnsmasq-2.90'
install -d /build/output_generic_x86_64/target/usr/sbin
install -d /build/output_generic_x86_64/target/usr/share/man/man8
install -m 644 man/dnsmasq.8 /build/output_generic_x86_64/target/usr/share/man/man8 
install -m 755 src/dnsmasq /build/output_generic_x86_64/target/usr/sbin
make[2]: Entering directory '/build/output_generic_x86_64/build/dnsmasq-2.90/src'
make[2]: 'dnsmasq' is up to date.
make[2]: Leaving directory '/build/output_generic_x86_64/build/dnsmasq-2.90/src'
make[1]: Leaving directory '/build/output_generic_x86_64/build/dnsmasq-2.90'
mkdir -p /build/output_generic_x86_64/target/var/lib/misc/
/usr/bin/install -m 0644 -D /build/output_generic_x86_64/build/dnsmasq-2.90/dbus/dnsmasq.conf /build/output_generic_x86_64/target/etc/dbus-1/system.d/dnsmasq.conf
```

What I wonder (but would have to be probably tested), would the DNS resolving, mDNS and whatnot work properly also for the clients? We use fairly exotic setup for that, which might be different from what NM expects. I think @agners is more knowledgeable in this area than me :innocent:

jkt628:
> I guess you added only the platforms you were able to test, 

correct, that is in the commit message.

> however, the options should be added to all the board configs pre-emptively

i wondered about that.  i'll update all the platforms and commit message, too.

jkt628:
> i should know more about `generic_x86_64` soon.

initial report: not working, 😢 .  i see the AP on my desktop and phone but cannot connect.  i found this in the journal:

```text
Feb 12 13:53:26 haos-x86-64 NetworkManager[423]: <warn>  [1739368406.3078] iptables: command /usr/sbin/iptables failed: Child process exited with code 1
Feb 12 13:53:26 haos-x86-64 NetworkManager[423]: <warn>  [1739368406.3132] iptables: command /usr/sbin/iptables failed: Child process exited with code 1
Feb 12 13:53:26 haos-x86-64 NetworkManager[423]: <warn>  [1739368406.3189] iptables: command /usr/sbin/iptables failed: Child process exited with code 1
```

but do not see similar on rpi4_64.

digging in....

jkt628:
i upgraded wpa_supplicant and my linux laptop and android phone work properly.

jkt628:
@agners NetworkManager creates a rule that prevents addons running on docker, like frigate-nvr, from connecting to cameras on the private network.

```text
table ip nm-shared-wlp3s0 { # handle 10
        chain nat_postrouting { # handle 1
                type nat hook postrouting priority srcnat; policy accept;
                ip saddr 10.42.0.0/24 ip daddr != 10.42.0.0/24 masquerade # handle 2
        }

        chain filter_forward { # handle 3
                type filter hook forward priority filter; policy accept;
                ip daddr 10.42.0.0/24 oifname "wlp3s0" ct state { established, related } accept # handle 5
                ip saddr 10.42.0.0/24 iifname "wlp3s0" accept # handle 6
                iifname "wlp3s0" oifname "wlp3s0" accept # handle 7
                iifname "wlp3s0" reject # handle 8
                oifname "wlp3s0" reject # handle 9
#             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^  this
        }
}
```

it is easily eliminated with [nft](https://github.com/buildroot/buildroot/tree/5f085911923983e9b8668017c50a44e8b5f0153f/package/nftables), see below, but

* that adds yet another package.
* basic haos has no way to import to [NetworkManager/dispatcher.d](https://people.freedesktop.org/~lkundrak/nm-docs/NetworkManager.html).
 
i don't mind doing the work to import additional configuration via `ha os import` but i wonder if this is the best approach.

```bash
#!/usr/bin/env bash

INTERFACE="$1"
EVENT="$2"

case "$EVENT" in
	up)
		iptables -P FORWARD ACCEPT
		nft -a list ruleset |& sed -n '/oifname "$INTERFACE" reject/{s/.*#//;p;q}' | xargs -n1 nft delete rule "nm-shared-$INTERFACE" filter_forward
		;;
esac
```

jkt628:
with a necessary upgrade to wpa_supplicant this [Practical Note](https://github.com/home-assistant/home-assistant.io/pull/37140/commits/7c22da70daca3b75cb4b960f77020ece58a0a06a) is spot on.

agners:
> @agners NetworkManager creates a rule that prevents addons running on docker, like frigate-nvr, from connecting to cameras on the private network.

Ugh yeah that sucks. I wonder if the firewalling can be influenced? I think rather than add nft and scripts which change the firewall settings, I'd rather just patch NetworkManager to do what we want (e.g. change how it inserts firewall rules).

From the NetworkManager's manual it seems that there are different firewall backends: https://networkmanager.pages.freedesktop.org/NetworkManager/NetworkManager/NetworkManager.conf.html#id-1.2.3.7

Maybe just setting it to `none` works for our use case? I guess it will route things transparently, even from the external interface, but maybe that is ok? :thinking:

jkt628:
@agners i tried `firewall-backend iptables` and see a new table created with a similar problematic rule, but it fails, probably because it was not actually linked into PREROUTING.

SentryCode (approved):

PR #3877, opened 2025-02-13
Use EROFS LZMA compression of firmware on x86 instead of per-file ZSTD

sairon:
Instead of using per-file ZSTD compression, compress firmware (and newly also kernel modules) using LZMA within EROFS image. LZMA was picked because ZSTD support in EROFS is still experimental and due to some limitations in the implementation, the compression takes significantly more time.

This change gives us more control over compression of the files and with the proposed settings, saves a bit of the space (~10 MiB) in the resulting image. In theory, even higher savings could be achieved through compressing other runtime binaries, but this would need to be thoroughly tested whether it doesn't have any detrimental effects. For firmware and modules, the overhead should be minimal, as they are usually touched only once per boot and are rather small before decompression.

* buildroot 74994c4f32...92fab35fed (6):
  > fs/erofs: add custom compression option with optional compress-hints file
  > package/erofs-utils: bump to version 1.8.5
  > package/erofs-utils: bump to version 1.8.3
  > package/erofs-utils: bump to version 1.8.2
  > package/erofs-utils: bump to version 1.8.1
  > package/erofs-utils: add libdeflate and zlib optional dependencies

## Summary by CodeRabbit

- **Chores**
  - Updated the core build reference to ensure a consistent and current system state.

- **New Features**
  - Introduced enhanced filesystem compression with custom algorithm settings.
  - Enabled LZMA support within the advanced filesystem.
  - Added updated compression guidelines to optimize key directory handling.


sairon:
Buildroot PR: https://github.com/home-assistant/buildroot/pull/75

agners (approved):
LGTM!

Issue #968, opened 2020-11-07
How can i increase my swap file size?


basdejong95:
**HassOS release with the issue:**

HassOS 4.15

**Supervisor logs:**

**Journal logs:**

**Kernel logs:**

**Description of problem:**

Hi everybody,

I am using home assistant on a RPI 4 with 1GB of ram. My swap file is only 230mb and fills up pretty fast, in the last week or so i had home assistant crashing a lot and i noticed my swap file is almost pretty full. While reading about this there are pretty clear instructions on the internet on how to increase the swap file on a standard RPI, but i cannot find it anywhere for HASSOS. Does anybody know a command or a setting i can change to increase this? I have a 128GB SD card which is A2 class so i dont mind creating a swap file of 1-2GB.

Thank you!

agners:
Currently Home Assistant OS does not really support that use case. This is also taking a strain on your SD card as swap usually generates quite a bit of read and writes. Since we already have users which have problems with SD cards wearing out, in a lot of cases using swap on SD card does not make sense.

By default HAOS configures a part of the RAM as compressed swap storage (using zram). This helps quite a bit for boards with low memory.

That said, when you access HAOS directly (via SSH port 22222), you can create a swapfile and enable it as swap, but the swap will not be automatically re-enabled after a reboot:
```
# fallocate -l 4G /mnt/data/test.swap
# mkswap /mnt/data/test.swap
# chmod 0600 /mnt/data/test.swap
# swapon /mnt/data/test.swap
```

bachy:
hi,
my hassos lives in a rpi3b+ with a 120Go ssd connected via usb
i moved the data partition on the ssd
my memory is at 80% and swap at 100%
i don't have so much devices and entities (discovered HA 2 weeks ago)
top says that homeassistant is consuming 31%, mysqld 16%, systemd-journald 5%, node-red 5% and supervisor 2.5%
i succesfully added a 4G swap file as describe above, i'll see if it prevent my instance to freeze
is there any way to make it permanent ? i mean active on reboot
i saw that there is nothing on /etc/fstab
i know nothing about buildroot
thanks

Jpsy:
Do yourself a favor and switch over to an RPi4 with more than 1GB of RAM. You will only waste time with you trials on 1GB. The difference that more RAM makes is overwhelming.

bachy:
[...]

PR #3883, opened 2025-02-19
Allow overriding sysctl parameters via /etc/sysctl.d files

sairon:
Relocate current content of /etc/sysctl.d to /usr/lib and make the /etc folder writable via a bind mount.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new configuration setup that ensures the kernel's persistent system settings are reliably applied during system startup. This enhancement helps maintain consistent system behavior by correctly managing the sequence of configuration application.


sairon:
One of the use cases can also be configuration of swappiness discussed in #968 (via `/etc/sysctl.d/15-vm.conf`).

agners (approved):

PR #3885, opened 2025-02-20
Use auditd to process AppArmor/audit logs

sairon:
Use auditd so logs from AppArmor and other audit events are processed by that instead of printed to the Systemd journal. This will reduce the log spam from BPF usually present in host logs and still preserve the audit logs for debugging.

The default configs seems to be sane for our purpose, rotating up to 5 files of 8MiB each. The difference is that /var/log/audit will be now on tmpfs but given how AppArmor is used on typical HA setup, we don't need to preserve the logs over reboots.

## Summary by CodeRabbit

- **New Features**
	- Enabled audit functionality to enhance security monitoring and event logging across multiple platforms without impacting existing configurations.


sairon:
Once https://github.com/home-assistant/supervisor/pull/5640 is in stable, we can add a test checking that no AA denied events are generated when the Supervisor tests end. Currently there's plenty of such errors for mkdir attempts for the Python bytecode.

agners:
> Once [home-assistant/supervisor#5640](https://github.com/home-assistant/supervisor/pull/5640) is in stable, we can add a test checking that no AA denied events are generated when the Supervisor tests end. Currently there's plenty of such errors for mkdir attempts for the Python bytecode.

Uh yeah I like that idea! This way we prevent that at least by default our containers don't attempt something we don't intend them to do.

PR #3887, opened 2025-02-20
Remove unnecessary GRUB userspace tools and other files

sairon:
Because the OTA hooks interact with GRUB environment using grub-editenv, we have BR2_TARGET_GRUB2_INSTALL_TOOLS enabled. However, that brings massive bloat of files that are never used in HAOS, as it also installs many other binaries, GRUB modules and translations.

As it's not possible to configure what gets installed in grub2 package, remove the undesired files in the post-build function. This brings savings of ~8.5MB of space in the root partition.

## Summary by CodeRabbit

- **Chores**
	- Improved system housekeeping by removing unnecessary GRUB components to streamline the environment and reduce clutter.

### DIFF
--- a/buildroot-external/scripts/rootfs-layer.sh
+++ b/buildroot-external/scripts/rootfs-layer.sh
@@ -26,6 +26,15 @@ function fix_rootfs() {
 
     # Use systemd-resolved for Host OS resolve
     sed -i '/^hosts:/ {/resolve/! s/files/resolve [!UNAVAIL=return] files/}' "${TARGET_DIR}/etc/nsswitch.conf"
+
+    # Remove unnecessary grub userspace tools, config, modules and translations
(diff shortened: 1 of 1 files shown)


agners (approved):
Nice find!

I am fine doing this as a HAOS specific post-rootfs task, but ideally we'd make the GRUB2 package support installing only the tools we need. Did you look into extending the package configurability?

sairon:
> I am fine doing this as a HAOS specific post-rootfs task, but ideally we'd make the GRUB2 package support installing only the tools we need. Did you look into extending the package configurability?

I did - it's currently handled by generic `make install` target of GRUB. There are no toggles for changing what's installed upstream. Without that, we'd have to implement our own install handling in Buildroot makefile, which is something too specific and IMO hardly upstreamable too.

PR #3888, opened 2025-02-20
Remove all info pages in post-build script

sairon:
As we don't have the info utility in HAOS, it's worthless to preserve info pages. While there are currently some files in /share/info (coming from GRUB2 tools install), /usr/share/info was added pre-emptively.

## Summary by CodeRabbit

- **Chores**
	- Enhanced system cleanup to remove unnecessary documentation files, ensuring a leaner deployment and optimized resource usage.


agners (approved):
LGTM

PR #3891, opened 2025-02-24
Fix RAUC tryboot handler set-state idempotency, add more checks

sairon:
When RPi is booted in the tryboot state and the set-state operation is called for the second time, the tryboot files don't exists anymore and the handler exits with an error code, printing an error in the Supervisor logs. Fix handling of this case and add few more checks to make the handler a bit more robust/traceable.

## Summary by CodeRabbit

- **Bug Fixes**
	- Refined the logic for system startup to ensure a smooth boot process.
	- Added robust checks for critical startup files to prevent proceeding with invalid states, enhancing overall boot reliability.


agners (approved):
LGTM

PR #3913, opened 2025-03-05
Update RPi firmware to fix boot with 2025-02-11 bootloader

sairon:
* buildroot 7d5c3b5e70...b079a02a9a (1):
  > package/rpi-firmware: bump to 1.20250305 with updated firmware binaries

Fixes #3911

## Summary by CodeRabbit

- **Chores**
  - Updated an external dependency to its latest stable version, ensuring the project tracks current improvements and maintains overall system stability.

PR #3914, opened 2025-03-05
Bump os-agent to v1.7.2

sairon:
Update to latest version that fixes start order in haos-agent.service. Without that, OS Agent reports incorrect swappiness after boot.

## Summary by CodeRabbit

- **Chores**
	- Updated an underlying subcomponent to its latest state.
	- Upgraded the OS agent version from 1.7.1 to 1.7.2 for streamlined performance.

### DIFF
--- a/buildroot-external/package/os-agent/os-agent.mk
+++ b/buildroot-external/package/os-agent/os-agent.mk
@@ -4,7 +4,7 @@
 #
 ################################################################################
 
-OS_AGENT_VERSION = 1.7.1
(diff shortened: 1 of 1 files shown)


agners (approved):

agners:
Actually, this one seems to be based on PR #3913. So you probably have to rebase before merging.

sairon:
I noticed when I started merging the rest too. But seems that GH was able to do its squash-merge without complaining, so I left it like this.

PR #3878, opened 2025-02-14
Update U-Boot to 2025.01

sairon:
Update U-Boot to 2025.01 and refresh all patches. Details in individual commits.

## Summary by CodeRabbit

- **New Features**
  - Added support for new boards (ODROID-M1S, NabuCasa Green) and expanded boot options on Raspberry Pi, including NVMe boot and a new command to load file content into environment variables.

- **Improvements**
  - Enhanced power management and eMMC performance with optimized boot sequences and improved regulator controls.
  - Updated LED indicators and reset controller functions for clearer status display.
  - Upgraded U-Boot to version 2025.01 across platforms for better compatibility and performance.

- **Bug Fixes**
  - Resolved issues with Ethernet configuration, PCIe/NVMe address translation, and boot sequence handling for more reliable operation.

PR #3946, opened 2025-03-18
Move ODROID-N2/C2/C4 U-Boot patch to correct folder to fix eMMC issues

### LINKED ISSUE #3942: OS 15 on ODROID-N2 unable to boot from Foresee 256 GB eMMC
> Yes, gets stuck in U-Boot, with a screen just like in this comment: https://github.com/home-assistant/operating-system/issues/3227#issuecomment-2018130825 ("no mmc device at slot 2").
> I'm using a Foresee 256 GB module. 

 _Originally posted by @ginkage in [#3938](https://github.com/home-assistant/operating-system/issues/3938#issuecomment-2731948367)_

sairon:
As pointed out by @ginkage, in #3878 the eMMC patch was moved to a wrong path. Move it to uboot subfolder so it's correctly applied.

Fixes #3942


frenck (approved):

sairon:
@vb7445 No, 15.0 stays unchanged and if your eMMC was affected there would be no difference in installing it then or now. 15.1.dev20250318 and newer are fine. Eventually 15.1 will be almost same as this dev, minus the suffix, so you can as well update to it with `ha os update --version 15.1.dev20250318`.

PR #3957, opened 2025-03-19
Enable amdgpu SI and CIK support for x86

sairon:
It seems that kernel 6.12 handles device probing less gracefully when these options are not enabled and causes crash on some AMD SoCs, e.g.:

*ERROR* Invalid callback to read register 0x58184

Link: https://gitlab.freedesktop.org/drm/amd/-/issues/4050


sairon:
Running a test build here: <https://github.com/home-assistant/operating-system/actions/runs/13952565640>

sairon:
@superm1 I don't have any feedback on this build yet, so I can't guarantee someone will pick the other build for testing. But I will have an HP t620 for testing available hopefully early next week, I will check it and let you know on GitLab.

frenck (approved):

raveit65:
No idea if this is useful.

superm1:
> No idea if this is useful. 

This _looks_ like a different problem now, especially if that console is not "frozen".  Could you capture the journal?

raveit65:
Output of journalctl -b0 -r
https://www.dropbox.com/scl/fi/0islcthqul5901r8r7m79/journal.txt?rlkey=g3l4702ucyk67011iu83e6k0i&dl=0

superm1:
I don't see anything pointing at amdgpu anymore, I think this is a secondary problem now.  Maybe the fact that the watchdog timer isn't working for some reason now?

peat-puck:
Thanks for the explanation and great work! I understand that there are 2 settings that need to be changed in kernel.config but since I am running HAOS I think I cannot edit kernel files. Should I be patient until a new OS version is available or is there a way to edit the kernel.config file anyway?

raveit65:
@sairon 
I tested the usb stick and your test-build with my notebook (AMD Ryzen 5 5600G (integrated gpu)) 
Here the boot stops after a while but it doesn't restart.
During the boot systemd wasn't able to mount the partition `hassio-data` and `Waiting for the Home Assistant CLI to be ready` is displayed at the end of the boot.
So the amdgpu driver seems to work on modern systems.

Not sure if this the right place to post this information but the t620 topic is locked.

sairon:
I just got the t620 for testing, so I could have a better look into the issues reported. At this point I can only say the original breakage is indeed fixed by enabling the driver options and the `amdgpu` (and HAOS in general) now works correctly on this device. The `it87_wdt` error is not a regression, as it's been present in 14.2 as well and it reportedly didn't cause any issues, but I'll look at it in detail to understand it a bit better later.

Anyway, the issues @raveit65 is seeing are unrelated to the matter discussed here. Please, create a separate issue for those. If there are more, create separate issues for different hardware.

For the patch from @superm1, I'll verify it tomorrow and report back in GitLab, as I promised before.

Thijs74:
Just for info / confirmation T620:
   14.2 Slot A has been updated to 15.0
   Not booting up properly
   Booted up from Slot  B (14.2)
   Manually applied the update 15.1.RC1 to slot A ( [https://developers.home-assistant.io/docs/operating-system/update-system/](url) )
   System is booting up from Slot A without issues

regepower:
Os 15.1 has same problem than 15.0 with Futro S520 / AMD G-Series GX-212ZC processor.

regepower:
How can I edit this in grub config? Must I boot another Linux from USB? Sorry I'm a noob on Linux.

Mikrofarad:
> How can I edit this in grub config? Must I boot another Linux from USB? Sorry I'm a noob on Linux.

So am I.
I did connect a keyboard and a monitor to my Thin Client and after power there is a menu with four entries from GRUB. Now you have to select (up and down arrows on the keyboard) the correct (A or B) boot slot (not the last two) and then press e (for edit). Then you can modify this entry and by key F10 the boot process continues with your new settings. But the changes are only temporarily 🥴 Unfortunately I cannot find cmdline.txt. This file is also read during boot process (via the GRUB settings).
HTH 👍

regepower:
Thanks a lot 💪

Mikrofarad:
I contact you right now via Discord...

PR #3996, opened 2025-04-02
change suffix version syntax

pepijndevos:
Very minor detail but `.dev` sorts higher than the release which makes version checks do the wrong thing. `~dev` seems to be the accepted syntax for this.

```
$ printf '15.1\n15.1.dev0\n15.0\n15.0.dev0' | sort -V
15.0
15.0.dev0
15.1
15.1.dev0
$ printf '15.1\n15.1~dev0\n15.0\n15.0~dev0' | sort -V
15.0~dev0
15.0
15.1~dev0
15.1
```

## Summary by CodeRabbit

- **New Features**
  - **Updated Version Display:** Version information now uses a distinct tilde separator when additional details are included. This change enhances clarity by clearly differentiating primary releases from those with extra identifiers, ensuring users can easily interpret and compare version details across updates.

### DIFF
--- a/buildroot-external/scripts/name.sh
+++ b/buildroot-external/scripts/name.sh
@@ -16,7 +16,7 @@ function hassos_version() {
     if [ -z "${VERSION_SUFFIX}" ]; then
         echo "${VERSION_MAJOR}.${VERSION_MINOR}"
     else
-        echo "${VERSION_MAJOR}.${VERSION_MINOR}.${VERSION_SUFFIX}"
+        echo "${VERSION_MAJOR}.${VERSION_MINOR}~${VERSION_SUFFIX}"
     fi
 }
 


sairon (commented):
We're using [awesomeversion](https://github.com/ludeeus/awesomeversion) in code that's comparing the versions, not lexicographic compare. In that the versions are actually compared correctly both for release versions and for notation we use for beta releases:
```python
>>> from awesomeversion import AwesomeVersion
>>> v1 = AwesomeVersion("15.1")
>>> v2 = AwesomeVersion("15.1.dev0")
>>> v3 = AwesomeVersion("15.1.rc1")
>>> v1 > v2
True
>>> v3 > v2
True
>>> v3 > v1
False
```

sairon:
Could you elaborate what "version checks do the wrong thing" exactly? Otherwise this PR is not the correct way to fix it.

pepijndevos:
I have an add-on that uses a bash script to check the OS version using the sort method above that behaves incorrectly. I've worked around it on my end already.

PR #4002, opened 2025-04-03
Enable PCI I/O virtualization in kernel

greyltc:
Setting `CONFIG_PCI_IOV=y` should enable SR-IOV functionality for a few NIC drivers, BNXT for example:

https://github.com/torvalds/linux/blob/a2cc6ff5ec8f91bc463fd3b0c26b61166a07eb11/drivers/net/ethernet/broadcom/Kconfig#L222

handy for those of us running HAOS via proxmox or the like.

## Summary by CodeRabbit

- **New Features**
  - Introduced support for Single Root I/O Virtualization (SR-IOV) to enhance networking capabilities for select interface cards without altering the existing setup.

### DIFF
--- a/buildroot-external/kernel/v6.12.y/device-support-pci.config
+++ b/buildroot-external/kernel/v6.12.y/device-support-pci.config
@@ -1,4 +1,4 @@
-# enable SR-IOV driver support for a few NICs
+# enable IOV (e.g. for SR-IOV support in some NICs)
 CONFIG_PCI_IOV=y
 
 CONFIG_IGB=y
--- a/buildroot-external/kernel/v6.12.y/device-support-pci.config
+++ b/buildroot-external/kernel/v6.12.y/device-support-pci.config
@@ -1,3 +1,6 @@
+# enable SR-IOV driver support for a few NICs
+CONFIG_PCI_IOV=y
+
 CONFIG_IGB=y
 CONFIG_I40E=m
 CONFIG_IGC=m


sairon (commented):
I might be missing some edge-cases of enabling this option, but AFAIK this should be only needed to be enabled on the host system, the guest doesn't care the device comes out of a SR-IOV virtualized HW. Or am I wrong? Quick search brought me to the [review comment](https://patchwork.kernel.org/project/kvm/patch/1235112888-9524-2-git-send-email-yu.zhao@intel.com/#24784) of the initial PC_IOV series where this was suggested to be added to the help text.

mabels:
> I might be missing some edge-cases of enabling this option, but AFAIK this should be only needed to be enabled on the host system, the guest doesn't care the device comes out of a SR-IOV virtualized HW. Or am I wrong? Quick search brought me to the [review comment](https://patchwork.kernel.org/project/kvm/patch/1235112888-9524-2-git-send-email-yu.zhao@intel.com/#24784) of the initial PC_IOV series where this was suggested to be added to the help text.

Whatever this means it's from 2009ish -- 
The guest should be able to use the passed devices --- and we should make it work as on all other OS.

HAOS also enabled that driver on some platforms, but on x86, it doesn't. 
I think x86 is due to proxmox, the common target for sr-iov, and it usually does not have any problems with RAM or disc size.

mabels:
Hi,

i just started to build my own haos and discovered that the problem is not existing for igb (1GBit) nics.
It's that the ixgbe (10GBit) driver is missing.

So this is not about sr-iov instead, it's about a missing nic, the intel 10Gbit driver -- I know nobody needs 10Gbit on HAOS, but -- my powerful home-lab might have 10Gbit.

greyltc:
> the guest doesn't care the device comes out of a SR-IOV virtualized HW

I know at least in some cases this is not true. For example, for Intel's X520 NICs, the host driver is `ixgbe` and the one the guest needs when it gets the sr-iov slice is `ixgbevf`.

buildroot-external/kernel/v6.12.y/device-support-pci.config
```diff
@@ -1,3 +1,6 @@
+# enable SR-IOV driver support for a few NICs
```
sairon, line 1:
```suggestion
# enable IOV (e.g. for SR-IOV support in some NICs)
```

sairon (approved):
Sorry it took so long, I wanted to take a look to better understand what would be affected. In general I think it's fine, technically it doesn't only enable SR-IOV on NICs though but it affects more things (e.g. the `xe` driver, even though we don't have it enabled). I just updated the comments and will adjust the commit message, otherwise LGTM.

PR #4013, opened 2025-04-09
Disable amdgpu SI/CIK support and backport upstream patch to prevent crashes

### LINKED ISSUE #4012: amdgpu driver still crashes in 15.1 on some HW
> 15.1 works only if I add _modprobe.blacklist=amdgpu_ to the respective entry/line in the GRUB editor. So my problem has nothing to do with _libata.force=nolpm_.
Hardware: Fujitsu Thin Client Futro S920 (AMD GX 415GA, 4x1,5GHz, 8GB RAM)

_Originally posted by @Mikrofarad in https://github.com/home-assistant/operating-system/issues/3957#issuecomment-2776069727_

sairon:
As we enabled SI/CIK support to fix crashes on bunch of AMD SoCs in #3957, the amdgpu driver still crashes on some hardware. It seems to be some GX-217GA SoCs or even GX-425GA in Fujitsu thin clients (even though the same SoC is fine on T620).

To get to the state before update to kernel 6.12, disable amdgpu for these platforms again (as it couldn't work properly there before OS 15.0) and just backport the patch that fixes the crashes during probing when the driver isn't compiled with SI/CIK support.

Fixes #4012

## Summary by CodeRabbit

- **Chores**
  - Removed legacy support for older AMD GPU architectures (Southern Islands and Sea Islands) from the configuration.
  
- **Bug Fixes**
  - Enhanced error handling to display clear messages when attempting to initialize AMD GPU features that are no longer supported.


agners (approved):

Grey-Lancaster:
I have a t630 that was behaving very badly.  It seems normal now with 15.2

Grey-Lancaster:
It still seems to crash when I update esphome devices

PR #4016, opened 2025-04-10
Allow for fluctuation of MemTotal for automatic swapfile size

sairon:
As the reported MemTotal can fluctuate a bit on some systems, e.g. because the reserved memory changes between kernel version or other factors affect it like VRAM, the swap file can be recreated unnecessarily between boots. Allow for some fluctuation (up to +-32MB) before the swapfile is recreated.

This was a problem already before the recent haos-swapfile changes, however, before it checked if the existing swapfile isn't smaller than the desired value. If the MemTotal fluctuated there, the swapfile size eventually settled on the highest value seen and it wasn't recreated anymore. With this change, things should be stable even more.

## Summary by CodeRabbit

- **New Features**
	- Enhanced swap file management by introducing a tolerance for small size discrepancies.
	- Improved error handling during swap file verification to reduce unnecessary file recreations.


agners (approved):

felipecrs:
Sorry, I know this is off topic, but there's no better place to ask:

Is REST API the only way to control the swap in HAOS for now?

PR #4040, opened 2025-04-25
Make usage of top-level make easier, drop 'all' target

sairon:
Make it easier when using top-level make - proxy all possible commands to Buildroot make and only wrap build for individual target builds. This way it's still possible to run e.g. 'make ova' which would read the defconfig and run the build, while we can also use the top-level make in the same way as it's in vanilla Buildroot.

Target 'all' was dropped in favor of Buildroot 'make' without any arguments - as it's fairly pointless to run all builds sequentially. With the current 19 targets it would take about a day even on a decent hardware and the build artifacts would be lost in the process.

## Summary by CodeRabbit

- **Refactor**
	- Improved user feedback with colored terminal messages during build and configuration steps.
	- Replaced the previous default build target with a new `default` target for streamlined usage.
	- Enhanced handling of unknown targets with informative fallback messages.
	- Simplified help output and added explanations about fallback behavior.
	- Added warnings when attempting configuration with a non-empty output directory.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,9 +36,8 @@ default:
 	$(MAKE) -C $(BUILDROOT) O=$(O) BR2_EXTERNAL=$(BUILDROOT_EXTERNAL)
 
 $(TARGETS_CONFIG): %-config:
-	@if [ -d $(O) ] && [ "$$(ls -A $(O))" ]; then \
-		echo "$(COLOR_WARN)WARNING: Output directory '$(O)' is not empty!$(TERM_RESET)"; \
-		echo "         This may cause errors if it was previously used for a different target."; \
+	@if [ -f $(O)/.config ] && ! grep -q 'BR2_DEFCONFIG="$(DEFCONFIG_DIR)/$*_defconfig"' $(O)/.config; then \
+		echo "$(COLOR_WARN)WARNING: Output directory '$(O)' already contains files for another target!$(TERM_RESET)"; \
 		echo "         Before running build for a different target, run 'make distclean' first."; \
 		echo ""; \
 	fi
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -14,37 +14,46 @@ else
...
+default:
+	$(MAKE) -C $(BUILDROOT) O=$(O) BR2_EXTERNAL=$(BUILDROOT_EXTERNAL)
 
 $(TARGETS_CONFIG): %-config:
-	@echo "config $*"
+	@if [ -d $(O) ] && [ "$$(ls -A $(O))" ]; then \
+		echo "$(COLOR_WARN)WARNING: Output directory '$(O)' is not empty!$(TERM_RESET)"; \
+		echo "         This may cause errors if it was previously used for a different target."; \
+		echo "         Before running build for a different target, run 'make distclean' first."; \
+		echo ""; \
```
agners, line 43: Can we maybe check what target was used for a given output directory before? :thinking: 

With this I'll get a warning for every build, I typically use command like this:
```
./scripts/enter.sh make O=output_ova ova
```

sairon: I thought about it as a warning for newcomers (as you can also do e.g. `make ova_defconfig && make`) so if you knew what you were doing, you could ignore it. But yeah, why not. Adjusted the condition to check the `.config` before regenerating it.

agners: Nice, yeah that seems to work here.

One more nit: Can we maybe add a sleep of 10s or so and continue only then? Right now if you see that error it essentially is "too late", your build directory is already tainted with that other defconfig :thinking:

sairon: Sure, done as well.

agners (approved):
Nice :heart_eyes:

PR #4043, opened 2025-04-28
Allow to override update channel for build, use dev for dev builds

sairon:
Add input allowing to override the channel that's used for hassio image downloads and in runtime for Supervisor updates, building on the option added in #3618.

The new default is dev for dev builds, for GH releases keep using the stable channel both for releases and pre-releases (so we could catch any stable issues before beta is moved to stable).

To keep it DRY and idiomatic, create a new in-repo GH action for running commands in the build container.

## Summary by CodeRabbit

- **New Features**
	- Introduced a configurable release channel option for builds, allowing selection between default, stable, beta, and dev channels.
- **Chores**
	- Added a new GitHub Action to standardize running commands in the build container.
	- Refactored the build workflow to use the new action, simplifying and unifying build steps.

### DIFF
--- a/.github/actions/haos-builder-command/action.yml
+++ b/.github/actions/haos-builder-command/action.yml
@@ -1,0 +1,22 @@
+name: "Run command in HAOS build container"
+inputs:
+  image:
+    description: "HAOS builder image to use"
+    required: true
+  command:
+    description: "Command to run in the container"
+    required: true
+runs:
+  using: 'composite'
+  steps:
+  - name: "Run command in HAOS build container"
+    shell: bash
+    run: |
+      docker run --rm --privileged \
+        -e BUILDER_UID="$(id -u)" \
+        -e BUILDER_GID="$(id -g)" \
+        -v "${GITHUB_WORKSPACE}:/build" \
+        -v "/mnt/cache:/cache" \
+        -v "/mnt/output:/build/output" \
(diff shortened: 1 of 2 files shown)


.github/actions/haos-builder-command/action.yml
```diff
@@ -0,0 +1,22 @@
...
+    shell: bash
+    run: |
+      docker run --rm --privileged \
+        -e BUILDER_UID="$(id -u)" \
+        -e BUILDER_GID="$(id -g)" \
+        -v "${GITHUB_WORKSPACE}:/build" \
+        -v "/mnt/cache:/cache" \
+        -v "/mnt/output:/build/output" \
+        ${{ inputs.image }} \
+        ${{ inputs.command }}
```
agners, line 22: I wonder if we maybe could create a whole job which runs inside that container using the GitHub Action `container` syntax :thinking: 

https://docs.github.com/en/actions/writing-workflows/choosing-where-your-workflow-runs/running-jobs-in-a-container

Not sure if it works as all steps have then to run inside the container. I guess we could move some things inside the container (like the AWS upload, then we wouldn't need to install that every time), but some are probably easier handled outside (probably most other steps).

In any case, this is a step in the right direction. Moving to the native `container` support can be explored separately.

agners (approved):
LGTM!

PR #4064, opened 2025-05-14
Fix hassio channel switching for dev builds

sairon:
Fix build job to write config option for channel switching from #4043 to the actual config. As it was written to .config in the top-level build directory, it was never correctly applied.

## Summary by CodeRabbit

- **Chores**
  - Updated the build process to modify configuration settings in a different location, which may affect the build output. No user-facing features or functionality have changed.


agners (approved):
Hm, I see so the channel override didn't work. Nice catch, LGTM!

PR #4079, opened 2025-05-27
Reenable systemd-resolved stub resolver

agners:
Enable the systemd-resolved stub resolver and make it available on the hassio host network interface (172.30.32.1). This allows to use systemd-resolved directly from all containers.

Note that this makes /etc/resolv.conf point to the stub resolver running at 127.0.0.53 by default. This stub resolver isn't reachable from within containers. However, Docker does regnize this situation [1] and falls back to the alternate path at /run/systemd/resolve/resolv.conf, which is what /etc/resolv.conf is today. So this should not affect the initial /etc/resolv.conf in containers in practise.

This will however bind to port 53 and affect add-on potentially attempt to use that port. Add-ons should not bind to 127.0.0.53 or the hassio host network (172.30.32.1).

[1] https://github.com/moby/moby/blob/v28.0.4/libnetwork/internal/resolvconf/resolvconf_path.go#L51C32-L51C45

### DIFF
--- a/buildroot-external/rootfs-overlay/etc/systemd/resolved.conf
+++ b/buildroot-external/rootfs-overlay/etc/systemd/resolved.conf
@@ -20,6 +20,6 @@ DNSOverTLS=no
 #MulticastDNS=yes
 #LLMNR=yes
 #Cache=yes
-DNSStubListener=no
+#DNSStubListener=yes
 #ReadEtcHosts=yes
 #ResolveUnicastSingleLabel=no
--- a/buildroot-external/rootfs-overlay/etc/systemd/resolved.conf.d/hassio.conf
+++ b/buildroot-external/rootfs-overlay/etc/systemd/resolved.conf.d/hassio.conf
@@ -1,0 +1,2 @@
+[Resolve]
+DNSStubListenerExtra=172.30.32.1


agners:
As mentioned in https://github.com/home-assistant/operating-system/pull/1004/commits/3aa7b218309eef76bf32a6423bdbaacfae610833, some add-ons bind to port 53 UDP as well as TCP:

> There are Add-Ons which try to bind port 53 on all interfaces including
127.0.0.53. Disable the stub resolver to make them continue working. We
don't need the resolver currently anyway.

In a quick test, specifically Adguard explicitly tries to bind to 172.30.32.1, which fails with this change:

```
2025/05/27 14:54:31.336442 [info] dnsproxy: creating udp server socket addr=172.30.32.1:53
2025/05/27 14:54:31.336458 [info] dnsproxy: listening to udp addr=172.30.32.1:53
2025/05/27 14:54:31.336465 [info] dnsproxy: creating udp server socket addr=127.0.0.1:53
2025/05/27 14:54:31.336477 [info] dnsproxy: listening to udp addr=127.0.0.1:53
2025/05/27 14:54:31.336483 [info] dnsproxy: creating udp server socket addr=[::1]:53
2025/05/27 14:54:31.336494 [info] dnsproxy: listening to udp addr=[::1]:53
...
2025/05/27 14:54:31.336740 [info] dnsproxy: creating tcp server socket addr=172.30.32.1:53
2025/05/27 14:54:31.336769 [info] dnsproxy: warning: binding attempt=1 err="listen tcp 172.30.32.1:53: bind: address already in use"
```

This got introduced a while back in https://github.com/hassio-addons/addon-adguard-home/pull/168.

Furthermore the dnsmasq add-on tries to claim tcp port 53 too, which fails:

```
Can't start addon_core_dnsmasq: 500 Server Error for http+docker://localhost/v1.48/containers/9ed8c7922e5eeb1bbabdd228218b39ded02ec92c437fe42367b35211bce73112/start: Internal Server Error ("failed to set up container networking: driver failed programming external connectivity on endpoint addon_core_dnsmasq (241702f3d55491dcd8e7f9e398a6887aebaec43f11cd643b2a4c5bdd0c57ba14): failed to bind host port for 0.0.0.0:53:172.30.33.0:53/tcp: address already in use")
```

For UDP ports it seems that binding twice works, at least the dnsmasq add-on can be started with UDP port 53 enabled only. It does seem that dnsmasq then takes precedence though :cold_sweat: (nslookup homeassistant.local. 172.30.32.1 fails as seemingly dnsmasq processes the request instead of systemd-resolved.

agners:
Could we maybe have a veth on hassio specifically for systemd-resolved stub resolver? That way it could be a drop in replacement for plugin-dns :thinking: .

PR #4089, opened 2025-06-03
enable accelerated virtualization for x86_64

zoltan0907:
enabled virtualization allows custom addons to run accelerated vms on homeassistant via libvirtd for example

## Summary by CodeRabbit

- **New Features**
  - Enabled virtualization support in the kernel, including KVM support for both Intel and AMD platforms.
  - Added hardware virtualization extensions and enhanced device assignment capabilities for improved virtual machine performance.


agners:
Going down the VM path would be really a different path to make Home Assistant extensible. At this point, it would definitely be an additional feature. But to make it seemless, it would need Supervisor support, essentially besides supporting container based add-ons we could offer VM based add-ons. Since HAOS is often run as a VM itself, this would mean lots of user would use nested virtualization. Also, this would really be quite a extension to the current system.

Another use case could be to use Kata Containers. But honestly, I don't think that we really require the additional safety which we get from Kata Containers over regular containers.

What is your use case? Did you consider running HAOS in Proxmox, and whatever else you intend to run in VM alongside HAOS?

zoltan0907:
Hi.

I like the way of updating HA and OS in the same way via HAOS-Settings and i already played around with my own container-addons. I come from a rpi with haos and now a i have a mini-pc with haos. I found some addons which are interesting for me, for example https://www.home-assistant.io/integrations/unifi/ . In my home-network a had my access-points in one vlan and my iot-stuff in another including the ha-mini-pc. To get the addon working i had to configure an additional vlan with an ip in my home-network on haos. Apart from the additional static (or specific dhcp-config) now both networks are somehow connected. Yes i know, forwarding is enable on haos but direct routing between both networks is not possible without extra actions ... but both networks should separated at layer2. Therefor i thought it would be a good idea to use libvirt and bridge this vlan into a vm, in which the unifi-controller is running. And because i didn't found another way to build and load the modules without the corresponding `symbols`in the running kernel i tried to create the pr.

But i got your points and i understand that this extension leads more/many changes. For me it would be sufficient to connect one or more containers with a specific network- or vlan-interface via a separat bridge. I hope this would be possible with docker and maybe via haos settings and addon-configuration and without the need of doing manual tasks on the ha-host (nm-cli, etc).

A docker-compose example would be:
```
services:
  alpine:
    image: alpine
    tty: true
    networks:
      - predefined_vlan_bridge
networks:
  predefined_vlan_bridge:
    driver: bridge
    driver_opts:
      com.docker.network.bridge.name: vlan_bridge
    ipam:
      config:
        - subnet: 10.0.0.0/8
```
and the predefined bridge:
```
10: vlan: <BROADCAST,NOARP,UP,LOWER_UP> mtu 1500 qdisc noqueue master vlan_bridge state UNKNOWN mode DEFAULT group default qlen 1000
    link/ether 5a:c8:fa:a1:99:f5 brd ff:ff:ff:ff:ff:ff
13: vlan_bridge: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP mode DEFAULT group default qlen 1000
    link/ether 3e:05:e3:1b:ce:33 brd ff:ff:ff:ff:ff:ff
16: veth7e6ab5c@if2: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue master vlan_bridge state UP mode DEFAULT group default 
    link/ether 76:79:3d:06:59:da brd ff:ff:ff:ff:ff:ff link-netnsid 0
```
But therefor it has to be possible to setup the network via the addon-config i think.

Regards
Tino

PR #4091, opened 2025-06-05
Add Intel Cannon Lake pinctrl driver

cubemike:
This allows x86_64 targets to access their onboard gpio, which is necessary to control certain onboard devices.

For example, my ctl. CBx2 (a former chromebox with a Cannon Lake Intel Celeron) had non-functional bluetooth and a front SD card slot before I made this change. The devices were showing up in `/sys/kernel/debug/devices_deferred`. With the pinctrl driver (all I needed was `CONFIG_PINCTRL_CANNONLAKE`), the bluetooth driver is able to find the reset line of the bluetooth module, and the sdhci driver is able to find the CD line of the card slot. Both "just work" now. In addition to the Cannon Lake driver, I enabled all the default pinctrl drivers selected by default by the kernel.

I imagine I'm not the first person to have issues with missing pinctrl, it just wasn't obvious that pinctrl was the problem. I didn't have any indication of a real problem until I enabled dynamic debug in the kernel and enabled prints for gpiolib.c. That gave me these "No GPIO consumer reset found" messages in dmesg:

```
[    1.335935] sdhci-pci 0000:00:14.5: SDHCI controller found [8086:02f5] (rev 0)
[    1.337168] sdhci-pci 0000:00:14.5: using ACPI '\_SB.PCI0.SDXC' for 'cd' GPIO lookup
[    1.337194] sdhci-pci 0000:00:14.5: No GPIO consumer cd found
...
[    1.350399] NET: Registered PF_BLUETOOTH protocol family
[    1.351343] Bluetooth: HCI device and connection manager initialized
[    1.352288] Bluetooth: HCI socket layer initialized
[    1.353225] Bluetooth: L2CAP socket layer initialized
[    1.354159] Bluetooth: SCO socket layer initialized
[    1.356386] r8169 0000:01:00.0 eno0: renamed from eth0
[    1.399050] usb 1-10: using ACPI '\_SB.PCI0.XHCI.RHUB.HS10' for 'reset' GPIO lookup
[    1.399940] usb 1-10: No GPIO consumer reset found
[    1.401697] usbcore: registered new interface driver btusb
```

## Summary by CodeRabbit

- **New Features**
  - Enabled extensive pin control (pinctrl) support in the kernel for generic x86-64 boards, including core pinctrl functionality and a wide range of drivers for AMD, Intel, and various hardware platforms.
  - Improved hardware pin configuration and multiplexing capabilities by adding support for multiple pinctrl drivers as built-in or modules.
  - Added support for Cannon Lake platform pin control as a loadable kernel module.


buildroot-external/board/pc/generic-x86-64/kernel.config
```diff
@@ -162,3 +162,44 @@ CONFIG_BMP280=m
...
+CONFIG_PINCTRL_EMMITSBURG=m
+CONFIG_PINCTRL_GEMINILAKE=m
+CONFIG_PINCTRL_ICELAKE=m
+CONFIG_PINCTRL_JASPERLAKE=m
+CONFIG_PINCTRL_LAKEFIELD=m
+CONFIG_PINCTRL_LEWISBURG=m
+CONFIG_PINCTRL_METEORLAKE=m
+CONFIG_PINCTRL_METEORPOINT=m
+CONFIG_PINCTRL_SUNRISEPOINT=m
+CONFIG_PINCTRL_TIGERLAKE=m
```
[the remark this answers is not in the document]
cubemike: I disagree. These settings are taken verbatim from the current default configs of the upstream kernel, and the kernel maintainers probably have a better idea about what needs to be built in vs a module. Also if the pinctrl driver isn't ready at a particular driver's probe time, the driver will simply be deferred and reprobed once the pinctrl driver itself is probed. 

Happy to reconsider on human input.

buildroot-external/board/pc/generic-x86-64/kernel.config
```diff
@@ -162,3 +162,44 @@ CONFIG_BMP280=m
...
+CONFIG_PINCTRL_MCP23S08_SPI=m
+CONFIG_PINCTRL_MCP23S08=m
+CONFIG_PINCTRL_SX150X=y
+CONFIG_PINCTRL_CS42L43=m
+CONFIG_PINCTRL_MADERA=m
+CONFIG_PINCTRL_CS47L15=y
+CONFIG_PINCTRL_CS47L35=y
+CONFIG_PINCTRL_CS47L85=y
+CONFIG_PINCTRL_CS47L90=y
+CONFIG_PINCTRL_CS47L92=y
```
sairon: These have no effect, as they're bool options without any prompt. You can't enable them in the config without enabling the option which explicitly selects them (generally the MFD drivers here).

buildroot-external/board/pc/generic-x86-64/kernel.config
```diff
@@ -162,3 +162,44 @@ CONFIG_BMP280=m
...
+
+CONFIG_PINCTRL_BAYTRAIL=y
+CONFIG_PINCTRL_CHERRYVIEW=y
+CONFIG_PINCTRL_LYNXPOINT=m
+CONFIG_PINCTRL_INTEL=y
+CONFIG_PINCTRL_INTEL_PLATFORM=m
+CONFIG_PINCTRL_ALDERLAKE=m
+CONFIG_PINCTRL_BROXTON=m
```
sairon: There might be more like this but just Broxton was Smartphone-targeted line which was cancelled. So it's quite unlikely we'll benefit from enabling it.

cubemike:
Point taken. Updated with a reduced set of additional drivers to just cannon lake, which still works for me on hardware. 

To be honest I hadn’t quite realized the list of intel modules I enabled was for less mainline architectures, and presumably pinctrl functionality for most intel architectures is built into PINCTRL_INTEL. That explains why I was having trouble with my specific hardware. I appreciate the review.

PR #4105, opened 2025-06-17
Add Marvell 88E1XXX driver for x86-64

### LINKED ISSUE #4025: Marvell 88E1543(4L) not detected

I've built a new clean install of HassOS 15.1 on an ASRock C3758D4I-4L 8 core Atom board with a Marvell 88E1543(4L) 4 port NIC but none of the NICs are detected. If I plug a USB eth adapter in it gets detected straight away but the onboard NICs never show up.
I booted the same board with Ubuntu 20 live and it detected the NIC automatically.

This is the motherboard if it helps - https://www.asrockrack.com/general/productdetail.asp?Model=C3758D4I-4L#Specifications

15.1

No

### Hardware details

bare metal x64 on 8 core Atom C3758 in ASRock C3758D4I-4L motherboard

1.
2.
3.
...

```txt
..
```

```txt
..
```

### System information

_No response_

### Additional information

_No response_

sairon:
Add driver for Marvell PHYs, such as 88E1543(4L) on an ASRock C3758D4I-4L board. Adding it to x86 config only, as it seems it's not widely used anywhere else.

Fixes #4025

## Summary by CodeRabbit

- **New Features**
  - Enabled support for Marvell PHY network devices.
- **Style**
  - Improved organization of kernel configuration options with clearer comments.

### DIFF
--- a/buildroot-external/board/pc/generic-x86-64/kernel.config
+++ b/buildroot-external/board/pc/generic-x86-64/kernel.config
@@ -163,4 +163,8 @@ CONFIG_BMP280=m
 # Required for some PCIe devices such as ath12k
 CONFIG_IRQ_REMAP=y
 
+# Pin control support
 CONFIG_PINCTRL_CANNONLAKE=m
+
+# Network devices
+CONFIG_MARVELL_PHY=m


buildroot-external/board/pc/generic-x86-64/kernel.config
```diff
@@ -163,4 +163,8 @@ CONFIG_BMP280=m
 # Required for some PCIe devices such as ath12k
 CONFIG_IRQ_REMAP=y
 
+# Pin control support
 CONFIG_PINCTRL_CANNONLAKE=m
+
+# Network devices
+CONFIG_MARVELL_PHY=m
```
agners, line 170: I guess this are mostly PCIe devices? I'd suggest to add it to `buildroot-external/kernel/v6.12.y/device-support-pci.config`, we already have `CONFIG_BROADCOM_PHY=m` there. Just in case someone passthrough a NIC..

agners: Actually, it seems that this is a rather odd quad-gigabit PHY, and the MAC is built into the Atom SoC

image

So at least in this case, this is not related to PCIe at all. It could be that Marvell PHYs are required for PCIe Ethernet controllers, but I guess we can move it if that comes up. So fine by me as is.

agners (approved):

buildroot-external/board/pc/generic-x86-64/kernel.config
sairon: Yes, I reached the same conclusion. `device-support-pci.config` is incorrect, `device-support.config` too broad.

PR #4113, opened 2025-06-23
Fix flaky timeout after boot slot switching test

sairon:
Add timeout to expect call when waiting for the OS reboot after switching slots. While it never fails for me locally, it regularly breaks tests in GHA.

## Summary by CodeRabbit

- **Tests**
  - Increased the timeout for a console output check in the boot slot test to improve reliability.
  - Removed an unused import from the test file.

### DIFF
--- a/tests/smoke_test/test_os_update.py
+++ b/tests/smoke_test/test_os_update.py
@@ -3,7 +3,6 @@ import logging
 from time import sleep
 
 import pytest
-from labgrid.driver import ExecutionError
 
 _LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


agners:
> Add timeout to expect call when waiting for the OS reboot after switching slots. While it never fails for me locally, it regularly breaks tests in GHA.

Hm, so this will make the test fail then?

Do you know why we don't get that string in CI?

sairon:
> Hm, so this will make the test fail then?

This change _fixes_ the failing test.

> Do you know why we don't get that string in CI?

We do get it but for some reason, the `expect` fails almost immediately with `Timeout of 0.07 seconds exceeded` (the time changes between runs but it's always below second). IIUC it should wait indefinitely - and it does when I run the tests here - but in CI it doesn't seem so. With the explicit timeout it's fine (I ran the test in my fork).

agners:
> We do get it but for some reason, the `expect` fails almost immediately with `Timeout of 0.07 seconds exceeded` (the time changes between runs but it's always below second). IIUC it should wait indefinitely - and it does when I run the tests here - but in CI it doesn't seem so. With the explicit timeout it's fine (I ran the test in my fork).

Ah interesting, so basically the default timeout of expect appears to behave differently :thinking: . Weird, but I guess if the explicit one works, then why bother :shrug: .

agners (approved):
LGTM.

PR #4123, opened 2025-06-25
Enable parallel gzip for faster Docker pulls

agners:
Docker checks if unpigz is available, and if so uses it to unpack container layers with multiple CPU cores. This should make Docker pulls faster, especially on lower end hardware.


agners:
I found this when I was looking into the package maintainers readme at [PACKAGERS.md](https://github.com/moby/moby/blob/v28.3.0/project/PACKAGERS.md?plain=1#L132).

sairon (approved):
Nice find!

Maybe we could also consider using zstd for our containers? :thinking:

PR #4132, opened 2025-07-01
Do not disable RPi Bluetooth if Wi-Fi is not present

### LINKED ISSUE #2975: HA OS Raspberry Pi 4 BlueTooth issue on version 11.2

I'have updated HA OS on my Raspberry Pi 4 from version 11.1 to 11.2 and i've lost bluetooth functionality. If i downgrade to 11.1, everthing works fine. I've discovered that happened because i have disactivated my wifi in the /mnt/boot/config.txt file. 
In version 11.2 of HA OS, if i want the bluetooth, i need to activate wifi in the /mnt/boot/config.txt file. Is it possible to give the choice of activating or not the wifi, because i don't need it ?

rpi4-64  (Raspberry Pi 4/400 64-bit OS)

11.2

### Did you upgrade the Operating System.

Yes

1. update HA OS to 11.2
2. desactivate wifi in the /mnt/boot/config.txt
3. reboot

the bluetooth will work no longer

```txt
no
```

```txt
no
```

### System information

_No response_

### Additional information

_No response_

sairon:
This reverts commit eab18076ad7717ff4bce848b95fdb4e6c9f88f29.

This change was added in #2948 as a workaround for all-zero adapter appearing in the HA frontend (#2944). With changes implemented in [1], this is no longer needed, the only minor issue is that the ghost adapter still appears in hciconfig (and other utilities') output as reported in [2]. However, this should be less problematic than the Bluetooth being unavailable if WiFi is disabled through disable-wifi DT overlay, so let's start with removing the workaround.

Fixes #2975

[1] https://github.com/Bluetooth-Devices/bluetooth-adapters/pull/105
[2] https://github.com/raspberrypi/linux/issues/5756

## Summary by CodeRabbit

* **Chores**
  * Removed the system service that disabled the HCI UART kernel module when a specific WiFi/Bluetooth module was absent. This service will no longer be installed or managed in the system.


sairon:
Eventually this test can be used to check if the board is a no-WiFi one:
```
[ -f /sys/firmware/devicetree/base/chosen/rpi-boardrev-ext ] && test $(od -An -t u1 -N1 /sys/firmware/devicetree/base/chosen/rpi-boardrev-ext) -eq 64
```
But if we disable `hci_uart` moduled based on that, it could also break usage of other UART-based Bluetooth adapters. But maybe we don't need to care in HAOS? :thinking: Anyway leaving that out of this PR.

PR #4216, opened 2025-08-06
Prevent root from running the enter.sh helper script

### LINKED ISSUE #4214: Configure Error while trying to build latest operating system

I am trying to build the latest version of the operating system in order to test a kernel module that we developed. The build is done under Debian 12:
`Linux debian12 6.1.0-37-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.140-1 (2025-05-22) x86_64 GNU/Linux`

The command line used to run the build:
`sudo scripts/enter.sh make generic_x86_64`

The error I receive:
```
checking whether mknod can create fifo without root privileges... configure: error: in `/build/output/build/host-tar-1.35':
configure: error: you should not run configure as root (set FORCE_UNSAFE_CONFIGURE=1 in environment to bypass this check)
```

The error persists even with FORCE_UNSAFE_CONFIGURE set to 1. Most articles I could find online recommend to build buildroot without elevated privileges, but the home assistant build does require root privileges. Executing the same command as a regular user errors out right away:
`support/scripts/br2-external: line 52: /build/output/.br2-external.mk: Permission denied`

What I am missing?

16.2

No

### Hardware details

Debian 12 running under VMware virtual machine, itself running under Windows 10

1. Follow the steps from the developer documentation at https://developers.home-assistant.io/docs/operating-system/getting-started
2. `git clone https://github.com/home-assistant/operating-system/
cd operating-system/
git submodule update --init`
3. Checked that docker is installed and running properly
4. Checked that FORCE_UNSAFE_CONFIGURE is set to 1 : printenv FORCE_UNSAFE_CONFIGURE
5. Started the build with `sudo scripts/enter.sh make generic_x86_64`...

```txt
N/A
```

```txt
N/A
```

### System information

_No response_

### Additional information

_No response_

sairon:
Since configure doesn't like being ran as root, check in the enter.sh script that the user running it is not UID/GID 0. The script itself takes care of running what needs to be executed privileged with explicit sudo commands.

Fixes #4214

## Summary by CodeRabbit

* **Bug Fixes**
  * Added a check to prevent running the script as the root user or group, displaying an error message and exiting if detected.

### DIFF
--- a/scripts/enter.sh
+++ b/scripts/enter.sh
@@ -8,7 +8,7 @@ ARGS="$*"
 COMMAND="${ARGS:-bash}"
 
 if [ "$BUILDER_UID" -eq "0" ] || [ "$BUILDER_GID" == "0" ]; then
-  echo "ERROR: Please use a non-root with sudo rights for running this script."
+  echo "ERROR: Please run this script as a regular (non-root) user with sudo privileges."
   exit 1
 fi
 


scripts/enter.sh
```diff
@@ -7,6 +7,11 @@ CACHE_DIR="${CACHE_DIR:-$HOME/hassos-cache}"
 ARGS="$*"
 COMMAND="${ARGS:-bash}"
 
+if [ "$BUILDER_UID" -eq "0" ] || [ "$BUILDER_GID" == "0" ]; then
+  echo "ERROR: Please use a non-root with sudo rights for running this script."
```
agners: Reads a bit hard, maybe:

```suggestion
  echo "ERROR: Please run this script as a regular (non-root) user with sudo privileges."
```

agners (approved):
Otherwise, LGTM.

PR #4232, opened 2025-08-14
Enable publishing of dev builds to R2 without bumping version

sairon:
We currently can only use Github artifacts for on-demand builds from feature branches. However, downloading of these requires authentication and it's tricky to update a device if we need feedback from user testing. On the other hand, we never want to publish to the dev channel from anything else than from the dev branch. Restrict version bump to builds from release channels or from the dev branch only.


agners (approved):
LGTM, nice feature!

sairon:
PTAL, I adjusted how dev versions are named to avoid collisions.

agners (approved):

PR #4238, opened 2025-08-16
sysctl: Enable linear RTO for thin TCP streams

RubenKelevra:
Set `net.ipv4.tcp_thin_linear_timeouts=1` to switch retransmission timeout (RTO) backoff from exponential to linear for 'thin' TCP flows. This reduces tail latency for API-style connections that typically have very few packets in flight, improving recovery from sporadic loss without changing anything for larger TCP transfers.

Kernel definition: A flow is considered thin when `tp->packets_out < 4` and while not in the initial slow start.

See [tcp_stream_is_thin(tp) in include/net/tcp.h](https://github.com/torvalds/linux/blob/90d970cade8e67e20b09bbfdc2f0b52064322921/include/net/tcp.h#L2231).


RubenKelevra:
The first suggestion by code rabbit makes sense, I've added more details to the file's comment.

The second suggestion doesn't make any sense, because this option was removed from the Linux-Kernel 4.11 - commit [4a7f600944114](https://github.com/torvalds/linux/commit/4a7f600944114).

sairon (commented):
I don't see this change as aggressive as #4239 (as it only changes behavior for thin streams) but I have similar concerns as I raised in the other issue.

Eventually, any sysctl parameters can be changed by power users who know what they're doing via override in `/etc/sysctl.d/` (which is writable on HAOS) but to make the change global, I'd like to see some measurable impact to be convinced the RFC violation is worth it.

RubenKelevra:
@sairon while technically true, it violates the compliance with RFC 6298, the positive effect of doing something like this has been proven in real world applications quite often and is implemented as TLP with SACK-enabled streams. While both approaches cannot be 1:1 compared, the idea behind both is very similar. TLP is enabled by default in the Linux kernel.

However, TLP's limitation of requiring RACK, which is requiring TCP SACK to work, makes it unavailable for non-SACK connections. 

Non-SACK connections are extremely niche, so not much effort are put into them anymore to improve anything about them. The only use case are very small TCP stacks, like the one in embedded devices, which is why knobs like the linear timeout switch in the Linux kernel exists in the first place - there's no RFC but there's a need to improve the recovery time on latency sensitive non-SACK connections, because there's no TLP available. 

Since Home Assistant is expected to communicate with a lot of embedded devices which do not support TCP SACK (all ESPHome devices for example come to mind here) enabling this setting does make sense IMHO. Especially since enabling it is not breaking any usecase.

sairon:
Okay, that makes sense. Let's go with the change, I just remove the extra newlines in the file.

sairon (approved):

PR #4239, opened 2025-08-16
sysctl: disable TCP slow start after idle

RubenKelevra:
This knob controls whether Linux throws away its congestion window (cwnd) after a connection has been idle for at least one retransmission timeout (RTO). With a value of 0, Linux keeps the cwnd it had before the idle period and can send that amount immediately when the application resumes writing (still bounded by the receiver's advertised window and by pacing).

With slow start after idle enabled (the default), Linux allows only about 10 MSS (~14 KiB) in the first burst after idle. Even when a connection stays open to web clients, a short idle forces multiple round trips to ramp back up.

On Wi-Fi, local connections often have very low RTTs, which drives the RTO down. Between page navigations the connection is considered idle by Linux. If the next request happens during a transient latency spike on the Wi-Fi link, the sender starts with a tiny cwnd and must grow it over many RTTs, so the spike causes outsized and visible loading delays.

For devices behind typical Internet uplinks, the higher RTT makes the initial ramp-up feel even slower until the window regains size. However, here the connection does take longer to drop to idle, for Linux standards. So the connection is less likely to be considered idle between navigations.

This change does not affect flows with very small receive windows (e.g. many microcontrollers), which are limited by the peer's advertised window rather than the sender's cwnd.

Example RTOs on low jitter, low loss connections:

Defaults:
```
TCP_RTO_MIN = 200 ms
TCP_RTO_MAX = 120 s
low-jitter path so rttvar_us = 200 ms
HZ = 1000 or 250 or 100 (depending on the kernel settings)
```

*31 ms average RTT*

```
- SRTT ≈ 31 ms; RTTVAR ≈ 200 ms → Sum = 231 ms
- 'usecs_to_jiffies(231000)' = 231 jiffies (HZ 1000) -> RTO ≈ 231 ms
- If 'HZ = 250' (4 ms tick), ceil(231/4)=58 jiffies -> 232 ms RTO
- If 'HZ = 100' (10 ms tick), ceil(231/10)=23 jiffies -> 240 ms RTO
```

*178 ms average RTT*

```
- HZ=1000 (1 ms tick): 378 ms RTO
- HZ=250 (4 ms tick): ceil(378/4)=95 -> 380 ms RTO
- HZ=100 (10 ms tick): ceil(378/10)=38 -> 380 ms RTO
```

*292 ms average RTT*

```
- HZ=1000 (1 ms tick): 492 ms RTO
- HZ=250 (4 ms tick): ceil(492/4)=123 -> 492 ms RTO
- HZ=100 (10 ms tick): ceil(492/10)=50 -> 500 ms RTO
```

Any loss or jitter will increase those RTO values.

### DIFF
--- a/buildroot-external/rootfs-overlay/usr/lib/sysctl.d/20-network.conf
+++ b/buildroot-external/rootfs-overlay/usr/lib/sysctl.d/20-network.conf
@@ -5,3 +5,6 @@ net.ipv4.igmp_max_memberships = 1024
 # Increase maximum receive and send buffer size
 net.core.rmem_max = 4194304
 net.core.wmem_max = 4194304
+
+# Disable TCP slow start after idle to keep cwnd across idle
+net.ipv4.tcp_slow_start_after_idle = 0


sairon (commented):
I am sure this is something that should be set as default. I think I kind of get your motivation - considering only smart home deployments, disabling the slow start can bring some benefits. But disabling it makes the whole OS violating RFC specs, and since people often use HAOS as a generic self-hosting platform, and many of such servers are exposed to the internet, I am doubtful such change should be applied generally to all instances. In the end, there's a reason it's a knob that has different default.

RubenKelevra:
@sairon thanks for this thoughtful consideration.

I like to point out that the behavior toggled by `tcp_slow_start_after_idle` implements RFC 2861 (Congestion Window Validation). That RFC is historic and was obsoleted by RFC 7661; disabling it does not violate current TCP standards. Initial slow start and congestion control remain governed by RFC 5681 and are unaffected by this sysctl.

Apart from this, changes on the host only affect the host network namespace. Since Home Assistant core runs with `--network=host` this change needs to be done here on OS level, to be effective. However, it does not effect any Add-ons which don't run with host networking, as they run in their own network namespace.

RubenKelevra:
I retract my previous comment about TCP standard compliance.

TL;DR: Disabling slow start after idle deviates from RFC 5681's **SHOULD** restart guidance (and Linux also does not implement RFC 7661 to the teeth). So it would be more precise to call this a deviation from recommended behavior, not a hard *violation*. 



RFC 7661 does allow something very similar to the `tcp_slow_start_after_idle=0`, but the implementation in the Linux master doesn't meet the standard's requirements.

Namely RFC requires to have a 5 minute guard for this faster restart process, which is missing if we set `tcp_slow_start_after_idle=0`

The standard also requires explicit “validated vs. non-validated” state & gating, which is kinda hard to replicate in the kernel itself. However in the end effect it basically requires the sender (us in this case) to not increase the `cwnd` if we're not `cwnd`-bound. However Linux doesn't do that anyway, because that's in the spec of Cubic.

The RFC also requires that if loss is detected during NVP, the sender (us in this case) must compute a new `cwnd` from `max(pipeACK, LossFlightSize)/2`, then at the end of recovery set `cwnd = (max(pipeACK, LossFlightSize) − R)/2` and adjust `ssthresh` per RFC 5681.

This approach is slightly more conservative than standard TCP where Linux follows the active CC algorithm’s rules (e.g., Cubic reduces with β≈0.7 and proceeds per its spec).

RFC also *recommends* using pacing, which the use of `fq` as qdisc would meet here.

I agree, disabling tcp_slow_start_after_idle is a standard violation, but the differences between RFC 7661 and Linux behavior with slow start off are miniscule in practice.

I'm currently working on a fix for the non-Standard behavior of Linux for non-Sack streams and I'll put better, RFC 7661 compliant behavior for idle recovery on my todo list.

sairon:
@RubenKelevra Thank you for the elaborate reply. With all that considered, I think we could go also with this change then.

PR #4243, opened 2025-08-19
Fix scripts/enter.sh so it is usable on macOS

slonopotamus:
Also, stop requiring `sudo` for actions that do not need it

Tested by building generic_x86_64 target on a macOS machine

----

This PR possibly also allows to build HAOS on Windows via WSL2, though I haven't tested it.

## Summary by CodeRabbit

* **New Features**
  * Forwards your command-line arguments directly into the container, defaulting to an interactive bash shell when none are provided.

* **Bug Fixes**
  * Only attempts loop-device setup when supported, avoiding failures on systems without losetup.

* **Refactor**
  * Removes sudo requirements for common operations and simplifies argument handling by eliminating unused environment variables.

### DIFF
--- a/scripts/enter.sh
+++ b/scripts/enter.sh
@@ -4,28 +4,26 @@ set -e
 BUILDER_UID="$(id -u)"
 BUILDER_GID="$(id -g)"
 CACHE_DIR="${CACHE_DIR:-$HOME/hassos-cache}"
-ARGS="$*"
-COMMAND="${ARGS:-bash}"
 
 if [ "$BUILDER_UID" -eq "0" ] || [ "$BUILDER_GID" == "0" ]; then
   echo "ERROR: Please run this script as a regular (non-root) user with sudo privileges."
   exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


scripts/enter.sh
```diff
@@ -4,28 +4,26 @@ set -e
...
+docker build -t hassos:local .
 
 if [ ! -f buildroot/Makefile ]; then
   # Initialize git submodule
   git submodule update --init
 fi
 
-# Make sure loop devices are present before starting the container
-sudo losetup -f > /dev/null
+if command -v losetup && [ ! -e /dev/loop0 ]; then
```
sairon: Avoid printing path to `losetup` when it's found:
```suggestion
if command -v losetup >/dev/null && [ ! -e /dev/loop0 ]; then
```

slonopotamus:
> IIUC there's no losetup on macOS

Exactly. Even if it was, it does not make sense to call it because docker lives inside Linux VM, not on host macOS system.

PR #4248, opened 2025-08-21
Rework /usr/sbin/hassos-supervisor script

adeepn:
Rework /usr/sbin/hassos-supervisor script:
- remove hardcoded url for image
- add get image url from updater.json/internet
- add SUPERVISOR_CHANNEL defaults to stable
- move version url to variable

## Summary by CodeRabbit

- New Features
  - Channel-based Supervisor updates for controlled rollouts.
  - Automatic selection of the correct architecture image at runtime.
  - Remote version fallback when local update metadata is missing or stale.

- Refactor
  - Switched from a fixed image target to a template-driven, arch-aware image resolution.
  - Improved cleanup of old images and more robust update pull/tag and failure handling.

### DIFF
--- a/buildroot-external/rootfs-overlay/usr/sbin/hassos-supervisor
+++ b/buildroot-external/rootfs-overlay/usr/sbin/hassos-supervisor
@@ -43,7 +43,7 @@ touch ${SUPERVISOR_STARTUP_MARKER}
 # Using updater information instead of config. If the config version is
 # broken, this creates a way back (e.g., bad release).
 SUPERVISOR_VERSION=$(jq -r --arg chan "$SUPERVISOR_CHANNEL" '.supervisor // $chan' "${SUPERVISOR_DATA}/updater.json" || echo "$SUPERVISOR_CHANNEL")
-SUPERVISOR_IMAGE_TPL="$(jq -r '.images.supervisor // empty' "${SUPERVISOR_DATA}/updater.json" 2>/dev/null || true)"
+SUPERVISOR_IMAGE_TPL="$(jq -r '.image.supervisor // empty' "${SUPERVISOR_DATA}/updater.json" 2>/dev/null || true)"
 # Get version from channel in case we have no local version
 # information.
 if [ -z "${SUPERVISOR_VERSION}" ] || [ "${SUPERVISOR_VERSION}" = "$SUPERVISOR_CHANNEL" ] || [ -z "${SUPERVISOR_IMAGE_TPL}" ]; then


buildroot-external/rootfs-overlay/usr/sbin/hassos-supervisor
```diff
@@ -9,12 +9,14 @@ set -e
 . /etc/os-release
 
 # Init supervisor
+SUPERVISOR_CHANNEL="stable"
+URL_HASSIO_VERSION="https://version.home-assistant.io/${SUPERVISOR_CHANNEL}.json"
```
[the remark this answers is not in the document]
adeepn: good catch

buildroot-external/rootfs-overlay/usr/sbin/hassos-supervisor
```diff
@@ -37,29 +39,32 @@ fi
 mkdir -p "$(dirname ${SUPERVISOR_STARTUP_MARKER})"
 touch ${SUPERVISOR_STARTUP_MARKER}
 
+# Get the latest from update information
+# Using updater information instead of config. If the config version is
+# broken, this creates a way back (e.g., bad release).
+SUPERVISOR_VERSION=$(jq -r --arg chan "$SUPERVISOR_CHANNEL" '.supervisor // $chan' "${SUPERVISOR_DATA}/updater.json" || echo "$SUPERVISOR_CHANNEL")
+SUPERVISOR_IMAGE_TPL="$(jq -r '.images.supervisor // empty' "${SUPERVISOR_DATA}/updater.json" 2>/dev/null || true)"
```
sairon: The key is `image` in `updater.json`, otherwise you'll always get the default.
```suggestion
SUPERVISOR_IMAGE_TPL="$(jq -r '.image.supervisor // empty' "${SUPERVISOR_DATA}/updater.json" 2>/dev/null || true)"
```
See: https://github.com/home-assistant/supervisor/blob/2d12920b35e7314cfd1fb4f26b12d24ffe50232a/supervisor/updater.py#L320-L326

adeepn: `self._data[ATTR_IMAGE][ATTR_SUPERVISOR] = data["images"]["supervisor"]` 

and `version.home-assistant.io/stable.json` use images.supervisor too:
```
  "images": {
    "core": "ghcr.io/home-assistant/{machine}-homeassistant",
    "supervisor": "ghcr.io/home-assistant/{arch}-hassio-supervisor",
...
   }
```

Or am I missing something?

adeepn:
> Generally looks good, although the `awk` parsing is a bit hard to read. There will be minor changes in the behavior (e.g. if you somehow ended up with supervisor images from multiple repos, they will be all deleted in the removal step, and similarly only the first image in the output will be treated as `SUPERVISOR_IMAGE_ID`) but none of that shouldn't pose a problem on production HAOS instance.

Thanks for the review.  
I believe that in a running HAOS system there shouldn’t be multiple supervisor images from different repositories at the same time, so the current logic is reasonable. We could avoid using `awk`, but that would require a slightly more complex construct (though arguably a bit more readable).


> I wonder what's motivation of the change - currently there's no way to override the environment variable on HAOS, so what's the ultimate goal?

Main goal: decouple the system supervisor startup script from a hardcoded repository address, and instead retrieve that address from the versions file when needed. 

Extraction of the `SUPERVISOR_CHANNEL=stable` variable is more of a forward-looking step, in case it becomes necessary. I noticed there were changes in `buildroot-external/package/hassio/hassio.mk` regarding the supply channel — that same approach could be applied for the supervisor container startup channel as well.

Additionally, if the proposal is accepted, I’d suggest externalizing the versions file URL into a variable as well — in a way that allows it to be modified in a running system.

buildroot-external/rootfs-overlay/usr/sbin/hassos-supervisor
sairon: `ATTR_IMAGE = "image"`: https://github.com/home-assistant/supervisor/blob/b889f94ca465389b76d6a9b33213781317f72ee6/supervisor/const.py#L217

The right part of the line is getting the fetched data, `self._data` is what's then serialized to the `updater.json` by `self.save_data()`.

The `image` key is a legacy one which was present even in the version file, it was removed just recently: https://github.com/home-assistant/version/pull/432

Probably to make things simpler, no migration was done in the `updater.json` schema when the version file format changed.

sairon:
> I believe that in a running HAOS system there shouldn’t be multiple supervisor images from different repositories at the same time, so the current logic is reasonable.

Right, that's what I said, it was rather me thinking aloud :)

> Extraction of the `SUPERVISOR_CHANNEL=stable` variable is more of a forward-looking step, in case it becomes necessary.

Okay, got it. I think it doesn't do any harm and can be useful in development, so I'm not against it.

> Additionally, if the proposal is accepted, I’d suggest externalizing the versions file URL into a variable as well — in a way that allows it to be modified in a running system.

It makes sense - actually, we would like to make it easier for users and developers to run and distribute forks of HAOS for platforms which aren't officially supported, which is something that I see is your goal as well. I'll keep a note to send you a reminder and discuss how it could be achieved with you as well.

buildroot-external/rootfs-overlay/usr/sbin/hassos-supervisor
adeepn: That's why I use the `images` key. Can I mark this change as resolved?

adeepn:
> I'll keep a note to send you a reminder and discuss how it could be achieved with you as well.

Feel free contact me on discord or telegram adeepv

buildroot-external/rootfs-overlay/usr/sbin/hassos-supervisor
sairon: I don't think we're on the same page. Since `updater.json` doesn't contain `images` key (it's created by Supervisor by methods I mentioned above), `SUPERVISOR_IMAGE_TPL` will be always empty on a perfectly fine system and fetch of the version file will be always triggered. I believe it's not what you wanted. On the other hand on line 52 `.images.supervisor` is correct.

adeepn: Oh, I actually missed that point. I double-checked again and you’re right. I’ve applied the corrections.

sairon (approved):
Looks good, thank you!

PR #4256, opened 2025-08-27
Revert "Rework /usr/sbin/hassos-supervisor script (#4248)"

sairon:
This reverts commit 22fe9b19eeaef32e20c9794df80e9e0648afe676.

There are major issues when OS has no internet connectivity - in such case the script doesn't go the expected happy path after the rework and eventually removes the Docker image, essentially bricking offline installations.

Since there is no immediate benefit for HAOS and such change turns out to be high risk considering the planned release, leave it to be implemented later.

## Summary by CodeRabbit

- New Features
  - Automatically resolves and installs the correct stable Supervisor version when the local image is missing.
- Bug Fixes
  - Improves reliability of Supervisor updates by using a consistent, official image source.
  - Enhances error handling and cleanup after failed image pulls.
- Refactor
  - Simplifies image detection and version resolution logic for faster, more predictable behavior.
- Chores
  - Removes legacy channel-based version selection to streamline updates.

### DIFF
--- a/buildroot-external/rootfs-overlay/usr/sbin/hassos-supervisor
+++ b/buildroot-external/rootfs-overlay/usr/sbin/hassos-supervisor
@@ -9,14 +9,12 @@ set -e
 . /etc/os-release
 
 # Init supervisor
-: "${SUPERVISOR_CHANNEL:=stable}"  # allow env override; default to "stable"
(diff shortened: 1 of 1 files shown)


sairon:
CC @adeepn

adeepn:
I will investigate this case more closely.

Issue #4301, opened 2025-09-12
Support for Realtek RTL8188S WiFi dongle


federicogualdi:
### Describe the issue you are experiencing

I connected a USB WiFi dongle based on the Realtek RTL8188S chipset to my Home Assistant OS installation. The device is detected by the system but fails to initialize properly and is not usable for networking.
From the logs, it seems the driver for this chipset is missing or not working.

### What operating system image do you use?

generic-x86-64 (Generic UEFI capable x86-64 systems)

### What version of Home Assistant Operating System is installed?

6.12.43-haos

### Did the problem occur after upgrading the Operating System?

No

### Hardware details

USB WiFi dongle: Realtek RTL8188S WLAN Adapter
Vendor ID: 0bda
Product ID: 8171

### Steps to reproduce the issue

1. Plug the Realtek RTL8188S WiFi USB dongle into the host running Home Assistant OS.
2. Check dmesg logs.
3. The dongle is detected but fails with error -71 and is not available as a network interface.

### Anything in the Supervisor logs that might be useful for us?

```txt
(MainThread) [supervisor.hardware.monitor] Detecting remove hardware /dev/bus/usb/001/012 - None
2025-09-13 01:10:02.305 INFO (MainThread) [supervisor.hardware.monitor] Detecting add hardware /dev/bus/usb/001/015 - None
```

### Anything in the Host logs that might be useful for us?

```txt
Relevant dmesg output:
usb 1-7: new high-speed USB device number 12 using xhci_hcd
usb 1-7: New USB device found, idVendor=0bda, idProduct=8171, bcdDevice= 2.00
usb 1-7: New USB device strings: Mfr=1, Product=2, SerialNumber=3
usb 1-7: Product: RTL8188S WLAN Adapter
usb 1-7: Manufacturer: Realtek
usb 1-7: device not accepting address 11, error -71
```

## System Information

version | core-2025.9.1
-- | --
installation_type | Home Assistant OS
dev | false
hassio | true
docker | true
container_arch | amd64
user | root
virtualenv | false
python_version | 3.13.7
os_name | Linux
os_version | 6.12.43-haos
arch | x86_64
timezone | Europe/Rome
config_dir | /config

<details><summary>Home Assistant Cloud</summary>

logged_in | false
-- | --
can_reach_cert_server | ok
can_reach_cloud_auth | ok
can_reach_cloud | ok

</details>

<details><summary>Home Assistant Supervisor</summary>

[...]

PR #4308, opened 2025-09-15
enable Realtek RTL8188SU (r8712u) USB WiFi driver

federicogualdi:
## Describe the change
This enables the `CONFIG_R8712U` kernel option as a module to support Realtek RTL8188SU/RTL8188S based USB WiFi adapters (driver: r8712u).

Affected devices (example):
- Realtek RTL8188S WLAN Adapter (VID:0bda, PID:8171)

With this change, these adapters are properly initialized and can be
used as network interfaces in Home Assistant OS.

## Why is this needed?
Currently, RTL8188S-based USB WiFi dongles are detected by the system but fail to initialize because the kernel driver is not enabled.  
Users see `device not accepting address 11, error -71` errors in dmesg and the adapter is unusable.  
Enabling `CONFIG_R8712U` resolves this issue.

## Additional information
- Tested on: Home Assistant OS 6.12.43 (generic-x86-64)
- Hardware: Realtek RTL8188S WLAN Adapter (0bda:8171)

## Related issue
Fixes: #4301

## Checklist
- [x] Change made against `dev` branch
- [x] Builds and boots successfully
- [x] Hardware verified with RTL8188S dongle

## Summary by CodeRabbit

* **New Features**
  * Enabled driver support for additional Realtek USB Wi‑Fi adapters, expanding wireless hardware compatibility.
  * Provided the driver as a loadable module, allowing on-demand usage without impacting systems that don’t require it.


sairon (commented):
Thank you for the effort you put into this issue, however, I'm afraid my response will disappoint you.

The driver you want to enable [was removed in Linux 6.13](https://github.com/torvalds/linux/commit/41e883c137ebe6eec042658ef750cbb0529f6ca8), and although nothing technically prevents us from enabling it in HAOS - which runs 6.12 now, it would be a short-sighted move, as we are likely to move onto a newer LTS kernel in the coming months. Moreover, the RTL8188SU is something rather vintage and outdated, given what people expect from WiFi nowadays. And I think that "stability" can't be used as an argument either, as we're talking about mostly unmaintained driver that stayed in the staging phase for 10 years.

Taking these points into account, I don't think it's a good idea to give false hopes about this hardware. It seems there's not even any well-maintained out-of-tree driver which could be used as a replacement, so these cards are virtually incompatible with modern Linux anymore. Sorry :/

federicogualdi:
Thanks for the detailed explanation. 
I understand the reasoning behind this decision, especially considering the upcoming kernel updates and the maintenance state of the driver.

Appreciate you taking the time to clarify!

PR #4335, opened 2025-10-03
rpi5 board with GPU support

pepijndevos:
Hi it's me again with a simplified and board-scoped version of #4159 that is hopefully more reasonable to upstream.

I've split this PR into two commits: one straight up copy of the rpi5-64 board, and a separate commit to add all the GPU changes. This should make it easier to see what was changed and rebase the changes in the future.

The breakthrough that I hope makes this a reasonable PR now is this new patch that is only a couple of lines: https://github.com/geerlingguy/raspberry-pi-pcie-devices/discussions/756 It even opens the door to support other GPU vendors in the future.

So essentially what this is is a copy of rpi5 with
- amdgpu firmware
- amdgpu kernel drivers
- a small drm patch to fix said kernel drivers on the raspberry pi
- cmdline parameters that fix amdgpu crashes I've observed

There is one other change that is specific to our Sentinel Core board: selecting the external wifi antenna. I'm open to suggestions to make this as useful as possible for anyone experimenting with GPUs, weather they use our board or a concoction of hats and eGPU cases.

The kernel patch applies cleanly on a wide range of kernels from the current 6.12 to the latest 6.17, and the rpi5-64 directory was last changed 8 months ago, so it seems like the maintenance burden of this PR would be extremely low. Getting this merged would relieve us of a great maintenance burden and make it a lot easier for our users to update their OS.

I'm not an expert on buildroot, so feedback welcome.

## Summary by CodeRabbit

* New Features
  * Added a Raspberry Pi 5 (64-bit, GPU-focused) target with AMD GPU support.
  * Introduced A/B boot slot management via RAUC tryboot (get/set primary, state, commit).
  * Enabled boot-time options: 64-bit mode, performance boost, VC4 DRM, auto camera/display detection, and zram.
  * Automated image preparation: kernel compression, overlays handling, and boot cmdline/config provisioning.
* Chores
  * Added board metadata and a comprehensive defconfig with required firmware, packages, and system services.
  * Integrated image conversion into the build workflow.


pepijndevos:
Without the antenna change it's really just a rpi5, so that should meet your board requirements.

But if neither adding the firmware to the rpi5 board nor a dedicated GPU variant are acceptable I look forward to the fork plans.

Fwiw raspberry pi seems interested in enabling these changes upstream but they would ship the firmware as a seperate package:

https://github.com/geerlingguy/raspberry-pi-pcie-devices/issues/764#issuecomment-3380946750

Sadly, haos users don't have any way to install os packages so either you have to burden all users with gpu firmware (like the x86 build) or provide a dedicated GPU image, official or forked.

PR #4346, opened 2025-10-09
Enable powersave CPU frequency governor

agners:
The CPUfreq governor "powersave" sets the CPU statically to the lowest frequency within the borders of scaling_min_freq and scaling_max_freq. This can be useful if a particular power budget should not ever be crossed. Can be set using `cpufreq.default_governor=powersave`. Note that this obviously affects performance.

## Summary by CodeRabbit

* **New Features**
  * Enabled a CPU "powersave" frequency governor on x86-64 builds so the system can reduce CPU frequency under light load.
  * Helps lower power consumption and heat (beneficial for laptops) while preserving existing default behavior; users can opt into the power-saving mode.


buildroot-external/board/pc/generic-x86-64/kernel.config
```diff
@@ -7,6 +7,7 @@ CONFIG_CPU_IDLE=y
 CONFIG_CPU_IDLE_GOV_MENU=y
 CONFIG_INTEL_IDLE=y
 
+CONFIG_CPU_FREQ_GOV_POWERSAVE=y
 # CONFIG_CPU_FREQ_DEFAULT_GOV_ONDEMAND is not set
 # CONFIG_CPU_FREQ_DEFAULT_GOV_USERSPACE is not set
 CONFIG_CPU_FREQ_DEFAULT_GOV_SCHEDUTIL=y
```
sairon: Please reorder to follow Kconfig order:
```suggestion
# CONFIG_CPU_FREQ_DEFAULT_GOV_ONDEMAND is not set
# CONFIG_CPU_FREQ_DEFAULT_GOV_USERSPACE is not set
CONFIG_CPU_FREQ_DEFAULT_GOV_SCHEDUTIL=y
CONFIG_CPU_FREQ_GOV_POWERSAVE=y
```

PR #4348, opened 2025-10-09
add infrastructure for x86 CPU microcode updates

jens-maus:
This PR adds the infrastructure to enable early application of CPU microcodes for the generic-x86_64 platform once the `BR2_PACKAGE_INTEL_MICROCODE` or `BR2_PACKAGE_LINUX_FIRMWARE_AMD_UCODE` options are enabled. Once enabled a kconfig fragment will be generated which will then dynamically add `CONFIG_EXTRA_FIRMWARE` option to the kernel config for embedding all microcode files accordingly. This should help to get CPU microcode fixes in place once HAos is directly installed on a x86-64 system using the generic-x86_64 platform images.

## Summary by CodeRabbit

- New Features
  - Early CPU microcode updates for Intel and AMD are now bundled into the kernel, improving boot-time reliability and security on more x86_64 hardware.
  - Broader firmware coverage enabled for modern CPUs.
  - Reduced root filesystem size by excluding duplicate microcode files from the image.
  - Transparent to users: no configuration changes required; systems automatically benefit when supported CPUs are detected.

### DIFF
--- a/buildroot-external/scripts/rootfs-layer.sh
+++ b/buildroot-external/scripts/rootfs-layer.sh
@@ -40,7 +40,7 @@ function fix_rootfs() {
         find "${TARGET_DIR}/share/locale" -type d -empty -delete
     fi
 
-    # Remove unnecessary microcodes in /lib/firmware
+    # Remove cpu microcodes from /lib/firmware because they are embedded in the kernel
     rm -rf "${TARGET_DIR}/lib/firmware/intel-ucode"
     rm -rf "${TARGET_DIR}/lib/firmware/amd-ucode"
 }


buildroot-external/external.mk
```diff
@@ -10,3 +10,26 @@ linux-check-dotconfig: linux-check-configuration-done
...
+		( cd $(BINARIES_DIR) 2>/dev/null && \
+		  find intel-ucode amd-ucode -type f \! -name "*README*" \! -name "*.asc" -printf '%p ' 2>/dev/null || true ); \
+		echo '"'; \
+	} > $(UCODE_FRAG)
+endef
+
+LINUX_PRE_PATCH_HOOKS += GEN_UCODE_FRAGMENT
+LINUX_KCONFIG_FRAGMENT_FILES += $(UCODE_FRAG)
+
+endif
```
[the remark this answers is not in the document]
jens-maus: @coderabbitai Adding these LINUX_DEPENDENCIES guards is not required because the kernel build linux.mk file of buildroot does already ensure proper dependency handling (see https://github.com/home-assistant/buildroot/blob/7ce414454b4ddc8cc8466f657d6044fdde58af37/linux/linux.mk#L83-L87)

jens-maus:
@sairon Any news of and when this PR might be integrated/merged? Would be great to get it potentially merged into 16.3!

sairon:
@jens-maus I intentionally postponed it after the 16.3 beta cut to stage it for the 17.0 release. I will have a closer look later this week.

jens-maus:
Thanks! Note, that I have this already integrated in two of my Buildroot projects and it works flawlessly so far! 😜

buildroot-external/configs/generic_x86_64_defconfig
```diff
@@ -102,6 +102,8 @@ BR2_PACKAGE_LINUX_FIRMWARE_BNX2X=y
 BR2_PACKAGE_LINUX_FIRMWARE_RTL_815X=y
 BR2_PACKAGE_LINUX_FIRMWARE_RTL_8169=y
 BR2_PACKAGE_LINUX_FIRMWARE_USB_SERIAL_TI=y
+BR2_PACKAGE_LINUX_FIRMWARE_AMD_UCODE=y
```
sairon, line 105: This config option doesn't exist yet in 2025.02.x.

buildroot-external/scripts/rootfs-layer.sh
```diff
@@ -39,6 +39,10 @@ function fix_rootfs() {
         find "${TARGET_DIR}/share/locale" -name "grub.mo" -delete
         find "${TARGET_DIR}/share/locale" -type d -empty -delete
     fi
+
+    # Remove unnecessary microcodes in /lib/firmware
```
sairon: Worth adding an explainer why they are unnecessary.

buildroot-external/external.mk
```diff
@@ -10,3 +10,26 @@ linux-check-dotconfig: linux-check-configuration-done
 		--src-kconfig $(LINUX_SRCDIR)Kconfig \
 		--actual-config $(LINUX_SRCDIR).config \
 		$(shell echo $(BR2_LINUX_KERNEL_CUSTOM_CONFIG_FILE) $(BR2_LINUX_KERNEL_CONFIG_FRAGMENT_FILES))
+
+# if cpu microcode is required we embed it into the kernel build
+ifeq ($(or $(BR2_PACKAGE_INTEL_MICROCODE),$(BR2_PACKAGE_LINUX_FIRMWARE_AMD_UCODE)),y)
+
```
sairon, line 16: Use the idiomatic pattern instead and remove the newline:
```suggestion
ifneq ($(BR2_PACKAGE_INTEL_MICROCODE)$(BR2_PACKAGE_LINUX_FIRMWARE_AMD_UCODE),)

```

buildroot-external/external.mk
```diff
@@ -10,3 +10,26 @@ linux-check-dotconfig: linux-check-configuration-done
...
+		printf 'CONFIG_EXTRA_FIRMWARE="'; \
+		( cd $(BINARIES_DIR) 2>/dev/null && \
+		  find intel-ucode amd-ucode -type f \! -name "*README*" \! -name "*.asc" -printf '%p ' 2>/dev/null || true ); \
+		echo '"'; \
+	} > $(UCODE_FRAG)
+endef
+
+LINUX_PRE_PATCH_HOOKS += GEN_UCODE_FRAGMENT
+LINUX_KCONFIG_FRAGMENT_FILES += $(UCODE_FRAG)
+
```
sairon: Remove the newline here too:

```suggestion
LINUX_KCONFIG_FRAGMENT_FILES += $(UCODE_FRAG)
```

jens-maus:
Thanks @sairon for your review. When I was first working on this PR I also thought if there might be a better idea than adding these things to `buildroot-external/external.mk`, however, that was currently the only possible and verbose way of adding the necessary modifications for dynamically generating such a kconfig fragement file on the fly before the linux kernel build which then uses it via `CONFIG_EXTRA_FIRMWARE`. But if you have another idea on how to dynamically inject the list of available microcodes located in `BINARIES_DIR`, feel free to put up a suggestion and I will try to work it out.

Regarding the missing `BR2_PACKAGE_LINUX_FIRMWARE_AMD_UCODE` in buildroot 2025.02.x, I indeed missed that because all of my other projects are already at 2025.08.x, of course ;-) But I can of course also remove it from the config file until you switch HAos to buildroot 2025.08.x. However, I would suggest to keep `BR2_PACKAGE_LINUX_FIRMWARE_AMD_UCODE` in the `ifeq()` in `external.mk` to have it ready for buildroot 2025.08.x and later.

sairon:
Sorry that this took so long, but unfortunately, we'll have to put this idea to the ice box. The issue is that the ucode binaries currently (even without the AMD ones) bloat the kernel way over the partition size:
```
ERROR: squashfs(kernel.img): generated image /build/output_generic_x86_64/images/kernel.img is larger than given image size (31580160 v 25165824)
```
We won't be able to do that without re-partitioning the disk, which is not an easy thing to do.

FWIW, I tried to implement it as a virtual package, so essentially the stuff needed to be done would be done after enabling `BR2_PACKAGE_EMBEDDED_X86_CPU_MICROCODE`, yet once I tried to build it, I ran into this issue. Then I noticed that this flaw affects your code as well...

sairon:
Closing as not currently feasible due to the above limitations.

PR #4374, opened 2025-10-30
Remove support of armv7 targets

### LINKED ISSUE #4372: Remove build of boards with deprecated architectures
Starting with HA Core 2025.12.0 (including dev), there won’t be no build for 32bit archs anymore. OS should logically not build new development builds using the old images anymore. The removal was announced in https://www.home-assistant.io/blog/2025/05/22/deprecating-core-and-supervised-installation-methods-and-32-bit-systems/

sairon:
Following the deprecation [announced in May 2025](https://www.home-assistant.io/blog/2025/05/22/deprecating-core-and-supervised-installation-methods-and-32-bit-systems/), HA Core 2025.12.x won't support 32-bit architectures anymore. This affects the following boards:

* ASUS Tinker
* ODROID XU4
* Raspberry Pi 2
* Raspberry Pi 3 (32-bit only, rpi3-64 is still available)
* Raspberry Pi 4 (32-bit only, rpi4-64 is still available)

HAOS 17.0 will be released after the December release of Core, so it's not desirable to include outdated components that are no longer supported in the OS builds.

Closes #4372 

## Summary by CodeRabbit

* **Documentation**
  * Updated supported hardware list (removed several board entries)

* **Chores**
  * Removed support and board-specific build/boot configurations for Raspberry Pi (multiple models), ASUS Tinker Board, and ODROID‑XU4
  * Cleaned up associated device-tree patches, boot scripts, kernel presets, and package selections

### DIFF
--- a/buildroot-external/board/raspberrypi/kernel-armv7.config
+++ b/buildroot-external/board/raspberrypi/kernel-armv7.config
@@ -1,4 +1,0 @@
-# Config options specific only to 32bit (armv7) builds
-
-CONFIG_KERNEL_LZ4=y
-# CONFIG_KERNEL_GZIP is not set


jens-maus:
> * Raspberry Pi 3 (32-bit only, rpi3-64 is still available)
> * Raspberry Pi 4 (32-bit only, rpi4-64 is still available)

What about renaming the `rpi3-64` and `rpi4-64` to simply `rpi3` and `rpi4` after the 32bit deprecation?  And probably also the `rpi5-64` for which anyway there is and was no 32bit at all. I think it would make sense to get rid of the `-64` addition, or is there something that would hinder such a rename?

agners:
How will this affect version json updates? Ideally we should stop updating the version number for the ones no longer supported here, but from what I remember we simply update all boards always no? :thinking:

sairon:
@jens-maus 

> What about renaming the `rpi3-64` and `rpi4-64` to simply `rpi3` and `rpi4` after the 32bit deprecation? And probably also the `rpi5-64` for which anyway there is and was no 32bit at all. I think it would make sense to get rid of the `-64` addition, or is there something that would hinder such a rename?

Discussed it shortly with @agners and we kind of agreed that this would cause more trouble to make it worth. We will need to handle these changes both in the RAUC hooks and on Supervisor side, to make sure that the old `-64` boards migrate correctly to the unsuffixed ones and that the updater knows about this change too. Similarly, the change will also prevent from downgrades (which we don't have a reason to block) as there's currently no migration path that allows for installing e.g. `rpi3-64` compatible bundle on `rpi3` system.

---

@agners 

> How will this affect version json updates? Ideally we should stop updating the version number for the ones no longer supported here, but from what I remember we simply update all boards always no? 🤔

Supervisor [handles missing board gracefully](https://github.com/home-assistant/supervisor/blob/1448a33dbfd0b6c071fa91034409fa7b9d7c0f7e/supervisor/updater.py#L311), so all that should be needed is to remove those entries, ideally before we run the build (i.e. after merging this PR in `dev.json` and then before the releases of beta and stable in other JSONs).

PR #4391, opened 2025-11-10
Enable iptables comments for all targets

monosoul:
Atm some targets don't have comments support enabled for `iptables`. There's no reason to limit that to just `arm64-rockhip` target, so this change addresses that.

## Summary by CodeRabbit

* **Chores**
  * Adjusted kernel configuration settings to optimize network filtering support across different platform builds.


monosoul:
Tbh, I think it prolly makes sense to enable other iptables-related modules for non `arm64-rockhip` targets too. I'm up for updating the PR in case the maintainers wouldn't mind that.

sairon:
Thank you! Sure there is no reason to enable it only for some targets.

> Tbh, I think it prolly makes sense to enable other iptables-related modules for non `arm64-rockhip` targets too. I'm up for updating the PR in case the maintainers wouldn't mind that.

I'd rather say the Rockchip fragment deserves a cleanup and enable modules on a justified basis.

PR #4394, opened 2025-11-11
Clean up hassio build scripts

sairon:
Extract some of the parts of the "image import" to the script creating the data partition to separate concerns. The Docker data directory is now passed as a daemon option, instead of only mounting the data partition's folder to the default directory, to be closer to the deployment setup. Also trap the exit and error signals to remove the build container and unmount the data partition, as failed or cancelled build have been leaking the containers/mounts when building interactively (attached to the build container shell).

## Summary by CodeRabbit

* **Chores**
  * Data partition handling improved: supervisor data and updater channel now persist in a dedicated data location across runs.
  * Container init streamlined: Docker-in-Docker uses the persistent data location as its data root, with enhanced cleanup on error/exit for more reliable start/stop.
  * AppArmor and init adjustments: AppArmor assets are placed in persistent data (where used), and the import flow no longer accepts a channel argument.

### DIFF
--- a/buildroot-external/package/hassio/create-data-partition.sh
+++ b/buildroot-external/package/hassio/create-data-partition.sh
@@ -29,7 +29,7 @@ container=$(docker run --privileged -e DOCKER_TLS_CERTDIR="" \
     -v "${build_dir}":/build \
     -d "docker:${docker_version}-dind" --feature containerd-snapshotter --data-root /mnt/data/docker)
 
-docker exec "${container}" sh /build/dind-import-containers.sh "${channel}"
+docker exec "${container}" sh /build/dind-import-containers.sh
 
 sudo bash -ex <<EOF
 # Indicator for docker-prepare.service to use the containerd snapshotter
@@ -37,7 +37,7 @@ touch "${data_dir}/.docker-use-containerd-snapshotter"
 
 # Setup AppArmor
 mkdir -p "${data_dir}/supervisor/apparmor"
-wget -O "${data_dir}/supervisor/apparmor/hassio-supervisor" "${APPARMOR_URL}"
+curl -fsL -o "${data_dir}/supervisor/apparmor/hassio-supervisor" "${APPARMOR_URL}"
 
 # Persist build-time updater channel
 jq -n --arg channel "${channel}" '{"channel": \$channel}' > "${data_dir}/supervisor/updater.json"


buildroot-external/package/hassio/create-data-partition.sh
```diff
@@ -7,30 +7,39 @@ channel=$3
...
-	-v "${build_dir}/data/docker/":/var/lib/docker \
-	-v "${build_dir}":/build \
-	-d "docker:${docker_version}-dind" --feature containerd-snapshotter)
+    -v "${data_dir}":/mnt/data \
+    -v "${build_dir}":/build \
+    -d "docker:${docker_version}-dind" --feature containerd-snapshotter --data-root /mnt/data/docker)
 
-docker exec "${container}" sh /build/dind-import-containers.sh "${channel}"
+# shellcheck disable=SC2064
+trap "docker rm -f ${container}; sudo umount '${data_dir}'" ERR EXIT
```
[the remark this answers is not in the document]
sairon: Good point, except that this wouldn't work in case the container starts, as the `$container` variable will be expanded before the startup and will stay empty. The `|| true` shouldn't be needed either because it seems that `docker -f` returns zero exit code in simple error cases (like when the container doesn't exist anymore).

buildroot-external/package/hassio/create-data-partition.sh
```diff
@@ -7,30 +7,38 @@ channel=$3
...
-docker stop "${container}"
-
+sudo bash -ex <<EOF
 # Indicator for docker-prepare.service to use the containerd snapshotter
-sudo touch "${build_dir}/data/.docker-use-containerd-snapshotter"
+touch "${data_dir}/.docker-use-containerd-snapshotter"
+
+# Setup AppArmor
+mkdir -p "${data_dir}/supervisor/apparmor"
+wget -O "${data_dir}/supervisor/apparmor/hassio-supervisor" "${APPARMOR_URL}"
```
agners: It seems that is pretty much the only place where we use `wget` still, most other places we use `curl`. Maybe we can use `curl` here too? 🤔

buildroot-external/package/hassio/create-data-partition.sh
```diff
@@ -7,30 +7,38 @@ channel=$3
...
 container=$(docker run --privileged -e DOCKER_TLS_CERTDIR="" \
-	-v "${build_dir}/data/":/data \
-	-v "${build_dir}/data/docker/":/var/lib/docker \
-	-v "${build_dir}":/build \
-	-d "docker:${docker_version}-dind" --feature containerd-snapshotter)
+    -v "${data_dir}":/mnt/data \
+    -v "${build_dir}":/build \
+    -d "docker:${docker_version}-dind" --feature containerd-snapshotter --data-root /mnt/data/docker)
 
 docker exec "${container}" sh /build/dind-import-containers.sh "${channel}"
```
agners: Seems with that we don't need to pass the channel anymore here.

sairon: Good catch, forgot about this when I removed handling of that argument.

buildroot-external/package/hassio/create-data-partition.sh
sairon: Sure, why not.

buildroot-external/package/hassio/create-data-partition.sh
```diff
@@ -7,30 +7,38 @@ channel=$3
...
-	-v "${build_dir}/data/docker/":/var/lib/docker \
-	-v "${build_dir}":/build \
-	-d "docker:${docker_version}-dind" --feature containerd-snapshotter)
+    -v "${data_dir}":/mnt/data \
+    -v "${build_dir}":/build \
+    -d "docker:${docker_version}-dind" --feature containerd-snapshotter --data-root /mnt/data/docker)
 
-docker exec "${container}" sh /build/dind-import-containers.sh "${channel}"
-
-docker stop "${container}"
```
agners, line 30: Is this not required anymore?

sairon: Should not be, it's stopped through the `trap` which does `docker rm -f`.

agners: Ah ok 👍

PR #4427, opened 2025-12-03
Updating hailo version to 4.23.0

BarryOCathain:


buildroot-external/package/hailo8-firmware/hailo8-firmware.hash
```diff
@@ -1 +1 @@
-sha256  2a5c94591d9e70d884242e64bf2388b0d2d46b816a335b4c00c3f81a07832635  hailo8_fw.4.21.0.bin
+sha256  1ba9528972091ec17bebc0dc7ea2e6f4449efe70664890f6387ccbc7b60626ee  hailo8_fw.4.21.0.bin
```
sairon, line 1: FWIW the version here is wrong:
```suggestion
sha256  1ba9528972091ec17bebc0dc7ea2e6f4449efe70664890f6387ccbc7b60626ee  hailo8_fw.4.23.0.bin
```

sairon (changes requested):
Thank you for the PR. However, currently I'll politely refuse it. The thing is that this change will break Frigate compatibility until a new Frigate release is published with exactly the same version hardcoded. This is a stupid limitation of the Hailo userspace runtime which doesn't have any forward or backward compatibility with the kernel driver/firmware. When checking the (also stupidly auth-walled) changelog of the Hailo driver, I don't see any changes that would justify this hassle.

BarryOCathain:
Hey @sairon thanks for this and apologies for wasting your time. I didn't realise that the PR would be reviewed at this point. I'm trying to work on this issue https://github.com/orgs/home-assistant/discussions/660, and basically it will require the hailo multi-process service running on the OS, so that multiple containers / addons will be able to access the hardware at the same time.

PR #4470, opened 2026-01-02
Enable modem manager and kernel modules

herrernst:


herrernst:
sorry, that was too fast/early

PR #4471, opened 2026-01-04
Add eq3_char_loop patch for termios2/glibc 2.42+ compatibility

jens-maus:
This PR updates the eq3_char_loop package to contain the latest eq3_char_loop v1.3 sources with termios2 compatibility which is required for glibc 2.42+ environments (latest OpenCCU) so that the HMIPServer within OpenCCU is able to startup correctly.

This refs https://github.com/OpenCCU/OpenCCU/pull/3441

## Summary by CodeRabbit

* **New Features**
  * Added support for extended termios2 serial communication ioctls.

* **Bug Fixes**
  * Fixed driver compatibility with Linux 6.4 and later kernel versions.

* **Chores**
  * Updated kernel module build configuration; bumped driver version to 1.3.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #4504, opened 2026-01-26
Add generic_raw_uart patch for termios2/glibc 2.42+ compatibility

jens-maus:
This PR adds termios2 compatibility to the ioctl interface of `generic_raw_uart` making it compatible to glibc 2.42+ environments which prefer termios2 rather than the old/obsolete termios ioctls. This is a similar change to what was introduced to `eq3_char_loop` in #4471. However, this change in `generic_raw_uart` is required so that `/dev/raw-uart` will be termios2 compatible for applications e.g. within the OpenCCU Add-on directly opening it in case a non-dualcopro RF module like `HmIP-RFUSB-TK` is used. (This refs https://github.com/OpenCCU/OpenCCU/issues/3498)

## Summary by CodeRabbit

* **New Features**
  * Added termios2 terminal I/O support for improved compatibility with modern serial applications.

* **Improvements**
  * Enhanced serial port operation handling and command support.
  * Version updated to 1.32.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/buildroot-external/package/generic_raw_uart/0002-generic_raw_uart-termios2.patch
+++ b/buildroot-external/package/generic_raw_uart/0002-generic_raw_uart-termios2.patch
@@ -1,0 +1,188 @@
+kernel/generic_raw_uart.c: added termios2/glibc 2.42+ compatibility
+
+Upstream: Not applicable
+
+Signed-off-by: Jens Maus <mail@jens-maus.de>
+
+diff --git a/kernel/generic_raw_uart.c b/kernel/generic_raw_uart.c
+index 0c3e3de..fac1ff6 100644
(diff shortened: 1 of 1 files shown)


jens-maus:
@sairon Can you please rather quickly review and integrate this PR because upon release of a new OpenCCU HA-Addon version two days ago quite some users are affected by this. In addition, would it be possible to speed up the HomeAssistantOS 17.1 release somewhat integrating this change?

jens-maus:
@sairon Sorry to bother you once more: when is 17.1 expected to hit the road because I need to release a hotfix of my OpenCCU addon which unfortunately requires this change here?

Issue #3832, opened 2025-01-27
wifi.powersave option reverts to default after host reboot


vturekhanov:
### Describe the issue you are experiencing

Exploring issue #3826 I tried to change powersave mode for my wireless adapter using nmcli tool and direct edit connection file using vi (/etc/NetworkManager/system-connections/Supervisor\ wlp1s0.nmconnection). Setting powersave=2 was successful in both cases but after host reboot I see powersave option reverts to its default value: powersave=1. So, I can't change powersave option.

### What operating system image do you use?

generic-x86-64 (Generic UEFI capable x86-64 systems)

### What version of Home Assistant Operating System is installed?

14.1

### Did the problem occur after upgrading the Operating System?

No

### Hardware details

G3 Mini PC Alder Lake N100 Intel 12th DDR4 8GB RAM 256GB ROM WiFi BT
https://aliexpress.ru/item/1005006340106437.html

### Steps to reproduce the issue

1. Set powersave option to 2 in the wifi section of network connection in the connection configuration file (/etc/NetworkManager/system-connections/Supervisor\ wlp1s0.nmconnection in my case).
2. Reboot the host.
3. View connection configuration file.

### Anything in the Supervisor logs that might be useful for us?

```txt
n/a
```

### Anything in the Host logs that might be useful for us?

```txt
n/a
```

## System Information

version | core-2025.1.4
-- | --
installation_type | Home Assistant OS
dev | false
hassio | true
docker | true
user | root
virtualenv | false
python_version | 3.13.1
os_name | Linux
os_version | 6.6.66-haos
arch | x86_64
timezone | Asia/Almaty
config_dir | /config

<details><summary>Home Assistant Cloud</summary>

logged_in | false
-- | --
can_reach_cert_server | ok
can_reach_cloud_auth | ok
can_reach_cloud | ok

</details>

<details><summary>Home Assistant Supervisor</summary>

host_os | Home Assistant OS 14.1
-- | --
update_channel | stable
supervisor_version | supervisor-2024.12.3
agent_version | 1.6.0
docker_version | 27.2.0
disk_total | 234.0 GB
disk_used | 5.9 GB
healthy | true
supported | true
host_connectivity | true
supervisor_connectivity | true
ntp_synchronized | true
virtualization | 
board | generic-x86-64
supervisor_api | ok
version_api | ok
installed_addons | Matter Server (7.0.0)

</details>

<details><summary>Dashboards</summary>

dashboards | 2
-- | --
[...]

PR #4529, opened 2026-02-12
Disable Wi-Fi powersave by default for all connections

sairon:
Set wifi.powersave to 2 (disabled) in NetworkManager settings by default for all connections. Since HAOS is generally used on servers, powersaving doesn't bring any obvious benefit and is often cause of problems and higher network latency. If needed, nmcli can be used to override the new default.

Refs #3832

## Summary by CodeRabbit

* **Chores**
  * Enabled Wi-Fi power saving in network configuration to optimize battery consumption and performance.


agners (approved):
LGTM, thanks!

PR #4548, opened 2026-02-18
Add Full Microsoft Surface (all devices) Support for HAOS Kiosk

keviinh:
This pull request introduces complete Microsoft Surface device compatibility for Home Assistant OS (HAOS), specifically targeting reliable usage in ( HAOS Kiosk @puterboy  ) deployments on Surface tablets, laptops, and PCs.

The changes are based on a full custom HAOS build process including:
- Surface kernel configuration
- Surface firmware inclusion
- Kernel patch integration
- Deterministic boot‑time input availability
- Update‑safe buildroot customization workflow
- These improvements ensure Surface hardware works out‑of‑the‑box at boot, which is required for kiosk‑style installations where no external keyboard or SSH access is available.

Background & Motivation
Microsoft Surface devices are increasingly used as:

Wall dashboards
- Touch control panels
- Always‑on HAOS kiosk systems
- However, current HAOS kernel configuration may boot without functional touch, keyboard, or hardware buttons because several critical Surface drivers are compiled as modules (=m).

HAOS uses a minimal immutable root filesystem and does not guarantee early module loading, which causes unreliable input during boot.

This PR resolves that by:

Building essential Surface drivers directly into the kernel (=y)
Including required firmware
Providing a reproducible buildroot integration path
Scope of Changes

Kernel Configuration Adjustments
Built‑in drivers (changed from =m → =y)
These are required for boot‑critical input and device control:

CONFIG_SURFACE_BUTTON
CONFIG_SURFACE_HOTPLUG
CONFIG_SURFACE_TYPECOVER
CONFIG_HID_MULTITOUCH
CONFIG_I2C_HID
CONFIG_I2C_HID_ACPI

Result:
- Working touchscreen at boot
- Functional Type Cover keyboard/trackpad
- Operational power and volume buttons
- Deterministic kiosk usability
- Drivers intentionally kept as modules
Non‑essential or experimental hardware paths:
CONFIG_INTEL_IPTS
CONFIG_INTEL_ITHC
CONFIG_VIDEO_IPU3_CIO2
CONFIG_VIDEO_IPU3_IMGU

Reasons:
- Not required for kiosk/dashboard usage
- Reduces kernel size
- Avoids unnecessary boot dependencies
- Surface Kernel Integration via Buildroot
- Custom kernel config reference

Added to:

generic_x86_64_defconfig
BR2_LINUX_KERNEL_CUSTOM_CONFIG_FILE = $(BR2_EXTERNAL_HASSOS_PATH)/kernel/v6.12.y/linux-surface.config

This ensures the Surface configuration is:

Reproducible
Version‑controlled
Automatically applied during HAOS builds
Surface Firmware Packaging
Introduced buildroot package:
linux-surface-firmware.mk

Responsibilities:

Fetch Surface firmware
Install into /lib/firmware
Guarantee availability during boot
This is required for:

Surface aggregator functionality
Input subsystem stability
Kernel Patch Support
Surface‑specific kernel patches are placed in:
buildroot-external/patches/linux

```
PATCHDIR=~/haos/linux-surface/patches/6.12 # adjust if needed
i=100
for p in "$PATCHDIR"/*.patch; do
printf -v n "%04d" "$i"
cp "$p" "buildroot-external/patches/linux/${n}-linux-surface-$(basename "$p")"
i=$((i+1))
done
```

This aligns HAOS with:

linux‑surface upstream work
Real‑world Surface hardware requirements
Secure Boot / Image Signing Compatibility
Custom SSL key generation is included in the workflow to maintain:

HAOS image signing
Compatibility with secure boot expectations
Functional Impact
What works after this PR

Touchscreen during early boot

Type Cover keyboard and trackpad

Hardware buttons (power/volume)

Stable kiosk interaction without external peripherals

Devices affected

Surface tablets
Surface laptops
Surface PCs used as HA dashboards
Systems NOT affected

Non‑Surface x86 systems
Existing HAOS installations
Kernel size impact is minimal
Testing Performed

Validated on real Surface hardware with custom‑built HAOS image:

Successful boot
Working touch input
Working Type Cover
Functional buttons
Normal HAOS runtime behavior
Also verified:
Rebuild after HAOS version update
Persistence of buildroot customizations
Upstream Alignment
This PR does not introduce new drivers.

Instead, it:

Uses existing Surface kernel support
Aligns HAOS configuration with linux‑surface ecosystem
Ensures compatibility with HAOS immutable architecture
Patch Example
-CONFIG_SURFACE_BUTTON=m
-CONFIG_SURFACE_HOTPLUG=m
-CONFIG_SURFACE_TYPECOVER=m
-CONFIG_HID_MULTITOUCH=m
-CONFIG_I2C_HID=m
-CONFIG_I2C_HID_ACPI=m
+CONFIG_SURFACE_BUTTON=y
+CONFIG_SURFACE_HOTPLUG=y
+CONFIG_SURFACE_TYPECOVER=y
+CONFIG_HID_MULTITOUCH=y
+CONFIG_I2C_HID=y
+CONFIG_I2C_HID_ACPI=y

Benefits to Home Assistant Community
Enables turn‑key Surface dashboards
Expands supported HAOS hardware
Improves kiosk reliability
Reduces need for custom user builds

This lowers the barrier for:
Wall‑mounted control panels
Dedicated HA touch interfaces
Re‑using existing Surface hardware
Request for Review

Feedback is welcome regarding:
Kernel config policy for device‑specific hardware
Firmware packaging approach
Inclusion scope for x86 HAOS targets

Happy to iterate or split this PR if maintainers prefer a staged approach.
Thank you for considering Surface support in Home Assistant OS.

## Release Notes

* **New Features**
  * Added comprehensive support for Microsoft Surface devices including Pro, Book, Laptop, and Go series
  * Enabled Type Cover backlight control and automatic tablet mode detection
  * Added dGPU power management for Surface Book 1
  * Implemented Intel Touch Host Controller driver for enhanced touch input
  * Improved Wi-Fi and Bluetooth performance across Surface devices
  * Added camera and sensor initialization improvements

* **Bug Fixes**
  * Fixed device power management and suspend/resume issues
  * Improved hardware initialization timing and device detection
  * Enhanced button and input device recognition


jens-maus:
> This pull request introduces complete Microsoft Surface device compatibility for Home Assistant OS (HAOS), specifically targeting reliable usage in ( HAOS Kiosk @puterboy ) deployments on Surface tablets, laptops, and PCs.

Sorry, but my two cents on this PR are: This would just unnecessarily bloat the whole OS environment of HAos up. IMHO the main purpose of HAos should be to provide a lightweight OS environment to run HAos primarily in a headless fashion with only its WebUI being accessible and a simple text console being displayed. So instead of trying to switch HAos into a head-aware system with a display why not simply generate/use a dedicated web-kiosk OS on your Microsoft Surface/Laptop/PC that simply boots up with a web browser in full screen instead directly connecting to the HAos WebUI? So, e.g. have a look at something like thinRoot (https://github.com/jens-maus/thinRoot) which is more targeted to provide a lightweight webkiosk OS that would enable to connect to your dedicated HAos running device and could be run on your Surface/Laptop/PC as a leightweight OS.

sairon (changes requested):
Please refer to our ADRs, especially the [ADR-0017](https://github.com/home-assistant/architecture/blob/master/adr/0017-hardware-screening-os.md) about requirements for the supported hardware. The patchset is way too massive, and it's not clear whether the changes wouldn't affect other hardware as well. This is one of the main reasons why this couldn't be accepted. I haven't tried to build the image to assess the actual bloat but otherwise Jens is quite right about the purpose of HAOS - this is just too much out of scope now.

Ideally, HAOS should have better support for 3rd party forks, which is something I'd like to work on soon and could make it simpler for you to distribute HAOS modified with any crazy ideas. However, I don't see a way these changes could be accepted here directly.

puterboy:
While I understand why this particular PR seems bloated given the number of kernel modules required just to enable a single brand of PC, I respectfully differ with @jens-maus 

Specifically, I think there are two types of HAOS users we need to consider:

1. **Sophisticated users** who have the capacity and footprint to set aside a dedicated device headless device to run HAOS OR have the sophistication to run HAOS in a VM on a non-headless server and configure a separate kiosk app on the same server to drive a display. (I imagine that @jens-maus and most devs fall in this class)

2. **Less sophisticated users** who may be environmentally conscious, space constrained, or money constrained and are just looking for a simple, all-inclusive server+cllient **appliance** they can install on their RPi or miniPC without having to invest in another device to act as display or figure out how to use VMs.

**Given the power of current RPis (let alone Mini-PC's and laptops), it seems to me to be a waste NOT to allow one to attach a display to your RPi or miniPC or tablet so that you can have a single, easy-to-install appliance + kiosk that just works out of the box** -- especially if you are giving new life to an old laptop or tablet that already has a beautiful screen attached!!!

While I may be biased as the author of HAOSKiosk, I like that the the add-on is dead simple to install and configure without requiring any additional hardware, software or commercial license (like some of the Android kiosk apps)

**If the goal is to make HA accessible to even casual users, it seems like supporting '# 2' is a good goal.**
Indeed, given all the effort being made to de-emphasize or even remove yaml support for configurations, it seems like we need to at least consider use case '# 2' for those who are more practical than purist or expert in how they allocate servers vs. clients.

puterboy:
> Ideally, HAOS should have better support for 3rd party forks, which is something I'd like to work on soon and could make it simpler for you to distribute HAOS modified with any crazy ideas. However, I don't see a way these changes could be accepted here directly.

It would be great if HAOS could be compiled to support nearly any module while not necessarily including every module in the base distribution -- either creating a separate "fat" version with all the modules or allowing users to add modules from a library.

That being said, given the size of sdcards and ssd's these days, it's hard to believe that even a few dozen or hundred megabytes of kernel modules make much of a difference -- and unused modules require no perceptible amount of additional memory or CPU to support.

puterboy:
@keviinh I truly respect all the work you have put into compiling your own custom kernel to make this all work for you on a an MS Surface -- but to the extent that this PR requires not just a slew of new modules but also several kernel patches, that makes me tend to agree with the devs that this is not a "simple" change without potential repercussions elsewhere -- to my non-expert opinion, it seems like this is more a MS Surface issue that OOTB Linux (and not just HAOS) does not support.

I say this because I don't want you to be frustrated and I truly appreciate all you have done to make this work for you.

If making it work on a Surface, could be reduced to a small list of modules, that would be easier to to push for your PR as low risk, low bloat in return for supporting a valid user need. But when it comes to kernel patches, I sympathize with the devs as they both are potentially dangerous as well as require continued maintenance as the kernel evolves.

keviinh:
Hi all, thanks again for the feedback.

I’ve updated the PR and incorporated the fixes that were suggested.
From my testing, I don’t believe the few extra megabytes introduced here meaningfully bloat HAOS, especially compared to the functional gain on supported hardware.

That said, I’m absolutely open to refining this further. Now that I have a fully working custom HAOS build, I can:
- review which patches are strictly necessary,
- evaluate whether some options can be switched from =y to =m,
- and reduce the overall delta to the minimum required for Surface devices.

As @puterboy  also mentioned, this enables a relatively low-cost setup for users who want a single device that acts as both their Home Assistant system and a touchscreen tablet/dashboard.

Alternatives do exist, i've tried them (Supervised installs with kiosk setups, or HAOS in Proxmox with GPU passthrough), but in practice these approaches are significantly more complex and not very approachable for most users.

My motivation here is aligned with the goal of making HAOS accessible to as many users as possible, while keeping both cost and complexity low. I’m happy to iterate further based on maintainer guidance.

Thanks for taking the time to review this.

jens-maus:
Sorry, but I still have the feeling that (even thought this might only increase the image size by a few MB) this might slightly shift the minimalistic focus of HAos and its underlying buildroot-based Linux environment. As said, HAos was IMHO never meant to provide a graphical user interface apart from a simple text console for maintenance reasons and a WebUI. Opening it up to actually provide a full display and having it targeted for mobile devices like the Microsoft Surface or other similar devices would significantly shifts its main purpose (provide a minimalistic and stable operating system for a 24/7 operation for a headless device like a HA yellow, RaspberryPi, etc. independently from the actual machine that loads the WebUI).

So, I am more on the side of @sairon here in arguing that the time would be IMHO better invested in more easily allowing other developers like you to actually better maintain an own fork of HAos targeted for different purposes. So instead of adding more and more hardware devices or kernel options for different HAos use cases within this main HAos project, interested parties like your HAOS Kiosk group should think about forking HAos for your own dedicated purpose. So if @sairon is anyway about to add some infrastructure to HA to more easily allow such HAos forks to be maintained and supply e.g. own updates, etc. this would be IMHO the more sustainable approach here.

puterboy:
> HAos was IMHO never meant to provide a graphical user interface apart from a simple text console for maintenance reasons and a WebUI.

That's why this is an add-on :) Add-ons allow HA to be expanded to support functionality far beyond the original conception of HA.

>  So if @sairon is anyway about to add some infrastructure to HA to more easily allow such HAos forks to be maintained and supply e.g. own updates, etc. this would be IMHO the more sustainable approach here.

Not a fan of forks in general as it makes it almost impossible for non-sophisticated users who don't understand forks or who don't want to have to choose between conflicting forks.

I think a better solution would be to compile HAOS with hooks for all modules -- *but* not necessarily include all the modules themselves in the basic release -- rather allow them to be added separately if the devs are too concerned about kernel bloat -- Debian and ubuntu both do this! -- For example have an "extra" package as an HAOS install option.

Regarding adding kernel patches, I agree that one needs to be careful to only add those that support common hardware and core use cases as patch management is complex and can have unforeseen adverse consequences for other users.

agners:
> I think a better solution would be to compile HAOS with hooks for all modules -- _but_ not necessarily include all the modules themselves in the basic release -- rather allow them to be added separately if the devs are too concerned about kernel bloat -- Debian and ubuntu both do this! -- For example have an "extra" package as an HAOS install option.

Unlike general purpose distros, Home Assistant OS uses a read-only file system, which is a single image. Unfortunately, it is not straight forward to deploy additional kernel modules separately.

> Regarding adding kernel patches, I agree that one needs to be careful to only add those that support common hardware and core use cases as patch management is complex and can have unforeseen adverse consequences for other users.

The unforseen side effects are one aspect. Another is the maintenance burden on rebasing those when we update kernels.

As @sairon mentions, our ADRs, especially the [ADR-0017](https://github.com/home-assistant/architecture/blob/master/adr/0017-hardware-screening-os.md) really put this PR outside of the scope of upstream HAOS.

That said, since a while we have that idea to make it easier to maintain HAOS forks. This would enable forks to be a test bed for specific boards or new OS features.  A while back #2943 made it easier to handle certificates required for the update system RAUC. But ultimately, we'd like to enable OS updates for forks, while still getting updates for the full higher level Home Assistant stack from our upstream version management. I've created a task to track progress on this effort: https://github.com/home-assistant/supervisor/issues/6587.

Closing this PR since it is out of scope for upstream HAOS.

PR #4565, opened 2026-03-05
Do not archive OS image artifacts for unpublished builds

sairon:
actions/upload-artifact v7 added possibility to disable creating a ZIP archive when only a single file is uploaded. Since this behavior is not desired for the images, as they're already compressed, disable it (default is enabled). The action/download-artifact in the test workflow should handle it gracefully based on mimetype.

## Summary by CodeRabbit

* **Chores**
  * Modified build artifact upload configuration in CI/CD workflows, adjusting how build outputs are processed during artifact storage.


agners (approved):
Nice, perfect feature for our use case! LGTM!

PR #4572, opened 2026-03-09
Fix retry in hassio container fetching, retry with backoff

sairon:
The retry when fetching containers from the registry didn't work because the script was executed with `set -e`. Capture the error code also for non-zero exit status.

Also use while loop instead of recursion and back off exponentially - start with 5s and multiply by 3 (i.e. 5s, 15s, 45s - waiting in total up to 1 minute for the registry to recover).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved container image fetch retry mechanism with exponential backoff to enhance reliability during transient failures.


agners (approved):
Nice catch!

PR #4589, opened 2026-03-19
Drop explicit IPv6 forwarding sysctl

agners:
Remove net.ipv6.conf.all.forwarding=1 from 60-otbr-ip-forward.conf and rely on Docker to enable IPv6 forwarding instead, just as we already rely on it for IPv4 forwarding (needed for NAT64 in OTBR). When this sysctl was added (d9ec60316), Docker did not enable IPv6 by default. Since Docker 27 (April 2024), IPv6 support — including ip6tables — is enabled by default, and Docker enables IPv6 forwarding at startup just like it does for IPv4.

Importantly, when Docker enables forwarding itself (rather than finding it already on), it also sets the FORWARD chain policy to DROP as a safety measure, Pre-enabling the sysctl prevents this, leaving the IPv6 FORWARD chain at ACCEPT. By removing our sysctl, we get the same protective DROP policy for IPv6 that we already benefit from for IPv4.

All Docker networks have separate rules which protect the network by default, so the default `ACCEPT` is not problematic for the `docker0` or `hassio` network.

The OpenThread Border Router app by default sets the `FORWARD` policy to `ACCEPT` as part of its startup script, so this does not break installations using the OpenThread Border Router app.

## Summary by CodeRabbit

* **Chores**
  * Adjusted system network configuration by removing IPv6 forwarding initialization setting. This change impacts automatic network routing behavior during system startup.

### DIFF
--- a/buildroot-external/rootfs-overlay/usr/lib/sysctl.d/60-otbr-ip-forward.conf
+++ b/buildroot-external/rootfs-overlay/usr/lib/sysctl.d/60-otbr-ip-forward.conf
@@ -1,1 +1,0 @@
-net.ipv6.conf.all.forwarding = 1


sairon (approved):

agners:
> When this sysctl was added ([d9ec603](https://github.com/home-assistant/operating-system/commit/d9ec60316474171b54ae06361cac0ab78f2d8c51)), Docker did not enable IPv6 by default. Since Docker 27 (April 2024), IPv6 support — including ip6tables — is enabled by default, and Docker enables IPv6 forwarding at startup just like it does for IPv4.

So turns out, this is not entirely accurate: What Docker 27 enabled by default is that ip6table rules get inserted if necessary. But this only happens if at least one Docker network bridge has IPv6 enabled. This is essentially what https://github.com/home-assistant/supervisor/pull/6720 now achieves (part of Supervisor 2026.04.0).

What we could do is enable IPv6 using `"ipv6": true` in `daemon.json`. This would make the default bridge (which is essentially unused other than for the Supervisor itself) have IPv6 enabled, which then also makes Docker enable IPv6 forwarding by default, independent of the hassio IPv6 configuration.

PR #4594, opened 2026-03-20
Fix skopeo for ARM images in hassio after builder changes

sairon:
Afer builder changes, ARM images are now correctly published with their platform, and when skopeo is used to inspect/pull the image on x86 without any other flags, it fails with:

Error parsing manifest for image: Error choosing image instance: no image found in image index for architecture amd64, variant "", OS linux

Pass the correct arch in skopeo operations to fix that.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved container image retrieval to correctly identify and handle different processor architectures, ensuring proper compatibility across various supported systems.

### DIFF
--- a/buildroot-external/package/hassio/fetch-container-image.sh
+++ b/buildroot-external/package/hassio/fetch-container-image.sh
@@ -42,7 +42,13 @@ image_tag=$(jq -e -r --arg image_json_name "${image_json_name}" \
 	'.[$image_json_name]' < "${version_json}")
 full_image_name="${image_name}:${image_tag}"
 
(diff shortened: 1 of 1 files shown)


agners (approved):
Make sense, since we run this on x86-64. This is also compatible with the old images is it?

sairon:
> Make sense, since we run this on x86-64. This is also compatible with the old images is it?

In fact, it would fail the other way around (for ova and generic-x86-64) on ARM64 :) 

The old images had no platform tag, so you can use whatever `--override-arch` and get the same result.

PR #4599, opened 2026-03-24
Fix handling of hassio container updates in tests

sairon:
When running tests on an image that contains older hassio components (Supervisor or plugins), the autoupdate may interfere with the test run. To avoid this, patch Suprvisor updater config as early as possible and restart Supervisor.

For Supervisor tests, we need all components to be updated, so parametrize the supervisor update test to update all plugins too.

## Summary by CodeRabbit

* **Tests**
  * Enhanced test setup to disable automatic updates and restart services during initialization for a controlled testing environment
  * Improved system readiness checks to properly handle additional connection conditions
  * Expanded test coverage for component updates, now testing supervisor, audio, cli, dns, observer, and multicast components

### DIFF
--- a/tests/smoke_test/test_basic.py
+++ b/tests/smoke_test/test_basic.py
@@ -10,6 +10,14 @@ _LOGGER = logging.getLogger(__name__)
 @pytest.mark.dependency()
 @pytest.mark.timeout(120)
 def test_init(shell):
+    # Disable auto-updates to avoid interference with other tests,
+    # do it directly on config level and restart Supervisor via systemd.
+    shell.run_check(
+        "jq '.auto_update = false' /mnt/data/supervisor/updater.json > /tmp/updater.json"
+        " && mv /tmp/updater.json /mnt/data/supervisor/updater.json"
+        " && systemctl restart hassos-supervisor.service"
+    )
+
     def check_container_running(container_name):
         out = shell.run_check(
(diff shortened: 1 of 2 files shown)


tests/smoke_test/test_basic.py
```diff
@@ -10,6 +10,14 @@
 @pytest.mark.dependency()
 @pytest.mark.timeout(120)
 def test_init(shell):
+    # Disable auto-updates to avoid interference with other tests,
+    # do it directly on config level and restart Supervisor via systemd.
+    shell.run_check(
+        "jq '.auto_update = false' /mnt/data/supervisor/updater.json > /tmp/updater.json"
+        " && mv /tmp/updater.json /mnt/data/supervisor/updater.json"
+        " && systemctl restart hassos-supervisor.service"
```
agners, line 18: Hm, I guess this is racy in theory, but fast enough typically? Would be kinda cool if we can inject files into the data dir before startup 🤔

agners (approved):
Seems sensible approach to make update in tests more predictable, so LGTM!

tests/smoke_test/test_basic.py
sairon: In theory it is, but takes at least couple of seconds before Supervisor starts to update and the unit restarts prevents it. I ran it couple of times and it worked well, so I'm fine with that. At least it's better than the state before when the auto-update could happen any time depending on network and other aspects.

PR #4643, opened 2026-04-14
Bump generic-raw-uart to 1.33 and add kernel 6.14+ patches

jens-maus:
This PR bumps the `generic-raw-uart` package (required for OpenCCU compatibility) to version 1.33 and adds a patch on top of it to introduce kernel 6.14+ compatibility so that the generic-raw-uart kernel driver works with newer kernel versions. This refs https://github.com/alexreinert/piVCCU/pull/575

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed emulated termios2/glibc 2.42 compatibility layer from the raw UART driver.

* **Chores**
  * Updated raw UART kernel driver source to support Linux kernel 6.14 and later.
  * Updated driver source revision and build configuration.


sairon (approved):
Looks good, thanks, Jens! 👍

PR #4683, opened 2026-05-05
Prepare 17.3

sairon:
- **Linux: Update kernel to 6.12.85**
- **Backport algif_aead patches to RPi 6.12.75 kernel**
- **Bump OS to pre-release version 17.3.rc1**
- **Bump OS to release version 17.3**


jens-maus:
Sorry for asking back here @sairon, but as I am also curious to find out how to tackle the effects of the copy.fail vulernability in the context of HomeAssistant. So why don't you simple remove the `algif_aead` kernel module/functionality altogether? In the Context of OpenCCU I never had it enabled and thus OpenCCU is/was never affected. Is there really some functionality in HA that relies on that `algif_aead` kernel module/functionality. If not, I feel that it would be perhaps more straight forward to simply disable the corresponding kernel module or for the time being (until kernel 6.18.x is integrated in the next HA release) to establish the RedHat workaround by blacklisting the kernel module (https://access.redhat.com/security/cve/cve-2026-31431) via `initcall_blacklist=algif_aead_init` in the kernel cmdline.

agners (approved):

PR #4693, opened 2026-05-08
Prepare 17.4.rc1

agners:
- Bump OS to release version 17.3
- Linux: Update kernel to 6.12.87
- Backport ESP fix patch to RPi 6.12.75 kernel
- Bump OS to pre-release version 17.4.rc1


sairon (commented):
Per https://github.com/V4bel/dirtyfrag#why-did-you-chain-two-vulnerabilities, we need at least one path to be explotable:

> xfrm-ESP Page-Cache Write provides a powerful arbitrary 4-byte STORE primitive like Copy Fail, and is included on most distributions, but it requires the privilege to create a namespace.
> 
> Ubuntu sometimes blocks unprivileged user namespace creation through AppArmor policy. In such an environment, xfrm-ESP Page-Cache Write cannot be triggered. RxRPC Page-Cache Write does not require the privilege to create a namespace, but the rxrpc.ko module itself is not included in most distributions. However, on Ubuntu, the rxrpc.ko module is loaded by default.
> 
> Chaining the two variants makes the blind spots cover each other, allowing root privileges to be obtained on every major distribution. For details, refer to the technical details document.

The first path (fixed by the included/backported patch) is not generally exploitable because users in unprivileged containers can't create namespaces (AFAIK the container would need at least `CAP_SYS_ADMIN`).

The second path is not viable either, because we don't enable `CONFIG_AF_RXRPC` in our kernel configs. 

Seems that this vulnerability doesn't hit HAOS as hard as Copy Fail.

agners:
Ok then, not worth the churn!

jens-maus:
> Seems that this vulnerability doesn't hit HAOS as hard as Copy Fail.

You are right, HAos is not really affected by the new "dirtyfrag" vulernability AFAICS.

However, don't want to annoy you @sairon and @agners, but IMHO it would be a better approach to simply install the known `/etc/modprobe.d/dirtyfrag.conf` and `/etc/modprobe.d/disable-algif_aead.conf` workarounds for Copy.Fail and this new `dirtyfrag` which simply prevent the vulernable kernel modules to be loaded at all. AFAICS they are anyway all not really necessary for HAOS. So instead of patching the kernel it would be IMHO better to concentrate on the upcoming v18 HAOS version and simply install the modprobe.d workarounds there or simply remove these unnecessary kernel modules altogether. But of course, these are just my two cents ;-)

sairon:
> IMHO it would be a better approach to simply install the known `/etc/modprobe.d/dirtyfrag.conf` and `/etc/modprobe.d/disable-algif_aead.conf` workarounds

I don't agree. Why should we pile up workarounds if there's an actual fix? Updating the kernel or backporting a fix isn't much more time consuming than adding modprobe flags. And it's more self-documenting as well.

> AFAICS they are anyway all not really necessary for HAOS.

Not for the core function, but the same reason why we backported the Copy Fail fix (which isn't exploitable in "standard HAOS use-case") applies here: there might be apps where it could be useful, and disabling these will disable/cripple IPSec for instance. I prefer to trust kernel devs who decide what should be enabled by default, and just the fact there was historically exploitability of a particular module doesn't justify disabling it.

PR #4701, opened 2026-05-11
Update build container to Debian Trixie

sairon:
The image for the build container was still using Bullseye which will be EOL this August. Update to Trixie and adjust Dockerfile accordingly.

## Chores
* Updated base operating system in Docker image from Debian bullseye to trixie.
* Modernized Docker repository configuration and setup process.
* Adjusted build dependencies to align with the updated system configuration.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,20 +1,24 @@
-FROM debian:bullseye
+FROM debian:trixie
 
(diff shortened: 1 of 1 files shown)


agners (approved):
Nice, LGTM!

Issue #2675, opened 2023-08-08
Watchdog can freeze ODROID-M1 and ODROID-XU4 (maybe others)


agners:
### Describe the issue you are experiencing

Since #2628 watchdog is enabled in systemd.

It seems that when stressing the system with `stress-ng`, it can trigger a watchdog reset by systemd. Typically, this should lead to a reboot. However, it seems that certain embedded boards have bugs in their watchdog implementation or elsewhere, which causes the system to not reboot but just freeze.

This has been observed with ODROID-M1 and ODROID-XU4

### What operating system image do you use?

odroid-m1 (Hardkernel ODROID-M1)

### What version of Home Assistant Operating System is installed?

11.0.dev20230803

### Did you upgrade the Operating System.

Yes

### Steps to reproduce the issue

1. Boot the board with the dev version
2. Use `stress-ng --all 4` to stress the system a lot
3. Wait for 5-10 minutes until a the board stops responding

A easier way is to use a stable release (e.g. 10.4) which does not have systemd watchdog enabled:
1. Use `cat /dev/watchdog`  to start the watchdog (this will lead to `cat: read error: Invalid argument`, but at this point the watchdog has been started because the device file has been opened.
2. Wait for a freeze to happen

### Anything in the Supervisor logs that might be useful for us?

```txt
No
```

### Anything in the Host logs that might be useful for us?

```txt
Some board warn that the watchdog continues to run after catting the file:


Aug 08 21:03:35 ha-shelf2-om1 kernel: watchdog: watchdog0: watchdog did not stop!
```
```


### System information

_No response_

### Additional information

_No response_

mingzhangqun:
Hi, everyone. I've tried enable BR2_INIT_BUSYBOX and disable BR2_INIT_SYSTEMD, the watchdog freeze also.
So I assume it has nothing to do with systemd.

[watchdog.txt](https://github.com/home-assistant/operating-system/files/12474480/watchdog.txt)

mingzhangqun:
Using the offical source code (linux 5.10) the watchdog works well. I'm tring the newest offical sdk(linux 6.1).

mingzhangqun:
I've tried official sdk(linux 6.1), the watchdog works well.
I replaced the kernel partition(`dd if=kernel.squash of=/dev/sdx2`) with haos, the watchdog works too.
[kernel_img.zip](https://github.com/home-assistant/operating-system/files/12484672/kernel_img.zip)

PR #4705, opened 2026-05-12
systemd: Increase runtime watchdog timeout to 5 minutes

agners:
## Summary

- Bump `RuntimeWatchdogSec` from `default` (driver-dependent: ~30s on iTCO_wdt/i6300esb, 15s on BCM2835) to a uniform `300` seconds.
- Motivation: PID1 can stall in `nfs_getattr` for an NFS RPC major-timeout while running `mount_enter_mounting()` (chase/open_tree, is_dir, mkdir_p_label, unit_warn_if_dir_nonempty), which exceeds the previous hardware-default heartbeats and causes spurious reboots on network-storage stalls rather than genuine kernel hangs. Reported by https://github.com/home-assistant/supervisor/issues/6827 and upstream as systemd/systemd#42050.
- Hardware-limit wrinkle: drivers using `max_hw_heartbeat_ms` (notably BCM2835 since kernel v6.8, commit [f33f5b1fd1be](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?h=v7.1-rc3&id=f33f5b1fd1be)) accept any userspace timeout and let the watchdog core multiplex hardware pings in the background, so 300s is honored on all Pi variants. Drivers still using `max_timeout` directly will fall back to their own value — no worse than the previous behavior.

Issue #2675 (ODROID-M1 / Green) remains addressed via disabled `CONFIG_DW_WATCHDOG`; this change has no impact on those boards since they expose no `/dev/watchdog`.

## Test plan

- [x] Boot generic-x86-64 (iTCO_wdt); verify `systemctl show | grep -i watchdog` reports `RuntimeWatchdogUSec=5min` and `WatchdogLastPingTimestamp` advances.
- [x] Boot OVA on KVM with `-watchdog i6300esb`; same check.
- [ ] Boot Raspberry Pi 4 / 5; `cat /sys/class/watchdog/watchdog0/timeout` should return 300 (post-v6.8 BCM2835 driver multiplexing).
- [x] Reproduce systemd#42050 scenario (NFS mount with unreachable server, `systemctl reload` then `restart`); confirm the box stays up instead of rebooting during the ~60s `nfs_getattr` D-state.
- [x] Genuine PID1 hang (e.g. `kill -STOP 1` in a throwaway VM) still reboots within ~5min + driver pre-timeout.

## Summary by CodeRabbit

* **Chores**
  * Updated system runtime watchdog timeout configuration to 300 seconds.

### DIFF
--- a/buildroot-external/rootfs-overlay/usr/lib/systemd/system.conf.d/watchdog.conf
+++ b/buildroot-external/rootfs-overlay/usr/lib/systemd/system.conf.d/watchdog.conf
@@ -1,3 +1,3 @@
 [Manager]
-RuntimeWatchdogSec=default
+RuntimeWatchdogSec=300
 WatchdogDevice=/dev/watchdog


agners:
> * [x]  Genuine PID1 hang (e.g. `kill -STOP 1` in a throwaway VM) still reboots within ~5min + driver pre-timeout.

systemd/pid 1 does not react on `SIGSTOP`. Had to use gdb to simulate:

```
docker run --rm -it \
      --pid=host \
      --cap-add=SYS_PTRACE \
      --security-opt seccomp=unconfined \
      --security-opt apparmor=unconfined \
      debian:trixie-slim \
      bash -c 'apt-get update -qq && apt-get install -y -qq gdb && gdb -p 1'
```

Confirmed the watchdog still works.

agners:
Note that Proxmox has no Watchdog configured by default, and there is no graphical way to setup one (as of Proxmox 9.1.9 anyways). The CLI allows to setup a Watchdog using:

```
qm set <vmid> --watchdog model=i6300esb,action=reset
```

sairon (approved):
What a skeleton hiding in the closet, nice find! 😅

PR #4707, opened 2026-05-13
Backport U-Boot patch for Raspberry Pi MMC init on BCM2712

sairon:
This backports patch currently waiting for a merge upstream that fixes issue we saw with CM5 on Yellow, where the SD card init took over 120 seconds. We added a workaround in #3700 lowering this roughly to 20 seconds, but this is no longer needed, so revert this patch as well.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed bootloader eMMC initialization to properly detect non-removable devices and prevent unintended SD pin configuration changes that could affect device functionality.

* **Revert**
  * Restored SD/SDIO support on Raspberry Pi Yellow boards by reverting a prior configuration restriction.

### DIFF
--- a/buildroot-external/board/raspberrypi/yellow/patches/linux/0024-Revert-ARM-dts-bcm2712-yellow-Disable-SD-SDIO-modes-.patch
+++ b/buildroot-external/board/raspberrypi/yellow/patches/linux/0024-Revert-ARM-dts-bcm2712-yellow-Disable-SD-SDIO-modes-.patch
@@ -1,0 +1,36 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Jan=20=C4=8Cerm=C3=A1k?= <sairon@sairon.cz>
+Date: Tue, 12 May 2026 16:56:52 +0200
+Subject: [PATCH] Revert "ARM: dts: bcm2712: yellow: Disable SD/SDIO modes on
+ eMMC interface"
+
+This reverts commit d7d55cabc452f1c03c4191412d3f3edde8d701d2.
+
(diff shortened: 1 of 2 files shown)


agners (approved):

buildroot-external/board/raspberrypi/yellow/patches/linux/0024-Revert-ARM-dts-bcm2712-yellow-Disable-SD-SDIO-modes-.patch
```diff
@@ -0,0 +1,36 @@
...
+-/*
+- * Yellow doesn't need to use eMMC interface as SD/SDIO, disable those
+- * modes to speed up boot. This is particularly handy to work around
+- * an issue with eMMC initialization but won't harm in general.
+- */
+-&sdio1 {
+-	no-sdio;
+-	no-sd;
+-
+-	/delete-property/ supports-cqe;
```
sairon, line 31: Note that while this removes this line, there's still another line removing it in the final DTS after all patches apply:
```
$ grep -rn cqe arch/arm64/boot/dts/broadcom/bcm2712-rpi-cm5-ha-yellow.dts
240:            /delete-property/ sd_cqe;
```

So this PR does NOT re-enable CQE (but my plan is to do it in another PR).

sairon: > but my plan is to do it in another PR

Actually, I won't. I tested it on CM5 with rpi5-64 image which uses the EEPROM bootlader and it still triggers the issue:
image

Will need to be investigated separately...

PR #4714, opened 2026-05-17
Use F2FS file system for the Data and Overlay partitions

Extravert-ir:
The f2fs filesystem should improve the longevity of SD cards and eMMC memory which is mostly used on all kinds of HAOS hardware. F2FS may also improve the overall performance as shown by some [benchmarks](https://www.phoronix.com/review/linux-615-filesystems).

The mounting rules were also changed so they can mount both ext4 and f2fs partitions.

This pull request is a RFC. If developers don't want to be so radical and change the default FS, we may just leave the mounting rules change so people (like me) can continue using a custom-formatted sdcard.

## Summary by CodeRabbit

* **Chores**
  * Migrated system storage partitions from ext4 to F2FS filesystem
  * Updated system mount configurations to support F2FS
  * Added required F2FS utilities to build tools

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -373,7 +373,7 @@ jobs:
           echo "## Partitions" >> $GITHUB_STEP_SUMMARY
           echo "| File | Size (bytes) | Size (formatted) |" >> $GITHUB_STEP_SUMMARY
           echo "|:-|:-|:-|" >> $GITHUB_STEP_SUMMARY
-          for f in boot.vfat kernel.img rootfs.erofs overlay.ext4 data.ext4; do
+          for f in boot.vfat kernel.img rootfs.erofs overlay.f2fs data.f2fs; do
             echo "| ${f} | $(du -b output/images/$f | cut -f1) | $(du -bh output/images/$f | cut -f1) |" >> $GITHUB_STEP_SUMMARY
           done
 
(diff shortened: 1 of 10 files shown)


sairon (commented):
I am personally not that much sold on F2FS, and while the actual longevity and tear isn't easy to measure, I believe that correctly tuned ext4 (especially with discard and longer commit interval) isn't too much more unfriendly to flash either. At the same time, what really bothers us, is the resilience and robustness of the FS, and that is supposedly _lower_ for F2FS. While the evidence may be anecdotal here, I'm hesitant to make the switch, as `fsck.f2fs` seems to be far less mature than `e2fsck` too.

I wouldn't have any objections against making the mount units more flexible as you suggest, but there's one easy to miss but major issue, and that's the removal of `commit=30` option for the data partition. There's in fact quite some work behind that in #2103 (mainly with the goal to do what you want - lower the wear of the flash storage) and this change will drop that. So if we should proceed with that, we need to find a way to preserve the current behavior.

Extravert-ir:
Ok, I have an idea on how to make mounts more flexible and preserving options at the same time.
I'll update the PR in a couple of days

sairon (changes requested):
@Extravert-ir Any progress, or hints how would you like to do it? We're closing in on the HAOS 18.0 milestone, so if you want to have that in, it'd be right time to do it soon.

kharris31576-sudo (approved):

PR #4719, opened 2026-05-19
Add pre-shutdown phase to coordinate shutdown with Supervisor

### LINKED ISSUE #4584: Handle CLI shutdown on tty1
When restarting the system via user interface, shutdown of the console is not handled gracefully: The CLI on tty1 gets killed since we shutdown the CLI plug-in. The systemd service then tries to restart the CLI multiple times. This probably slows down the shutdown process overall as well.

### LINKED ISSUE #4642: Improve systemd shutdown ordering for more graceful shutdowns
Currently when initiate a shutdown many services get stopped at once, which doesn't allow an orderly shutdown of Home Assistant Core and other containers.

The shutdown procedure should be coordinated with Supervisor.

See also https://github.com/home-assistant/supervisor/pull/6631#issuecomment-4068129770.

sairon:
Introduce haos-pre-shutdown.target, a phase Supervisor enters before a API-triggered reboot or power off. ha-cli@.service conflicts with the target, so the console CLI is stopped by a regular stop job instead of restart-looping once Supervisor stops the CLI plugin (#4584).

While the CLI is down, haos-shutdown-notice@.service takes over its tty and shows a shutdown notice, so the console no longer shows just a blinking cursor. It is a template enabled through ha-cli@.service's Also=, keeping the console instance defined in a single place.

hassos-supervisor.service gets a longer "docker stop" timeout. As the unit is already ordered after docker.service, Supervisor can now stop Home Assistant Core, add-ons and plugins gracefully on SIGTERM before Docker tears the containers down (#4642).

Supervisor will the target and perform the graceful shutdown via a separate change.

Closes #4584, closes #4642

## Summary by CodeRabbit

* **New Features**
  * Console shutdown/restart notifications shown on system consoles during shutdown.
  * A pre-shutdown phase that ensures services are stopped in a controlled order before reboot/shutdown.

* **Improvements**
  * More graceful service termination with a longer, explicit stop timeout to reduce abrupt container/service exits.
* **Quality**
  * Ordering tweaks to avoid console gaps during shutdown notices.

### DIFF
--- a/buildroot-external/rootfs-overlay/usr/lib/systemd/system/haos-pre-shutdown.target
+++ b/buildroot-external/rootfs-overlay/usr/lib/systemd/system/haos-pre-shutdown.target
@@ -1,0 +1,7 @@
+[Unit]
+Description=Home Assistant OS pre-shutdown phase
+# Started explicitly by Supervisor before an API-triggered reboot/shutdown so
+# that units conflicting with it (e.g. ha-cli@.service) are stopped cleanly
+# before Supervisor stops the matching plugin containers.
+DefaultDependencies=no
+RefuseManualStop=yes
--- a/buildroot-external/rootfs-overlay/usr/lib/systemd/system/hassos-supervisor.service
+++ b/buildroot-external/rootfs-overlay/usr/lib/systemd/system/hassos-supervisor.service
@@ -15,7 +15,12 @@ Restart=always
 RestartSec=5s
 ExecStartPre=-/usr/bin/docker stop hassio_supervisor
 ExecStart=/usr/sbin/hassos-supervisor
-ExecStop=-/usr/bin/docker stop hassio_supervisor
+# This unit is ordered After=docker.service, so on host shutdown it is
+# stopped before Docker tears containers down. Supervisor handles the
+# SIGTERM from "docker stop" by gracefully stopping Home Assistant Core,
+# apps and plugins; allow ample time for that to complete.
+ExecStop=-/usr/bin/docker stop -t 420 hassio_supervisor
+TimeoutStopSec=450s
 
 [Install]
 WantedBy=multi-user.target
--- a/buildroot-external/rootfs-overlay/usr/libexec/haos-shutdown-notice
+++ b/buildroot-external/rootfs-overlay/usr/libexec/haos-shutdown-notice
@@ -7,20 +7,14 @@
 # without this the console would just show a blinking cursor.
 # ==============================================================================
 
-# Move to the top, clear the screen and hide the cursor.
-printf '\033[H\033[2J\033[?25l'
+# Move to the top and clear the screen.
+printf '\033[H\033[2J'
 
 cat <<'EOF'
 
 
     Home Assistant is shutting down...
 
-    This can take a moment, the system will power off or restart shortly.
+    This can take a moment, the system will power off or reboot shortly.
 
 EOF
-
-# Keep the service alive so the notice stays on screen until Systemd kills
-# this service and starts printing its messages.
-while true; do
-    sleep 3600
-done
--- a/buildroot-external/rootfs-overlay/usr/libexec/haos-shutdown-notice
+++ b/buildroot-external/rootfs-overlay/usr/libexec/haos-shutdown-notice
@@ -1,0 +1,26 @@
+#!/bin/sh
+# ==============================================================================
+# Display a shutdown notice on the console.
+#
+# Started via haos-pre-shutdown.target when a reboot or power off is triggered
+# via API. The Home Assistant CLI on default tty is stopped at that point, so
+# without this the console would just show a blinking cursor.
+# ==============================================================================
+
+# Move to the top, clear the screen and hide the cursor.
+printf '\033[H\033[2J\033[?25l'
+
+cat <<'EOF'
+
+
+    Home Assistant is shutting down...
+
+    This can take a moment, the system will power off or restart shortly.
+
+EOF
+
+# Keep the service alive so the notice stays on screen until Systemd kills
+# this service and starts printing its messages.
+while true; do
+    sleep 3600
+done
--- a/buildroot-external/rootfs-overlay/usr/lib/systemd/system/ha-cli@.service
+++ b/buildroot-external/rootfs-overlay/usr/lib/systemd/system/ha-cli@.service
@@ -2,12 +2,12 @@
 Description=Home Assistant CLI
 Wants=hassos-supervisor.service
 After=systemd-user-sessions.service plymouth-quit-wait.service getty-pre.target hassos-supervisor.service
-Conflicts=getty@%i.service
+Conflicts=getty@%i.service haos-pre-shutdown.target
 
 # If additional gettys are spawned during boot then we should make
 # sure that this is synchronized before getty.target, even though
 # getty.target didn't actually pull it in.
-Before=getty.target
+Before=getty.target haos-pre-shutdown.target
 
 # IgnoreOnIsolate causes issues with sulogin, if someone isolates
 # rescue.target or starts rescue.service from multi-user.target or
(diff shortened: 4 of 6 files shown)


buildroot-external/rootfs-overlay/usr/libexec/haos-shutdown-notice
```diff
@@ -0,0 +1,26 @@
...
+
+# Move to the top, clear the screen and hide the cursor.
+printf '\033[H\033[2J\033[?25l'
+
+cat <<'EOF'
+
+
+    Home Assistant is shutting down...
+
+    This can take a moment, the system will power off or restart shortly.
```
agners: We use `reboot` for system level restarts typically.

```suggestion
    This can take a moment, the system will power off or reboot shortly.
```

buildroot-external/rootfs-overlay/usr/lib/systemd/system/hassos-supervisor.service
```diff
@@ -15,7 +15,12 @@ Restart=always
 RestartSec=5s
 ExecStartPre=-/usr/bin/docker stop hassio_supervisor
 ExecStart=/usr/sbin/hassos-supervisor
-ExecStop=-/usr/bin/docker stop hassio_supervisor
+# This unit is ordered After=docker.service, so on host shutdown it is
+# stopped before Docker tears containers down. Supervisor handles the
+# SIGTERM from "docker stop" by gracefully stopping Home Assistant Core,
+# apps and plugins; allow ample time for that to complete.
+ExecStop=-/usr/bin/docker stop -t 420 hassio_supervisor
+TimeoutStopSec=450s
```
agners, line 23: Nice, yeah that make sense!

buildroot-external/rootfs-overlay/usr/lib/systemd/system/haos-pre-shutdown.target
```diff
@@ -0,0 +1,7 @@
+[Unit]
+Description=Home Assistant OS pre-shutdown phase
+# Started explicitly by Supervisor before an API-triggered reboot/shutdown so
+# that units conflicting with it (e.g. ha-cli@.service) are stopped cleanly
+# before Supervisor stops the matching plugin containers.
```
agners, line 5: Did you consider using `AllowIsolate=yes`? I think that would cause "not required" units to automatically shutdown.

buildroot-external/rootfs-overlay/usr/libexec/haos-shutdown-notice
```diff
@@ -0,0 +1,26 @@
...
+
+EOF
+
+# Keep the service alive so the notice stays on screen until Systemd kills
```
agners: Is it the service which needs to keep alive? Then maybe:

```
Type=oneshot
RemainAfterExit=yes
```
works.

Otherwise I guess this should read:

```suggestion
# Keep the process alive so the notice stays on screen until Systemd kills
```

agners (changes requested):
Nice, looking forward for more graceful shutdowns 😃

buildroot-external/rootfs-overlay/usr/lib/systemd/system/haos-pre-shutdown.target
sairon: IIUC that's not what we want - it would only enable running it with `systemctl isolate` (it does nothing on its own) but then we'll have to declare units we want to stay alive in the pre-shutdown phase, otherwise Systemd will take them out. This will conflict with the graceful shutdown sequence in Supervisor.

buildroot-external/rootfs-overlay/usr/libexec/haos-shutdown-notice
sairon: That's indeed what the comment meant. But you've got a point as well, oneshot service with RemainAfterExit is better suited here. Turns out the wait isn't needed at all (in fact it leads to log noise about failure without `SuccessExitStatus=SIGTERM`), so I removed it. I also left the cursor enabled, as without it, the system may appear frozen.

buildroot-external/rootfs-overlay/usr/lib/systemd/system/haos-pre-shutdown.target
agners: > but then we'll have to declare units we want to stay alive in the pre-shutdown phase, otherwise Systemd will take them out. This will conflict with the graceful shutdown sequence in Supervisor.

Yes, and I guess it would mean we want at least `DefaultDependencies` (speaking of which, do we need to say `no` there?).

The reason I bring it up I compare this `haos-pre-shutdown.target` to something like `multi-user.target` and `graphical.target`. Multi-user is everything running, but no UI, and `graphical.target` is then with the UI. Both have `AllowIsolate=yes`.


> it would only enable running it with `systemctl isolate` (it does nothing on its own)

I guess we can do this through D-Bus as well, if it's what it takes to make the target happen...

agners: Today we seem to have `multi-user.target` as `default.target`. I guess we could consider introducing a `haos.target`, which is similar to graphical target but pulls in the CLI on tty and make that as default.target. Then we'd isolate `multi-user.target` in Supervisor before shutting down the containers etc. 🤔 

But maybe it is better to do this as a follow up, not sure. I am trying to find "prior art" in systemd to what we are aiming to do here, and align with it.

sairon: The `haos.target` could be a way to handle the stage separation, yes. But it'd be more complicated to do what the current approach is doing (like showing the notice - which I think is very nice to have - otherwise Supervisor API shutdown leads to black screen with no feedback to user). `isolate` only helps with _stopping_ the units, so it would only handle the graceful CLI shutdown - we still need running Docker and Supervisor in `multi-user.target` and do the tear-down when it's shutting down. So I see no big benefit here, other than it might be slightly more idiomatic.

Also, I'm afraid that adding a new custom boot target is more risky than reordering the shutdown phase. Maybe we can reconsider in a follow-up if we find there's more we can get out of it, but currently I'm quite happy with the current state.

agners: I did think about this a bit more, and I'd like us to really reconsider what we are doing here.

Currently our shutdown/reboot sequence takes different paths depending on how a shutdown is trigggered. This actually lead that we only see some of the problems sometimes (e.g. the tty shutdown, it didn't show in ACPI/power button shutdown since this triggers a logind shutdown before a Supervisor coordinated shutdown). It makes testing harder, and is just plain unnecessary complexity.

Looking at https://github.com/home-assistant/supervisor/pull/6631#issuecomment-4067986795 you can see the paths I am alluding to.

Adding this `haos-pre-shutdown.target` adds another wrinkle to one of the shutdown/reboot sequences (UI triggered one), while it is entirely skipped by the ACPI/power button triggered flow.

Digging through it a bit with Claude/and all three PRs in context, I think we are actually quite close to simplify things and make things cleaner/neater at the same time:

Supervisor side:
- The Supervisor PR (#6851) already has the most important piece: the stop_supervisor() / `host_is_shutting_down()` SIGTERM handler that runs `core.shutdown()` when systemd state is stopping. This becomes the only shutdown path instead of one of two.
- In `supervisor/host/control.py`: simplify reboot() and shutdown() to call directly logind
- No `_enter_haos_pre_shutdown()`

HAOS:
- Remove `haos-pre-shutdown.target`
- The current `After=hassos-supervisor.service` should already stop `ha-cli@.service` correctly if shutdown/reboot is triggered through logind
- Change `haos-shutdown-notice@.service` to `Before=shutdown.target`, which I think should make this to start on shtudown. needs testing. Maybe also needs `Conflicts=getty@%i.service`.

Maybe this needs a gating depending on HAOS version, but given the cleanness/simplification its a tradeoff I take. Not introducing `haos-pre-shutdown.target` today also makes the transition less complex, I think if we'd merge the PRs as is today it would make this unification/simplification of reboot/shutdown flows harder.

agners (changes requested):
As discussed in https://github.com/home-assistant/operating-system/pull/4719#discussion_r3288631580.

sairon:
Replaced by #4736 and home-assistant/supervisor#6887.

Issue #4631, opened 2026-04-09
Add RPi EEPROM update feature


agners:
Make it possible to check the version of the FW on Raspberry Pi and update it using an update entity.

Update entity should have a warning that this updates firmware and the process must not be interrupted (maybe with links to Raspberry Pi docs?)!

iceymanz:
Exposing the firmware version as a sensor so something could run an update check would be great.

Flashing from a USB Booted pi4 seems to not be supported by the platform. Pi5's model is different so it may be suitable.

I'd like to offer an opinion that since I've had to pull an sdcard out, write an image to it and run 3 commands to fix this once in 3+ years - for the effort required  I'd suggest there is more value in exposing the value to the platform and or adding a compatibility check to hassios update process than building in a firmware update tool into HA.

hundredfire:
> Exposing the firmware version as a sensor so something could run an update check would be great.
> 
> Flashing from a USB Booted pi4 seems to not be supported by the platform. Pi5's model is different so it may be suitable.
> 
> I'd like to offer an opinion that since I've had to pull an sdcard out, write an image to it and run 3 commands to fix this once in 3+ years - for the effort required I'd suggest there is more value in exposing the value to the platform and or adding a compatibility check to hassios update process than building in a firmware update tool into HA.

From what I recall, only old bootloader versions required a microSD card to update eeprom (before mid-2020 iirc). I might be remembering wrong, but newer versions only require host access

dana-se:
> From what I recall, only old bootloader versions required a microSD card to update eeprom (before mid-2020 iirc). I might be remembering wrong, but newer versions only require host access

rpi-eeprom-update: CURRENT: Tue Jan 25 14:30:41 UTC 2022 (1643121041)
rpi-eeprom-update -a: rpi-eeprom-update is only available when booting from an SD card or if flashrom can be used

hundredfire:
> > From what I recall, only old bootloader versions required a microSD card to update eeprom (before mid-2020 iirc). I might be remembering wrong, but newer versions only require host access
> 
> rpi-eeprom-update: CURRENT: Tue Jan 25 14:30:41 UTC 2022 (1643121041) rpi-eeprom-update -a: rpi-eeprom-update is only available when booting from an SD card or if flashrom can be used

I stand corrected.. Thanks for the data point

iceymanz:
My 2c to put the effort of building firmware updates for HA's soc based platforms (I'm using the public reporting stats):
[...]

PR #4735, opened 2026-05-26
Patch rpi-eeprom-update to show info on CM4 and make findBootFS faster

sairon:
On CM4, rpi-eeprom-update only prints warning if the env variable enabling self-update isn't set. In that case, also print bootloader info so we can propagate it further.

Also, add patch making findBootFS faster, avoiding scan of all block devices.

Refs #4631

## Summary by CodeRabbit

* **Bug Fixes**
  * Adjusted EEPROM update flow for CM4 devices to continue operation when updates are not enabled, preventing unintended early termination while preserving system guidance and information gathering capabilities.

* **Refactor**
  * Optimized boot filesystem detection process for improved system performance.

### DIFF
--- a/buildroot-external/package/rpi-eeprom/0004-rpi-eeprom-update-don-t-exit-early-on-CM4-and-set-UP.patch
+++ b/buildroot-external/package/rpi-eeprom/0004-rpi-eeprom-update-don-t-exit-early-on-CM4-and-set-UP.patch
@@ -1,0 +1,50 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Jan=20=C4=8Cerm=C3=A1k?= <sairon@sairon.cz>
+Date: Tue, 26 May 2026 11:12:26 +0200
+Subject: [PATCH] rpi-eeprom-update: don't exit early on CM4 and set
+ UPDATE_BLOCKED
+MIME-Version: 1.0
+Content-Type: text/plain; charset=UTF-8
+Content-Transfer-Encoding: 8bit
+
+On CM4, the script exits early and only prints hints how to update. We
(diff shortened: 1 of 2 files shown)


agners (approved):

buildroot-external/package/rpi-eeprom/0004-rpi-eeprom-update-don-t-exit-early-on-CM4-and-set-UP.patch
```diff
@@ -0,0 +1,50 @@
...
+ 
+    if [ ${BOARD_TYPE} -eq 17 ] && [ ${BOARD_REVISION} -lt 4 ]; then
+@@ -744,7 +744,9 @@ printVersions()
+ # status inline instead of only failing once an update is attempted.
+ checkUpdateMethod()
+ {
+-   if blkid | grep -qE "/dev/mmcblk0p1.*LABEL_FATBOOT.*hassos-boot.*TYPE.*vfat"; then
++   if [ "${UPDATE_BLOCKED}" = 1 ]; then
++      return
++   elif blkid | grep -qE "/dev/mmcblk0p1.*LABEL_FATBOOT.*hassos-boot.*TYPE.*vfat"; then
```
agners, line 47: Did you consider using `/proc/device-tree/chosen/bootloader/boot-mode`? 🤔 

The current approach is pre-existing, is just something I remembered that some SoC have that boot device information. And it seems this would be a more reliable source to determine if we really booted from there.

sairon: I didn't, I've used the pre-existing code. In fact, while it's unlikely it will happen, the system doesn't need to be booted from SD, it only needs to have that partition so bootloader picks the update file.

PR #4736, opened 2026-05-27
Extend hassos-supervisor.service stop timeout for graceful shutdown

sairon:
Give Supervisor enough time to gracefully stop Home Assistant Core, apps and plugins when the host shuts down. The unit is already ordered After=docker.service, so on a host reboot or power off it is stopped before Docker tears containers down.

Supervisor should handle the SIGTERM from "docker stop" by running its managed shutdown within 420 seconds, after that point it would be killed by Docker, and ultimately 30 seconds later by Systemd.

Also, drop the delay after SIGTERM for ha-cli added in #2507. Once Systemd shutdown is invoked, the ha-cli is no longer restarted.

Refs #4642, refs #4584

## Summary by CodeRabbit

* **Chores**
  * Refined supervisor service shutdown sequence with extended timeout to ensure graceful container termination
  * Improved system CLI termination handling to respond faster during shutdown scenarios


agners (approved):
LGTM!

PR #4739, opened 2026-06-01
Drop `_workstation._tcp` service annoucement from HAOS

agners:
Since switching to systemd-resolved we've announced Home Assistant Operating System as a workstation through `_workstation._tcp` DNS-SD. This most likely came in to preserve what avahi announced by default. However, the service has not a clear definition what it means and there is no need for the service announcement really. The host name gets announced even without a service announcement. This commit removes the service announcement file from the rootfs.

## Summary by CodeRabbit

* **Chores**
  * Removed the workstation DNS-SD service advertisement configuration, disabling network discovery of workstation services.

### DIFF
--- a/buildroot-external/rootfs-overlay/usr/lib/systemd/dnssd/workstation.dnssd
+++ b/buildroot-external/rootfs-overlay/usr/lib/systemd/dnssd/workstation.dnssd
@@ -1,3 +1,0 @@
-[Service]
-Name=%H [%m]
-Type=_workstation._tcp


sairon (approved):
> This most likely came in to preserve what avahi announced by default. 

Commit 25a0dd30823e5863b4d90f2556d05149099be864 says:

> Also continue to offer the _workstation._tcp.local service since it is used by the CoreDNS mDNS plug-in.

Anyway, I tried to look up what would be a legitimate use-case that would be broken by disabling this announcement and I couldn't find any, and apparently nothing in HA ecosystem relies on it.

agners:
> > Also continue to offer the _workstation._tcp.local service since it is used by the CoreDNS mDNS plug-in.
> 
> Anyway, I tried to look up what would be a legitimate use-case that would be broken by disabling this announcement and I couldn't find any, and apparently nothing in HA ecosystem relies on it.

Right, should have mentioned that: I have no memories of why that would be true.

But Claude did find something: It seems that pre-systemd-resolved CoreDNS mDNS plug-in required services to be active, then took the host name from that, see: https://github.com/home-assistant/plugin-dns/pull/73/changes#diff-73342df3595355fa8979978d50e645864e2a6b612d0efaa5a9976fa030502577L96-L108. Anyhow, that PR removed that code, and now no service is required for mDNS resolves to work. This was in 2022, it is safe to remove this "guard service" today.

PR #4741, opened 2026-06-01
Fix OS update tests when update returns no response

sairon:
When the update command returns no response, which may happen after the shutdown reordering implemented as part of #4642, the OS update test hangs waiting for this response. To fix it, fire away the OS update command and wait either for message about missing OTA info or for early boot output.

## Summary by CodeRabbit

* **Tests**
  * Enhanced OS update test reliability through improved boot detection patterns and refined retry mechanisms for handling OTA update scenarios, ensuring more robust verification of system reboot processes during platform updates.

### DIFF
--- a/tests/smoke_test/test_os_update.py
+++ b/tests/smoke_test/test_os_update.py
@@ -55,15 +55,15 @@ def test_os_update(shell, shell_json, target):
 
     # Core (and maybe Supervisor) might be downloaded at this point, so we need to keep trying
     while True:
-        output = "\n".join(shell.run_check(f"ha os update --no-progress --version {stable_version} || true", timeout=120))
-        if "Don't have an URL for OTA updates" in output:
-            shell.run_check("ha su reload --no-progress")
-        elif "Command completed successfully" in output:
+        shell.console.sendline(f"ha os update --no-progress --version {stable_version} || true")
+        # a successful update may reboot the system immediately, so wait for the new slot to boot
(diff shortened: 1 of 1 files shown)


tests/smoke_test/test_os_update.py
```diff
@@ -55,16 +55,16 @@ def check_container_running(container_name):
...
     while True:
-        output = "\n".join(shell.run_check(f"ha os update --no-progress --version {stable_version} || true", timeout=120))
-        if "Don't have an URL for OTA updates" in output:
-            shell.run_check("ha su reload --no-progress")
-        elif "Command completed successfully" in output:
+        shell.console.sendline(f"ha os update --no-progress --version {stable_version} || true")
+        # a successful update may reboot the system immediately, so wait for the new slot to boot
+        index, *_ = shell.console.expect(["Booting `Slot ", "Don't have an URL for OTA updates"], timeout=120)
+        # matched on "Booting `Slot "
+        if index == 0:
```
agners, line 62: Wouldn't this be boot slot 1 then? 🤔

agners: Ah right, that is index of the pattern list, so we just expect any slot to boot.

I guess we could check if the other slot is booted here? :thinking:

agners: Anyways, feature creep 😅

agners (approved):

PR #4764, opened 2026-06-11
Shrink data partition for leaner disk images and faster flashing

sairon:
Borrow shrinking used in home-assistant/operating-system-full-images for reducing the data partition size to only fit its contents. Currently the data partition is intentionally overprovisioned to comfortably fit all docker images in the hassio build. This results in unnecessarily large image which takes longer to flash, as all the zeroes at the end of the filesystem need to be written to the SD card.

For OVA and aarch64 VM formats, the image is resized before creating the VM images - this also makes all generic-aarch64 images sized to 32GB, unlike 6GB which inherently needed resizing before use. Some juggling extra juggling is needed in the aarch64 post-image step, as we want to preserve the raw image (e.g. for generic aarch64 boards) but it's desirable to keep it minimal as well, as it's meant to be flashed to real hardware storage.

## Summary by CodeRabbit

* **Chores**
  * Removed hardcoded disk size configurations from board metadata files.
  * Refactored disk sizing logic to use dynamic calculations instead of fixed values.

* **Improvements**
  * Virtual machine disk images are now resized to a standard 32GB.
  * Data partition is optimized to use minimal required size, reducing build artifact sizes.


sairon:
For instance, the rpi4-64 image is shrinked down from 2G to 914M, effectively cutting flashing time in half.

agners (approved):
Nice! LGTM!

PR #4771, opened 2026-06-15
Add sanity checks for correct release channel to the build workflow

sairon:
There's currently nothing preventing us from publishing RC build to the stable channel if the GH release is configured incorrectly. Since we have the meta file which dictates where the deployment should go, use this as a sanity check and reject the workflow if the expected configuration doesn't match.

## Summary by CodeRabbit

* **Chores**
  * Enhanced the build workflow with channel-aware pre-publish validation, ensuring deployment settings and version formats match the intended release channel. The process now stops the publish with clear error messages when configuration or version patterns are incorrect, improving release safety.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -146,25 +146,30 @@ jobs:
             fi
           fi
 
-      - name: Validate deployment against publish channel
+      - name: Validate deployment and version against publish channel
         if: ${{ steps.check_publish.outputs.publish_build == 'true' }}
         env:
           CHANNEL: ${{ steps.channel.outputs.channel }}
+          VERSION_FULL: ${{ steps.version.outputs.version_full }}
         run: |
           . "${GITHUB_WORKSPACE}/buildroot-external/meta"
-          case "${DEPLOYMENT}" in
-            development) expected_channel="dev" ;;
-            staging) expected_channel="beta" ;;
-            production) expected_channel="stable" ;;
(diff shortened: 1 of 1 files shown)


agners (commented):
Nice! Yeah checking the deployment makes sense!

That said, I think while at it I'd also check the versioning. Maybe we can keep it simple and only do that on stable (check that it's a simple major.minor version).

sairon:
> That said, I think while at it I'd also check the versioning. Maybe we can keep it simple and only do that on stable (check that it's a simple major.minor version).

We already do that:

https://github.com/home-assistant/operating-system/blob/4e8633c7a90f44c0da4b0e7d0e895f888f0154e3/.github/workflows/build.yaml#L111-L118

agners:
> > That said, I think while at it I'd also check the versioning. Maybe we can keep it simple and only do that on stable (check that it's a simple major.minor version).
> 
> We already do that:
> 

Right, that checks tag against metadata, and certainly make sense!

But it doesn't prevent a stable release with rc or dev in it's name :sweat_smile:  . Checking `DEPLOYMENT` in the same file doesn't necessarily mean the version was correctly set. We could set a second variable, based on DEPLOYMENT, `version_re=`

Maybe reverse the whole logic, so it becomes "channel mandates certain meta rules" (make the case be about channel).

agners (approved):
LGTM!

PR #4776, opened 2026-06-17
Enable usbip on all targets

pvizeli:
## Summary

Enables the `usbip` userland tools on all Home Assistant OS targets so USB devices can be shared/attached over the network.

`package/usbip` is already provided and registered by the pinned Buildroot (`2025.02.x-haos`), so this PR only needs to flip `BR2_PACKAGE_USBIP=y` in the defconfigs. The USB/IP kernel modules (`CONFIG_USBIP_*`) are already enabled via `device-support.config`, so no kernel config changes are needed either.

## Notes
- Defconfig entries were placed by hand (after `BR2_PACKAGE_UDISKS2=y`); `make <board>_defconfig savedefconfig` will normalise ordering if needed.

## Summary by CodeRabbit

* **Chores**
  * Enabled USB over IP (USBIP) support by turning on the USBIP package in the provided Buildroot configurations, including generic x86-64 and ARM64 builds and multiple supported boards (such as Raspberry Pi, ODROID, and Khadas VIM3).


sairon (approved):
> Defconfig entries were placed by hand (after BR2_PACKAGE_UDISKS2=y); make <board>_defconfig savedefconfig will normalise ordering if needed.

The bot knows that but doesn't do that, bad bot! 😅 

Anyway, fixed. Looks good otherwise, bloat in the x86 image is 60kB which is totally acceptable.

Issue #4811, opened 2026-06-22
RPi 5 firmware update via `ha os boards raspberrypi firmware update` rewrites BOOT_ORDER to SD-first - Pi stops booting from NVMe


msnapka:
### Describe the issue you are experiencing

On a Pi 5 booting from NVMe, running `ha os boards raspberrypi firmware update` followed by a reboot caused the system to boot from a bootable microSD card that was present in the slot, instead of from the NVMe it had been running on previously.

There were no hardware errors and no LED flash codes - the Pi booted cleanly, just from the wrong device. After the update the active `BOOT_ORDER` was `0xf461` (SD → NVMe → USB), i.e. SD-first.

The fix was simply to remove the SD card and power on again - HAOS then booted from NVMe normally.

### Why this is a boot-order change caused by the update (not a pre-existing config)

The same bootable SD card had been sitting in the slot the entire time the system was booting from NVMe normally, for a long period before the firmware update. If `BOOT_ORDER` had been SD-first (`0xf461`) all along, the SD would have won on every boot - but it never did; the Pi consistently booted from NVMe with that card in place.

After the firmware update, with nothing else changed, the same card suddenly took priority. The only thing that changed was the bootloader firmware. Therefore the effective boot order was NVMe-first before the update and SD-first (`0xf461`) after it - the update rewrote it.

This contradicts the assumption that HAOS / the firmware update does not modify `BOOT_ORDER`.

###Expected behaviour
A system booting from NVMe should continue to boot from NVMe after a firmware update. The update must preserve the existing `BOOT_ORDER`, or at minimum warn the user that the boot order will be reset and that any bootable medium in the SD/USB slot may take priority afterwards.

### What operating system image do you use?

HAOS for RPi - installed via RPi Imager

### What version of Home Assistant Operating System is installed?

18

### Did the problem occur after upgrading the Operating System?

No

### Hardware details

- Board: Raspberry Pi 5 (rpi5-64), Argon ONE V3 case, NVMe (Samsung SSD 970 EVO 500GB)
- Home Assistant OS: 18.0
- Home Assistant Core: 2026.6.4
- Bootloader before update: 2024-06-05 → after: 2026-05-11
- `BOOT_ORDER` observed after update: `0xf461` (default) — changed to `0xf416` as workaround

### Steps to reproduce the issue

1. Pi 5 with HAOS booting from NVMe, with a bootable microSD also present in the slot, booting from NVMe consistently.
[...]

PR #4815, opened 2026-06-23
rpi-eeprom: migrate config via flashrom when blconfig is unavailable

agners:
On a Raspberry Pi 5 the live bootloader configuration is only exposed through the blconfig nvmem device; vcgencmd bootloader_config is not implemented on BCM2712 and fails with "ioctl_set_msg failed:-1".

On firmware older than pieeprom-2025-02-11 the blconfig nvmem device does not appear: the bootloader fills the reserved-memory reg using #size-cells=1, while the kernel DTB now declares #size-cells=2, so the region cannot be parsed and nvmem-rmem never binds (see raspberrypi/linux@527d7643ec01). getBootloaderConfig then returns nothing, prepareImage trips its "fewer than 3 lines" guard, and the update flashes the new image with its default config - silently resetting BOOT_ORDER to the SD-first default (#4811).

Add a flashrom fallback to getBootloaderConfig that reads the current configuration directly from the SPI EEPROM when blconfig is unavailable, so the update can migrate the existing config (including BOOT_ORDER) even when booting from old firmware.


agners:
This is not necessary when using a kernel which can properly communicate with the old firmware (https://github.com/raspberrypi/linux/commit/ec0e437419f3). Closing.

Issue #4788, opened 2026-06-20
HAOS 18.0 Update completely destroys partition table on ODROID N2+ (eMMC) - Reproducible Bootloop


FasteryXD:
### Describe the issue you are experiencing

I attempted to update my ODROID N2+ from HAOS 17.3 to the new 18.0 release via the standard UI update button. After the reboot, the system did not come back online. 

Connecting a monitor revealed that the device was stuck in a bootloop. The update process seems to completely corrupt the partition table of the eMMC module. 

The screen shows the following U-Boot errors:
- `Found U-Boot script /boot.scr`
- `Error reading cluster`
- `Failed to load '/boot.scr'`
- `** No partition table - mmc 1 **`

After this, the ODROID falls back to PXE boot, endlessly looping with `BOOTP broadcast` and `Retry time exceeded; starting again`.

### What operating system image do you use?

odroid-n2 (Hardkernel ODROID-N2/N2+)

### What version of Home Assistant Operating System is installed?

Updating from 17.3 to 18.0

### Did the problem occur after upgrading the Operating System?

Yes

### Hardware details

- Board: ODROID N2+
- Storage: eMMC module (mounted directly on the board)
- No additional specific hardware attached that interferes with the boot process.

### Steps to reproduce the issue

1. Flash HAOS 17.3 fresh onto an eMMC module using BalenaEtcher/Raspberry Pi Imager.
2. Insert eMMC into ODROID N2+, boot, and complete the initial setup (or restore a backup).
3. Trigger the OS update to version 18.0 via the Home Assistant UI.
4. Wait for the automatic reboot.
5. The system never comes back online. 
6. Connect a monitor via HDMI to see the U-Boot crash (`** No partition table - mmc 1 **`) followed by an endless PXE boot loop.

### Anything in the Supervisor logs that might be useful for us?

```txt
Not accessible. The system does not boot the OS or Supervisor at all after the update. It crashes at the U-Boot stage because the partition table is destroyed.
```

### Anything in the Host logs that might be useful for us?

```txt
Not accessible for the same reason. Please see the monitor output photos attached in the "Additional information" section.
```

### Additional information

I have to say I am more than disappointed. A stable release update should under no circumstances brick the entire device by destroying the partition table. For users who rely on Home Assistant for critical network infrastructure, a sudden and complete system death caused by a simple UI update is an absolute worst-case scenario. Please look into this urgently.

[...]

PR #4823, opened 2026-06-26
ODROID-N2: Fall back to other slot when a kernel fails to load

agners:
The U-Boot boot script only persisted the decremented boot-attempt counter (storebootstate) when a kernel was successfully loaded. When *neither* slot's kernel could be loaded, it instead reset both counters back to 3 and rebooted - pinning the counters and looping forever without ever falling back or giving up. A failed kernel load therefore never counted as a boot attempt, defeating the A/B rollback safety net (#4788).

Persist the (possibly decremented) counters in all cases, so a failed load consumes an attempt just like a hang/crash after booti. Repeated load failures now count down and we fall back to the other slot. Only re-arm the counters as a true last resort, when no slot had any attempts left (tracked via boot_attempted) or BOOT_ORDER was invalid, instead of on every transient load failure.

Note: this does not address the underlying eMMC read unreliability seen on some N2 units (kernel/boot.scr loads failing in U-Boot); it makes the slot-fallback logic behave correctly when a load does fail.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved boot slot handling so the device only boots from a slot that was actually attempted and selected.
  * Added clearer behavior when no bootable slot is found, avoiding unnecessary boot attempts.
  * Boot attempt counters are now saved consistently after slot checks, helping preserve boot state across restarts.
  * If no slot is available, the system can restore the default boot order and retry more reliably.

### DIFF
--- a/buildroot-external/board/hardkernel/odroid-n2/uboot-boot.ush
+++ b/buildroot-external/board/hardkernel/odroid-n2/uboot-boot.ush
@@ -72,11 +72,13 @@ fi
 # 8. SystemB
 # 9. BootInfo
 setenv bootargs
+setenv boot_attempted
 for BOOT_SLOT in "${BOOT_ORDER}"; do
   if test "x${bootargs}" != "x"; then
-    # skip remaining slots
+    # A bootable slot has already been selected, skip the remaining ones.
   elif test "x${BOOT_SLOT}" = "xA"; then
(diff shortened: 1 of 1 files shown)


agners:
Ultimately this should be applied to all boards. But since we experience these boot issues on ODROID-N2(+) currently, I'd prefer to only touch this board for now.

agners:
Closing in favor of #4832.

PR #4831, opened 2026-06-29
Add usage to busybox utils

### LINKED ISSUE #4830: Add usage information to busybox utils so we can debug OS-level problems easier

Hi, thanks for your work on Home Assistant, I'm really enjoying it.

I was recently debugging some problems with my network by SSH'ing into the home assistant core OS. I couldn't just SSH into the Terminal docker container as I was trying to debug Thread/Matter Apple TV bridge network problems and I wanted to rule out docker's virtual networking and access any firewall config in the host kernel.

I had a tough time debugging because many busybox commands just exitted silently with status `1` when they have an unknown option.

e.g. ping with count 3 just exits with nothing:

```
# ping -c 3 192.168.1.75
# echo  $?
1
```

And `--help` outputs nothing, so it's very hard to figure out what arguments are supported:

```
# ping --help
```

This silence is because we set this config:

https://github.com/home-assistant/operating-system/blob/b080fea94752747bfb048ab8f528126e565489fc/buildroot-external/busybox.config#L16

I looked into the history of this, it goes right back to https://github.com/home-assistant/operating-system/blame/b080fea94752747bfb048ab8f528126e565489fc/buildroot-external/busybox.config#L16 "Cleanup not needed stuff" in 2018, where we tried to reduce the binary size.

I think we would benefit from enabling this config setting. Humans would benefit, but it'd also be invaluable for AI agents trying to debug problems too.

It would increase the home assistant OS size very slightly - probably a few KB, though I haven't measured it. I think it'd be worth it; whenever we have to debug in the host, we're already in big debugging trouble :-), and hiding the command line argument options makes it even tougher.

I'm happ to talk about optimising busybox size, but it's currently at half a meg, seems a little... inconsequential, when we also include Docker, which is over 100MB.

```
# ls -lh libbusybox.so.1.37.0
-rw-r--r--    1 root     root      453.8K Jun 19 03:18 libbusybox.so.1.37.0
```

```
# ls -lh /usr/bin/docker /usr/bin/dockerd
-rwxr-xr-x    1 root     root       81.4M Jun 19 03:18 /usr/bin/dockerd
-rwxr-xr-x    1 root     root       27.3M Jun 19 03:18 /usr/bin/docker
```

Thanks for your consideration. I hope one day this can also make your debugging easier!

Home Assistant OS 18.0

No

### Hardware details

Generic x86_64 (Dell OptiPlex)

Setup SSH by putting authorized_keys in the USB drive and copying them over. SSH into the bare host with `ssh root@192.168.1.75 -p 22222`.

```txt
No
```

```txt
No
```

### System information

_No response_

### Additional information

_No response_

mhansen:
This costs a few KB, but is invaluable when debugging host problems, otherwise busybox just exits failure on any argument it doesn't understand.

We could also enable `CONFIG_FEATURE_VERBOSE_USAGE` and `CONFIG_FEATURE_COMPRESS_USAGE` potentially, but I wanted to start the discussion with just this feature first.

I measured this and it's +4KB to libbusybox.so.

Fixes #4830

## Summary by CodeRabbit

- **New Features**
  - BusyBox now shows command usage information when help is requested, making it easier to discover available options and correct command syntax.


agners (approved):
Yeah enabling usage makes sense and given the small bloat this appears uncontroversial to me. LGTM!

agners:
> We could also enable `CONFIG_FEATURE_VERBOSE_USAGE` and `CONFIG_FEATURE_COMPRESS_USAGE` potentially, but I wanted to start the discussion with just this feature first.

Yeah let's make this another PR along with the numbers what it would add. Are there any tradeoffs compressing usage text? 🤔

mhansen:
Thank you!

PR #4853, opened 2026-07-02
Enable CONFIG_FEATURE_VERBOSE_USAGE in busybox

mhansen:
Following on from enabling CONFIG_SHOW_USAGE=y.

Rationale: make it easier to debug the host OS.

Size increase of libbusybox.so: +32 KB over CONFIG_SHOW_USAGE=y baseline.

We could enable COMPRESS_USAGE to bring it down to +8KB over CONFIG_SHOW_USAGE=y baseline, but I'm not sure it's worth optimizing? If we want to compress, it's probably fine: just a slight CPU hit when asking for usage, which is fine for interactive use.

This is a follow up to #4830, we discussed this in https://github.com/home-assistant/operating-system/pull/4831


agners (approved):
32KB seems quite reasonable to me. LGTM, thanks!

mhansen:
Thanks. Did you mean to merge or just leaving open for others to chime in?
I'm happy either way

agners:
> Thanks. Did you mean to merge or just leaving open for others to chime in?

Yeah I left it open in case someone wants to chime in. Chime-in periode elapsed, so let's merge it! :smile:

PR #4854, opened 2026-07-02
Enable CONFIG_FEATURE_FANCY_PING in busybox

mhansen:
Ping is a foundational tool for network debugging, and with so many network integrations in Home Assistant, we deserve a fancy ping.

This enables flags like `-c` count ping and packet loss statistics. Nothing totally groundbreaking, but it enables more detailed debugging.

The prompt for wanting this was when I tried debugging a Matter issue with an AI agent, and it tried invoking `ping -c 5 $hostname` to get more data per shell invocation.

Without fancy ping, you get very limited flags, that just do one ping per invocation.

```
# ping 127.0.0.1
127.0.0.1 is alive!
# ping --help
BusyBox v1.37.0 (2026-07-01 07:04:23 UTC) multi-call binary.

Usage: ping HOST
```

This adds +4KB to libbusybox.so


agners (approved):
What I also like that it shows the resolved IP. This was a bit annoying with the rudimentary `ping` command,

LGTM, thanks!

PR #4855, opened 2026-07-02
Enable CONFIG_WHICH in busybox

mhansen:
This lets us run `which`

I've found myself wanting this a bunch while trying to figure out, well, which binary I'm running. I can walk $PATH myself, but it's no fun.

libbusybox.so size impact: +0 bytes (that is, it fits in padding in the 4KB-padded binary sections we already have).


agners (approved):
LGTM, thanks!

PR #4912, opened 2026-07-20
Enable Apple SMC hardware monitoring support

KingKaom:
Tested on:

Macmini4,1
Intel x86_64
Linux 6.18.37-haos

current HAOS kernel reports:

CONFIG_SENSORS_APPLESMC is not set

while a Linux Mint kernel on the same machine exposes Apple SMC and allows control of fan speed on Intel macs

## Summary by CodeRabbit

* **New Features**
  * Added support for Apple System Management Controller (SMC) hardware monitoring.
  * Enabled the Apple SMC sensors driver as a module for compatible devices.


KingKaom:
I would like to clarify that this is not about adding Mac support per se. The goal is just to enable an existing upstream hardware monitoring interface for any x86_64 systems that expose Apple SMC which happens to help older Mac minis with broken or missing thermal sensors (common issue on older intel mac minis). On unsupported systems the module remains unused.

buildroot-external/kernel/v6.18.y/device-support.config
```diff
@@ -60,6 +60,9 @@ CONFIG_BT_MRVL_SDIO=m
 CONFIG_BT_MTKSDIO=m
 CONFIG_BT_MTKUART=m
 
+# Hardware monitoring
```
agners: Doesn't seem a very helpful comment. The commit description is all which is needed ultimately.

```suggestion
```

agners (changes requested):

agners (approved):

PR #4918, opened 2026-07-27
Do not start Supervisor from HA CLI and serial getty units

agners:
The HA CLI unit (and the serial getty drop-in) pulled in `haos-supervisor.service` via `Wants=`. Since #4326 the CLI unit is
configured to never reach its start limit (`RestartSec=100ms`, `RestartMaxDelaySec=3s`, `StartLimitIntervalSec=3s`), so it restarts indefinitely - and every restart starts the Supervisor again.

This makes it impossible to stop the Supervisor for debugging: stopping `haos-supervisor.service` and removing the containers terminates `hassio_cli`, which makes `ha-cli@tty1.service` exit and restart, which immediately starts the Supervisor back up:

```
10:00:48 systemd[1]: ha-cli@tty1.service: Scheduled restart job, restart counter is at 14.
10:00:48 systemd[1]: Starting HAOS supervisor...
```

The `Wants=` is not needed: `haos-supervisor.service` is `WantedBy=multi-user.target` and gets started on boot regardless. Drop it and keep only the `After=` ordering.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved service startup behavior by preserving startup order without automatically starting the Home Assistant Supervisor for CLI and serial console services.
  * Reduced unnecessary service activation during system initialization.

### DIFF
--- a/buildroot-external/rootfs-overlay/usr/lib/systemd/system/ha-cli@.service
+++ b/buildroot-external/rootfs-overlay/usr/lib/systemd/system/ha-cli@.service
@@ -1,6 +1,5 @@
 [Unit]
 Description=Home Assistant CLI
-Wants=haos-supervisor.service
 After=systemd-user-sessions.service plymouth-quit-wait.service getty-pre.target haos-supervisor.service
 Conflicts=getty@%i.service
 
--- a/buildroot-external/rootfs-overlay/usr/lib/systemd/system/serial-getty@.service.d/haos.conf
+++ b/buildroot-external/rootfs-overlay/usr/lib/systemd/system/serial-getty@.service.d/haos.conf
@@ -1,3 +1,2 @@
 [Unit]
-Wants=haos-supervisor.service
 After=haos-supervisor.service


agners:
**Safety verification of this change** (review done with Claude Code):

- **Boot startup is unaffected.** `haos-supervisor.service` is `WantedBy=multi-user.target` and gets enabled at image build time via Buildroot's `systemctl preset-all` (the HAOS `70-haos.preset` doesn't override it). Verified the `multi-user.target.wants/haos-supervisor.service` symlink is present in `target/etc/systemd/system/` of local builds across targets (generic-x86-64, rpi4/5, green, ova, odroid).
- **No other unit relied on these `Wants=` lines.** The only remaining references to `haos-supervisor.service` are pure ordering (`After=` in the touched units, `Before=` in `raucdb-update.service`), which is inert when the unit isn't part of the transaction.
- **Crash recovery is unchanged.** The Supervisor unit self-heals via its own `Restart=always` / `RestartSec=5s`.
- **One subtle behavioral change**, arguably a second fix: previously, if the Supervisor exhausted its start limit (`StartLimitBurst=5`/30 min → `failed`), a console interaction that restarted `ha-cli@`/getty could resurrect it via `Wants=` once the window expired. That accidental retry path is gone — a start-limit-failed Supervisor now stays down until manual start or reboot. This was a weak safety net anyway (`haos-cli` parks in an emergency shell when the Supervisor is down, so the retry only fired on human console interaction), and five consecutive start-script failures need intervention regardless. Flip side: `systemctl stop`/`disable` of the Supervisor now actually sticks, which is the point of this PR.

PR #4923, opened 2026-07-28
Keep Dropbear alive during Supervisor and Docker shutdown

agners:
## Description

`dropbear.service` has no ordering relationship to `haos-supervisor.service` or `docker.service`, so systemd schedules all three stop jobs in the same batch and runs them concurrently. Dropbear exits within milliseconds while the Supervisor gets up to 450s to stop containers (`ExecStop=docker stop -t 420`), so an SSH session is dropped right when the interesting part of shutdown starts. That makes it needlessly hard to debug slow shutdowns.

This orders Dropbear before both units. Since stop order is the reverse of start order, systemd now stops it only once the Supervisor and Docker are gone.

Notes on safety:

- Ordering-only dependency. If Dropbear fails, is masked, or is skipped by its `ConditionFileNotEmpty=/root/.ssh/authorized_keys`, the ordering counts as satisfied and both units start as before.
- No added boot latency: `docker.service` already waits for the overlay partition that Dropbear needs for `/etc/dropbear`.
- No ordering cycle: `/mnt/overlay` is a separate partition with `DefaultDependencies=no`, and nothing below it depends on Docker. Verified with `systemd-analyze verify` against a built rootfs.

This only covers the unit-stop phase. Unmounting `/mnt/data` and the final `systemd-shutdown` phase still happen after Dropbear is gone, since its own mount dependencies have to be released first. Covering those would need `DefaultDependencies=no` + `IgnoreOnIsolate=yes` (debug-shell style), which is both riskier and a poor instrument — the live session's cwd on `/mnt/data` would itself become an unmount blocker. That phase is better observed on the console with `systemd.log_level=debug`.

Verified on hardware — see the comment below for the timings.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved shutdown sequencing to keep SSH/remote access available during supervisor and container teardown.
  * Updated service ordering so the remote access service stops after the supervisor/Docker services begin stopping, reducing interruptions during shutdown.

### DIFF
--- a/buildroot-external/rootfs-overlay/usr/lib/systemd/system/dropbear.service.d/haos.conf
+++ b/buildroot-external/rootfs-overlay/usr/lib/systemd/system/dropbear.service.d/haos.conf
@@ -1,6 +1,11 @@
 [Unit]
 RequiresMountsFor=/etc/dropbear
 ConditionFileNotEmpty=/root/.ssh/authorized_keys
+# Systemd stops units in reverse start order, so ordering dropbear before the
+# Supervisor and Docker keeps SSH available while their (potentially minutes
+# long) teardown runs. Both already require the same mounts dropbear needs, so
+# this adds no extra work to boot.
+Before=haos-supervisor.service docker.service
 
 [Service]
 Type=forking


agners:
## Verification on hardware

Tested on a generic-x86-64 VM with an SSH session on port 22222 running `journalctl -f` across a `reboot`. The session survived the entire Supervisor teardown, where previously it was dropped at the very first `Stopping HAOS supervisor...` line.

### Stop ordering

From the persisted journal of the previous boot (`journalctl -b -1 -u dropbear -u docker -u haos-supervisor -o short-precise`):

```
15:14:41.972  Stopped HAOS supervisor.
15:14:41.974  Stopping Docker Application Container Engine...
15:14:42.068  Stopped Docker Application Container Engine.
15:14:42.071  Stopping Dropbear SSH daemon...      <-- last
15:14:42.080  Stopped Dropbear SSH daemon
```

Both ordering edges hold: Dropbear is stopped last, 3ms after `docker.service` is gone.

### Timings

| Phase | Start | Container gone | Duration |
|---|---|---|---|
| reboot initiated | 15:14:16.0 | | |
| homeassistant (Core) | 15:14:16.35 | 15:14:23.17 | 6.8s |
| app_core_matter_server | 15:14:23.22 | 15:14:26.29 | 3.07s |
| hassio_cli | 15:14:26.33 | 15:14:29.36 | 3.02s |
| hassio_dns | 15:14:29.43 | 15:14:32.47 | 3.04s |
| hassio_audio | 15:14:32.54 | 15:14:35.58 | 3.04s |
| hassio_multicast | 15:14:35.66 | 15:14:38.70 | 3.04s |
| `Supervisor is down` | | 15:14:38.74 | **22.4s** |
| `Stopped HAOS supervisor` | | 15:14:41.97 | |
| `Stopped Docker` | | 15:14:42.07 | 94ms |
| `Stopped Dropbear` | | 15:14:42.08 | 9ms |

All 22.4s of the Supervisor teardown is now observable over SSH.

### An earlier revision of this PR also set `TimeoutStopSec=10s` — dropped, and why

The reasoning was: Dropbear is built without PAM, so SSH sessions live in the service's own cgroup, and `KillMode=mixed` only sends SIGTERM to the listener. An open session would therefore keep the cgroup populated until the stop timeout expires — and with Dropbear stopped last, that 90s default would sit on the critical path of shutdown.

The journal shows that is not how systemd behaves:

```
15:14:42.072353  dropbear[472]: Early exit: Terminated by signal
15:14:42.072563  dropbear.service: Killing process 2432 (dropbear) with signal SIGKILL.
15:14:42.072608  dropbear.service: Killing process 2433 (sh) with signal SIGKILL.
15:14:42.072648  dropbear.service: Killing process 2697 (journalctl) with signal SIGKILL.
```

The live session (`sh` plus the `journalctl -f`) was SIGKILLed 200µs after the listener exited, and the whole stop took 9ms. `TimeoutStopSec=` bounds how long systemd waits for the **main** process to exit; once that is gone, systemd proceeds straight to the final SIGKILL of whatever remains in the cgroup. Since Dropbear exits promptly on SIGTERM, the tail cannot occur, so the setting was dead config and has been removed. This PR is now a single `Before=` line.

### Unrelated lead: ~3s of idle time per plugin container

Worth recording since this change is what made it visible. Every plugin takes almost exactly 3.0s to stop, and it is not spent doing work. For `hassio_dns`, s6 reports `legacy-services`, `legacy-cont-init` and `fix-attrs` all `successfully stopped` within ~30ms at 15:14:29.4, and then nothing happens until the scope deactivates at 15:14:32.47. The same shape repeats on every container.

Five plugins idling ~3s each is roughly 15s of the 22.4s, and it is a fixed delay rather than load-dependent. That points at one of s6-overlay's 3000ms defaults (`S6_SERVICES_GRACETIME` / `S6_KILL_GRACETIME`) — container PID 1 waiting out a timer after everything beneath it has already stopped. Which of the two it is would need confirming. Not actionable in this repo (it belongs to the Supervisor plugin images), but it is the concrete lead behind a good chunk of HAOS shutdown time.

Issue #2876, opened 2023-10-29
HA wont start after Update to 11.1 on raspberry pi 3b with USB3 SSD


magin1:
### Describe the issue you are experiencing

After updating from 11.0 (i think) to 11.1  and now my HA is not booting anymore.
It hangs in u-boot unable to find storage device to boot from.
And the network does not seem to be seen, while cable is connected
After a while it shows some lines of which i dont know what they mean.
![image](https://github.com/home-assistant/operating-system/assets/41379338/f217bed3-9977-4f70-974d-028ba243127d)

I tried redoing to config needed to boot from ssd. no luck
I installed HA 11 on a usb drive and booted from that. this worked. Network also worked

So i thought it must be my SSD. so i installed virtualbox and Linux mint to check the SSD partitions.
And they are all there and i can see my config.
![image](https://github.com/home-assistant/operating-system/assets/41379338/9ba70f30-7136-424c-89f8-04491aad1ca0)

So now i'm out of ideas what i could be.
Would love some help.

### What operating system image do you use?

rpi3 (Raspberry Pi 3 32-bit OS)

### What version of Home Assistant Operating System is installed?

11.1

### Did you upgrade the Operating System.

Yes

### Steps to reproduce the issue

Not relevant

### Anything in the Supervisor logs that might be useful for us?

```txt
No
```

### Anything in the Host logs that might be useful for us?

```txt
No
```

agners:
There was no change between HAOS 11.0 and 11.1. I can really say for sure what the issue is, but it seems the bootloader can't talk to the disk. I'd try to power off the whole system, maybe use a different USB hub or powered hub in between the Raspberry Pi and the USB SSD.

magin1:
> There was no change between HAOS 11.0 and 11.1. I can really say for sure what the issue is, but it seems the bootloader can't talk to the disk. I'd try to power off the whole system, maybe use a different USB hub or powered hub in between the Raspberry Pi and the USB SSD.

Hi thanks for you reply. toady i tested the following:
- left the raspberry off power whole night, in the morning still no differance.
- had another raspberry pi3b so i also made that albe to boot from ssd but sadly same result. Led from ssd flahses so it it talked to by the u-boot but no boot.
- also had some more powerfull chargers then the default raspberry one, connected those and same results:(

If its the bootloader not beeing able to talk to the disk for any reason, how can i check/repair this?

agners:
[...]

Issue #1712, opened 2022-01-12
U-Boot fails to boot from USB SSD sometimes on RPi 3b+


mmalina:
### Describe the issue you are experiencing

Today I updated my HAOS from 7.0 to 7.1. After reboot, I am stuck on "No ethernet found". This is displayed several times and eventually I am shown a HAOS> prompt. My RPi 3b+ is connected to ethernet and I never had connectivity issues until this update today. Reboot doesn't help.

### What operating system image do you use?

rpi3 (Raspberry Pi 3 32-bit OS)

### What version of Home Assistant Operating System is installed?

7.1

### Did you upgrade the Operating System.

Yes

### Steps to reproduce the issue

1. Update from HAOS 7.0 to 7.1
2. Wait for HAOS to reboot
3. The OS fails to connect to ethernet

### Anything in the Supervisor logs that might be useful for us?

```txt
I have no connectivity to my RPi because of this, so I can't check. (I'm not sure if HA Core even started at all.)
```

### Anything in the Host logs that might be useful for us?

```txt
I have no connectivity to my RPi because of this, so I can't check. (I'm not sure if HA Core even started at all.)
```

mmalina:
Here's a photo of my screen after updating to 7.1:
![IMG_5746](https://user-images.githubusercontent.com/2291011/149172407-a44da825-7e35-4888-a094-85ae7aaafe6e.jpeg)

mmalina:
```
Net:
No ethernet found.
starting USB.
I Bus usbe7e980000: usb dr mode not found
1 USB DUCZ
; scanning bus usbe7e980000 for devices... 4 USB Devices) found
scanning usb for storage devices... O Storage Device(s) found
WARNING at drivers/mmc/bcm2835_sdhost.c:414/bcm2835_send_command()!
WARNING at drivers/nmc/bcm2835_sdhost.c:414/bcm2835_send_command()!
Card did not respond to voltage select! : -110
)Card did not respond to voltage select! :
-110
Device O: unknown device
I No ethernet found.
I missing environment variable: pxeuuid
I missing environment variable: bootfile
I Retrieving file: pxelinux.cfg/00000000
I No ethernet found.
I missing environment variable: bootfile
I Retrieving file: pxelinux.cfg/0000000
1 Mo ethernet found.
I missing environment variable: bootfile
Retrieving file: pxelinux.cfg/000000
I No ethernet found.
I missing environment variable: bootfile
[...]

PR #4940, opened 2026-08-03
Increase USB hub debounce timeout to fix USB SSD boot on RPi 3

dlxmax:
On Raspberry Pi 3, U-Boot runs `usb start` early in boot to find the boot
device. Many USB-SATA/NVMe bridges haven't finished enumerating within the
default USB hub debounce window, so U-Boot reports "0 storage devices found"
and fails to boot from a USB SSD. This happens intermittently or on every boot
depending on the adapter.

`CONFIG_USB_HUB_DEBOUNCE_TIMEOUT` is the knob for exactly this: its own Kconfig
help notes that "some usb device needs around 1.5s to be initialized and a 2s
value should solve detection issue on problematic USB keys." Raising it gives
slow bridges time to present the drive. This is preferred over overriding
`CONFIG_PREBOOT`, which would shadow the default preboot string and could
silently mask a future upstream change to it.

Scoped to `rpi3-64` via a per-board fragment, matching the existing
`rpi4-64/uboot.config`, so Pi 4/5 images are unaffected. On my Pi 3B with a
USB-SATA bridge and SSD, stock fails with "0 storage devices"; 2000 ms was
still intermittent (roughly two boots in three), while 2500 ms boots reliably
across repeated cold starts.

The manual `setenv usb_pgood_delay ...; usb reset` workaround from #1712 doesn't
persist, because `CONFIG_ENV_IS_NOWHERE=y` means `saveenv` is unavailable, and a
hand-patched `u-boot.bin` is overwritten by OS updates. So this needs to ship in
the image.

Addresses the USB SSD boot detection reported in #1712 and #2876.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved USB hub initialization reliability on Raspberry Pi 3 64-bit systems by allowing additional debounce time during startup.
  * Applied the updated USB configuration to the Raspberry Pi 3 64-bit build.


sairon (changes requested):
Thanks for picking this up. There are couple of issues with this solution though. First of all, it delays boot of all Pi 3's, no matter if they use USB as the main boot device or not (which is another issue - direct USB boot was always rather discouraged, using SD card for OS and USB drive as data disk is the better way - anyway, it was never completely _unsupported_). But on the other hand, Pi 3 is already rather underspec'd for today's Home Assistant deployments, and that one extra second will be lost in the noise, so... I think I can live with that.

The bigger issue for me is from the maintainability perspective, as this change overrides the default `CONFIG_PREBOOT` defined either upstream or potentially in upper configs. This way it could shadow another default should it ever change. Currently it's a default string from the Kconfig, which is most likely to be changed without us noticing when updating U-Boot.

On the bright side, I think there's more appropriate way to delay the initialization instead, and that's by increasing `CONFIG_USB_HUB_DEBOUNCE_TIMEOUT`. From its [Kconfig description](https://github.com/u-boot/u-boot/blob/v2026.04/drivers/usb/Kconfig#L121-L131) it sounds it's exactly what you need and adjusting a config option is IMHO better than extending preboot environment. Would you be able to build and test if increasing that value (e.g. to 2000 from the default 1000) helps in your case?

dlxmax:
Thanks, that's a cleaner fit and I agree on the maintainability concern. Overriding `CONFIG_PREBOOT` to shadow the default string bothered me for the same reason, and the debounce timeout is purpose-built for this per its own Kconfig help.

I rebuilt with the stock preboot untouched and only `CONFIG_USB_HUB_DEBOUNCE_TIMEOUT` raised, then tested on my Pi 3B with the USB-SATA bridge and SSD:

- 2000 ms: still intermittent, roughly two cold boots in three.
- 2500 ms: reliable across repeated cold starts.

Pushed as `CONFIG_USB_HUB_DEBOUNCE_TIMEOUT=2500`, scoped to `rpi3-64`. Title and description updated to match.

sairon (approved):
Thanks!

Issue #4898, opened 2026-07-15
[HA Yellow] BLE fails regularly after upgrade from v17. 3 to 18.0


Kelniimir:
### Describe the issue you are experiencing

I'm using HA Yellow rev. 1.3 PoE version. 
Its in use for 2-2.5 years now.
It's connected over PoE and caused no issues with ZHA or BLE reachability.
I did HA OS and other components upgrades regularly, usually in a span of 2-5 days after release (for reason of monitoring chats for possible defects). 

My BLE devices setup included 4 Xiaomi LYWSD02MMC (Temperature and Humidity Clock monitor).
Shortly before upgrading to 18.0, I added 5th Xiaomi Clock, and a Switchbot for switching FAN in bathroom On/Off with automation through presence sensors.
As of now I effectively have 5 passive and 1 active BLE device. 

After upgrade to 18.0 I have noticed that frequently ALL of BLE devices are not connected for periods of time.
Battery values became unknown, or even if they are known, it's obvious connection is not reliable to due blank spaces in battery report graphs.
Also active BLE device (Switchbot) becomes unresponsive both through manual control and automation. 
Testing devices via their own official applications confirms they are reachable and controllable. 
ZHA devices unaffected, WiFi devices unaffected, non-BLE battery reports unaffected. 

Example of battery report (last time reporting enabled after full HA Yellow reboot):

<img width="1080" height="2316" alt="Image" src="https://github.com/user-attachments/assets/5004c2c0-2c57-437a-8b1b-085ea204a730" />

Sometimes it is becoming available again without any intervention, sometimes I need full reboot of HA to make it operational again. 
Only physical connectivity for HA Yellow includes Ethernet+PoE and ZBT-2 as of now. 
Issue affects all BLE devices simultaneously, regardless if they are on other end of apartments, or in same room as HA Yellow. 
Prior upgrade, I had NO BLE connectivity issues and NO blanks in any value graphs even for furthest BLE devices, and am currently considering ESP32 BLE Proxy as a measure to mitigate the problem, until it is resolved. 

Since the issue started, I have upgraded to latest available HA OS 18.1 and issue still persist. 

### What operating system image do you use?

yellow (Home Assistant Yellow)

### What version of Home Assistant Operating System is installed?

18.1

### Did the problem occur after upgrading the Operating System?

Yes

## System Information

version | core-2026.7.2
-- | --
installation_type | Home Assistant OS
dev | false
hassio | true
docker | true
container_arch | aarch64
[...]

Issue #4884, opened 2026-07-09
Yellow CM4: dtoverlay=disable-bt does not remove hci0 but prevents BCM4345C0 timeout loop


dweston:
### Describe the issue you are experiencing

On a Home Assistant Yellow PoE with Raspberry Pi CM4 8GB, HAOS 17.3, the onboard Broadcom Bluetooth adapter appears to enter a wedged initialisation/reset loop on boot with the stock `/mnt/boot/config.txt`.

The issue causes repeated kernel messages on the USB-C serial console, rendering serial access noisy and ultimately unresponsive to use.

The repeated messages include:

`Bluetooth: hci0: command tx timeout`
`Bluetooth: hci0: BCM: failed to write update baudrate (-110)`
`Bluetooth: hci0: Failed to set baudrate`
`Bluetooth: hci0: command 0x1001 tx timeout`
`Bluetooth: hci0: BCM: Reading local version info failed (-110)`
`Bluetooth: hci0: BCM: Reset failed (-110)`

A full shutdown and PoE power removal restores serial console responsiveness, but the Bluetooth hci0 timeout/reset loop recurs and serial console responsiveness degradation returns.

A runtime `rfkill/sysfs` soft block was tested but didn't stop the kernel timeout problem. Adding `dtoverlay=disable-bt` under the `[cm4]` section of `/mnt/boot/config.txt` and then doing a host reboot stops the Broadcom timeout/reset loop but the `hci0` still appears.

This process is repeatable. **[See update below contradicting this.]** Removing the added `dtoverlay=disable-bt` from the `[cm4]` section of `/mnt/boot/config.txt` causes the Bluetooth hci0 timeout/reset loop problem to recur. Adding the `dtoverlay=disable-bt` line back in again stops the Bluetooth hci0 timeout/reset loop problem but does not remove `hci0` from `/sys/class/bluetooth`.

The presence of `dtoverlay=disable-bt` in the `config` file appears to alter the Bluetooth initialisation path enough to stop the timeout/reset loop.

With the overlay present, the boot log includes:

`Bluetooth: hci0: BCM43455 37.4MHz Raspberry Pi 3+-0190`
`Bluetooth: hci0: BCM4345C0 (003.001.025) build 0382`
`Bluetooth: MGMT ver 1.23`

and the repeated timeout/reset messages stop.

**Update:** the earlier apparent mitigation using `dtoverlay=disable-bt` is not repeatable.

[...]

Issue #4590, opened 2026-03-19
Bluetooth not loading on HA Yellow


mrknns01:
### Describe the issue you are experiencing

Hardware: HA Yellow
HAOS version: 17.1 (slot B)
/lib/firmware/brcm/ directory missing entirely
BCM43438 Bluetooth non-functional since 17.1 update on 18 March 2026
Slot A (17.0) still shows as good

### What operating system image do you use?

yellow (Home Assistant Yellow)

### What version of Home Assistant Operating System is installed?

17.1

### Did the problem occur after upgrading the Operating System?

Yes

### Hardware details

Tried reboot system etc

### Steps to reproduce the issue

1.
2.
3.
...

### Anything in the Supervisor logs that might be useful for us?

```txt
Log details (WARNING)
Logger: bluetooth_auto_recovery.recover
Source: components/bluetooth/__init__.py:385
First occurred: 18 March 2026 at 21:44:33 (1416 occurrences)
Last logged: 11:12:21

Could not determine the power state of the Bluetooth adapter hci0 [00:00:00:00:00:00] (0) due to timeout after 5 seconds
Could not cycle the Bluetooth adapter hci0 [00:00:00:00:00:00] (0): [Errno 110] Operation timed out
Could not reset the power state of the Bluetooth adapter hci0 [00:00:00:00:00:00] (0) due to timeout after 5 seconds

Raspberry Pi Trading Ltd bcm43438-bt (D8:3A:DD:39:F9:E3)
Failed setup, will retry: Bluetooth adapter None with address D8:3A:DD:39:F9:E3 not found
```

### Anything in the Host logs that might be useful for us?

```txt
See above
```

sairon:
I can not reproduce the exact same behavior, and the `/lib/firmware/brcm/` firmware missing is probably an incorrect conclusion (I guess you're checking it from SSH/terminal app, not on the Yellow's console). 

I have reproduced some flakiness of the devices Bluetooth, does it not come back on every reboot or is it intermittent in your case? Can you share the full Host logs of the current boot (make sure the scrollback captures the start of the boot before you download them).

gargross:
System: HA Yellow with CM5 (upgraded from CM4), HAOS 17.3, HA Core 2026.4.4
Symptoms:

hciconfig command not found in SSH terminal
bluetoothctl show returns "No default controller available"
HA Bluetooth integration shows "No unconfigured Bluetooth adapters found"
[...]

PR #4946, opened 2026-08-04
yellow: Add opt-in overlay to route CM4 Bluetooth via PL011

gvlekke:
## Summary

Adds an **opt-in** device tree overlay for the Yellow (CM4) that routes the Bluetooth HCI UART through the PL011 (`uart0`) instead of the mini-UART, as a robust workaround for the unrecoverable Bluetooth failures tracked in #4898 (also #4590, #4884).

This is explicitly **not** a replacement for the mini-UART flow-control fix @sairon is working on — it's the "change pin muxing to use the full UART" option he mentioned in https://github.com/home-assistant/operating-system/issues/4898, packaged so bare-board users can enable it with one persistent `config.txt` line instead of hand-patching the DTB after every OS update.

## Background

Per @sairon's analysis in #4898: the mini-UART has a small FIFO and no flow control, so under load its RX FIFO overruns during HCI traffic. There is no recovery path — once the BCM43455 has been switched up from 115200, re-initialization keeps failing (`command 0xfc18 tx timeout`, `BCM: Reset failed (-110)`) until a cold power cycle.

The BT chip's UART lines (GPIO 30–33) can be muxed to either UART. The overlay muxes them to the PL011 in ALT3 — fixed clock, CTS/RTS hardware flow control, 3 Mbps — and disables the mini-UART. FIFO overruns become physically impossible.

**Trade-off (why opt-in):** the PL011 normally serves the J11 pin header, so the overlay disables the J11 UART. Boards with UART extensions (RaZberry, HM-MOD-RPI-PCB) must not use it. CM4 only; CM5 BT is on the BCM2712's own UART and unaffected.

## Implementation

- New `yellow-overlays` package (follows the `rpi-rf-mod` precedent): compiles `ha-yellow-bt-pl011.dtbo` with `dtc -@`, installs to `BINARIES_DIR`; `haos-hook.sh` already copies any `.dtbo` into the boot partition's `overlays/` directory.
- Overlay: 5 fragments — enable the (already present, disabled) `&bt` node on uart0, point uart0 pinctrl at the stock `uart0_gpio32` + `uart0_ctsrts_gpio30` groups, disable `&uart1` + `&minibt`, repoint `/aliases/bluetooth` so the firmware injects the BD address into the right node.
- Commented `#dtoverlay=ha-yellow-bt-pl011` with explanation in the Yellow `config.txt` `[cm4]` section. Since the rauc boot hook preserves `*.txt` across updates, users who uncomment it keep the fix permanently.

## Testing

- Applied the equivalent DTB modification directly on my Yellow (CM4, HAOS 18.2): clean HCI init on every boot since (`BCM4345C0 (003.001.025) build 0382`, no `tx timeout`), correct BD address, and stable across multiple soft reboots — which previously wedged the chip reliably. Details in https://github.com/home-assistant/operating-system/issues/4898
- Verified `fdtoverlay -i <stock dtb> -o merged.dtb ha-yellow-bt-pl011.dtbo` produces a result semantically identical to the tested hand-patched DTB (same node states, same pinctrl phandles).
- Overlay compiles clean with `dtc -@`; `__fixups__` reference only labels present in the Yellow DTB's `__symbols__`.
- Draft until I've run a full `make yellow` build to confirm the package integrates; happy to adjust naming/placement to whatever fits the flow-control fix's timeline.

Related to #4898, #4590, #4884

## Summary by CodeRabbit

* **New Features**
  * Added optional Home Assistant Yellow device-tree overlays to Buildroot.
  * Added support for routing CM4 Bluetooth through the PL011 UART with hardware flow control.
  * Enabled overlay packaging in the Yellow device configuration.
* **Documentation**
  * Added guidance on Bluetooth routing, UART-header limitations, and the impact on the J11 UART.
  * Bluetooth routing remains disabled by default unless explicitly enabled.


sairon (commented):
Hey @gvlekke, thanks for the contribution, but I'm going to reject it. The change you suggest is rather invasive and it's not reasonable to pull it in as an interim solution to the problem. The flow-control solution is imminent, I'm finishing final testing and I'm expecting to put it on review later today. It will make these changes immediately obsolete.

However, it will take a while before it gets to the stable channel, so if you or others don't want to run on dev OS version in the meantime, feel free to share the compiled `.dtbo` file and instructions how to apply it in the referred issue. Future update will remove the file from the overlays directory and it will revert to mini UART, so it's worth keeping that in mind, however, the stray `dtoverlay=` pointing to non-existing overlay should do no harm.

Also note that OHF/Home Assistant project have the [AI Policy](https://github.com/home-assistant/operating-system/blob/dev/AI_POLICY.md) in place which defines how LLMs should be used to make contributions and communicate on issues. I'd not say it's in direct violation of it here, but it's worth being aware of it if you decide to make more complex contributions to the project in the future.

gvlekke:
@sairon yeah I can understand was more of an example basicly. If I need to test it here let me now

And I will check the Ai policy. Was not aware.

PR #4968, opened 2026-08-17
Avoid polling for SD card on CM5 Lite on Yellow

### LINKED ISSUE #4961: Yellow + CM5 Lite: mmc0 polls a non-existent SD slot, kworker permanently in D state, load average inflated by ~1.0

On a Home Assistant Yellow with a **CM5 Lite** (no eMMC) booting from NVMe,
a kworker thread is permanently blocked in uninterruptible sleep (D state)
waiting on the mmc0 controller.

Root cause appears to be known upstream: the CM5 has no card detect signal
(see raspberrypi/linux#6744), so the controller cannot tell whether a card is
present and keeps polling. On a Lite module the SDIO lines are routed to the
board-to-board connector instead of an on-module eMMC — and the Yellow has no
microSD slot, so they terminate at nothing. mmc0 is nevertheless enabled in
the device tree (with `no-sd`/`no-sdio` set, which is not sufficient).

The condition appeared after replacing a CM4 (with eMMC) by a CM5 Lite, not
after an OS update. All known mmc0 workarounds for Yellow+CM5 (#3700, #3793)
are already present in this version.

Effects:
- Load average permanently inflated by ~1.0 with no actual CPU work
  (all containers together use ~2.3% CPU, `/proc/pressure/io` is 0.00)
- The reported load is misleading in the System Monitor integration and in
  any automation using it

No hung task warning is emitted: the blocked worker rotates between PIDs,
so no single task exceeds `hung_task_timeout_secs`. The condition is
therefore silent despite being effectively permanent.

yellow (Home Assistant Yellow)

18.2

No

### Hardware details

- Home Assistant Yellow
- Raspberry Pi CM5 Lite (16 GB RAM, no eMMC)
- Boot + data on NVMe (WDC PC SN530 256 GB)
- No microSD slot on the Yellow

1. Yellow with CM5 Lite, HAOS on NVMe
2. `ps -eo state,pid,comm,wchan:32 | awk '$1 ~ /^D/'`

```txt
No relevant entries
```

```txt
`dmesg` shows the controller registering and then nothing at all —
no card, no timeout, no error:


[    0.381558] sdhci: Secure Digital Host Controller Interface driver
[    0.381588] sdhci-pltfm: SDHCI platform and OF driver helper
[    0.636636] mmc0: SDHCI controller on 1000fff000.mmc [1000fff000.mmc] using ADMA 64-bit


`lsblk` lists no mmcblk device; `/sys/class/mmc_host/` contains mmc0.
```

### System information

# ha info
arch: aarch64
channel: stable
docker: 29.6.2
features:
- reboot
- shutdown
- services
- network
- hostname
- timedate
- os_agent
- haos
- resolved
- journal
- disk
- mount
hassos: "18.2"
homeassistant: 2026.7.4
hostname: homeassistant
logging: info
machine: yellow
operating_system: Home Assistant OS 18.2
state: running
supervisor: 2026.07.5
supported: true
supported_arch:
- aarch64

### Additional information

Sampled once per second over 120 seconds, twice — once at 7 days uptime,
once 24 minutes after a cold reboot. Identical results:

| wchan | hits / 120 |
|---|---|
| `mmc_wait_for_req_done` | 111 |
| `mmc_power_up.part.0` | 1 |
| `rpi_firmware_property_list` | 34 |

The blocked PID rotates (6 distinct kworkers in the first run), each holding
the state for 10–20 seconds. The recurring `mmc_power_up.part.0` is
consistent with `mmc_rescan` repeatedly retrying the power-up sequence
rather than failing once.

**Possible fix — upstream device tree parameters**

This board already carries `&sdio1 { no-sdio; no-sd; };` from #3700, but that
only restricts which card types the driver accepts — it does not disable the
node, so `mmc_rescan` still runs and still blocks. That patch targeted boot
delay on modules *with* eMMC; it does not address a Lite module in a carrier
with no SD slot.

raspberrypi/linux#6744 (merged Mar 2025, shipped in kernel 6.12.21) added
`sd` and `sd_poll_once` dt-params to the CM5 device tree, explicitly because
"The CM5 lacks a card detect signal, so it can be useful to be able to
disable the external SD card slot (or onboard EMMC on a non-Lite board)."

Since the Yellow has no microSD slot, `sd=off` — or `status = "disabled"` on
the node — would seem correct here, at least when a Lite module is detected.
I have not verified whether those params are reachable on the Yellow, which
carries its own DTS
(`0016-ARM-dts-bcm2712-Add-device-tree-for-CM5-on-HA-Yellow.patch`,
plus #3688 syncing it against the plain CM5 DTS) rather than the stock one.

Related prior art: raspberrypi/linux#6647 reports CM5 modules *without WiFi*
hanging on reboot due to the unused sdio2 interface; disabling that node via
overlay fixed it reliably in all tests. Different interface, same pattern —
an unused MMC block left enabled on a Lite module.

The review discussion on #3700 also notes that every BCM2712 SDHCI instance
unconditionally writes `SDIO_CFG_SD_PIN_SEL = SDIO_CFG_SD_PIN_SEL_CARD_DETECT`,
which may be relevant given the CM5 has no such signal.

sairon:
On CM5 Lite, kernel keeps polling the mmc0 interface indefinitely as it doesn't indicate the non-removable property while it has no SD card slot. Patch the BCM2712 Yellow device tree to poll only once.

Fixes #4961

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved boot behavior on the CM5 Yellow board by preventing repeated scanning of the onboard eMMC.
  - The onboard storage is now recognized as permanently connected, reducing unnecessary polling during operation.

### DIFF
--- a/buildroot-external/board/raspberrypi/yellow/patches/linux/0006-ARM-dts-bcm2712-Scan-sdio1-only-once-at-boot.patch
+++ b/buildroot-external/board/raspberrypi/yellow/patches/linux/0006-ARM-dts-bcm2712-Scan-sdio1-only-once-at-boot.patch
@@ -1,0 +1,33 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Jan=20=C4=8Cerm=C3=A1k?= <sairon@sairon.cz>
+Date: Mon, 17 Aug 2026 14:48:21 +0200
+Subject: [PATCH] ARM: dts: bcm2712: Scan sdio1 only once at boot
+MIME-Version: 1.0
+Content-Type: text/plain; charset=UTF-8
+Content-Transfer-Encoding: 8bit
+
+Yellow has no SD card slot and without the non-removable property it
+keeps polling the onboard eMMC on CM5 Lite forever. Fix it by indicating
+it's non-removable, polling the interface only once at boot.
+
+Signed-off-by: Jan Čermák <sairon@sairon.cz>
+---
+ arch/arm64/boot/dts/broadcom/bcm2712-rpi-cm5-ha-yellow.dts | 5 +++++
+ 1 file changed, 5 insertions(+)
+
+diff --git a/arch/arm64/boot/dts/broadcom/bcm2712-rpi-cm5-ha-yellow.dts b/arch/arm64/boot/dts/broadcom/bcm2712-rpi-cm5-ha-yellow.dts
+index 0dbb19898636..b85fa4c0b777 100644
+--- a/arch/arm64/boot/dts/broadcom/bcm2712-rpi-cm5-ha-yellow.dts
++++ b/arch/arm64/boot/dts/broadcom/bcm2712-rpi-cm5-ha-yellow.dts
+@@ -97,6 +97,11 @@ &uart10 {
+ 	skip-init;
+ };
+ 
++/* Scan the eMMC only once at boot, Yellow has no microSD slot. */
++&sdio1 {
++	non-removable;
++};
++
+ /* RP1 USB ports are not connected on Yellow. */
+ &rp1_usb0 {
+ 	status = "disabled";


agners (approved):
Nice fix, LGTM!

PR #4809, opened 2026-06-22
Enable iwlmld driver for Wi-Fi 7 Intel devices

agners:
## Summary

Newer Intel Wi-Fi 7 devices (BZ/GL/SC/DR families, e.g. the **BE200**) regressed on HAOS 18.0 after the kernel bump from 6.12 → 6.18. These devices are no longer driven by the `iwlmvm` op-mode — their firmware now requires the new `iwlmld` driver (added upstream in 6.13+).

HAOS never enabled `CONFIG_IWLMLD`, so `iwlwifi` aborts the probe:

```
Bluetooth: <...>
iwlwifi 0000:02:00.0: IWLMLD needs to be compiled to support this firmware
iwlwifi 0000:02:00.0: probe with driver iwlwifi failed with error -22
```

This matches the `iwl-drv.c` path that runs when `CONFIG_IWLMLD` is disabled and a Wi-Fi 7 device with new-enough firmware is detected.

## Fix

Enable `CONFIG_IWLMLD=m` in `device-support-wireless-pci.config` (shared by `generic_x86_64`, `generic_aarch64`, and `ova`). The matching firmware (`iwlwifi-gl-*`) already ships via linux-firmware, so no firmware change is needed.

## Testing

Verified against the already-built 6.18 kernel tree:
- BE200 is the `iwl_gl_mac_cfg` device (`cfg/bz.c`); requires the `mld` op-mode.
- Pre-fix `.config` had `# CONFIG_IWLMLD is not set` and `mld/iwlmld.ko` was not built.
- `iwlwifi-gl-c0-fm-c0-*.ucode` / `.pnvm` already present in the image.

Fixes #4784

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Enabled support for Intel wireless network adapters on PCI/PCIe connections.

This update improves hardware compatibility for users with compatible Intel wireless network devices.

PR #4973, opened 2026-08-20
Use rootfs compression hints also for OVA, LZMA-compress large binaries

sairon:
Some recent features (e.g. #4809, #4816, #4837) increased the rootfs bloat between 18.0 and 18.1, bringing the rootfs size on OVA and generic-x86-64 just 0.5 MiB below the 256 MiB cap, blocking us from adding new features that require space in the rootfs. To gain some space, also start compressing large daemon binaries using the LZMA algorithm. The selected binaries are normally invoked only once per boot, so the runtime cost of the more expensive decompression is minimal.

The same compression hints are now used for OVA, where it brings us around 45 MiB in the rootfs partition image

## Summary by CodeRabbit

* **Performance**
  * Improved filesystem image compression for PC and OVA builds using optimized LZ4HC and LZMA settings.
  * Applied stronger compression to firmware, kernel modules, and selected long-running container services.
  * Increased compression block sizes for applicable workloads to improve storage efficiency.


agners (approved):
LGTM!

PR #4976, opened 2026-08-24
Re-add rootfstype to kernel command line

sairon:
The rootfstype parameter was removed in HAOS 13.0 when the root filesystem switched from SquashFS to EROFS in #3456, because during the transition one slot could still contain a SquashFS rootfs. The enforced upgrade path now only allows upgrading from 15.2 or newer, so after an update both slots are guaranteed to contain an EROFS rootfs and the parameter can be added back. Specifying rootfstype avoids probing other filesystem types when mounting the rootfs.

For U-Boot boards add rootfstype=erofs to the boot scripts and to the grub.cfg on UEFI targets. On Raspberry Pi 5 the cmdline.txt on the boot partition persists across updates, so adjust it in a targeted migration in the RAUC hook. The migration replaces the obsolete removal of rootfstype=squashfs, which is already handled on all systems the update can be installed from.

## Summary by CodeRabbit

* **New Features**
  * Boot configurations now explicitly identify the root filesystem as EROFS across supported devices and boot slots.
  * Raspberry Pi 5 installations add the EROFS filesystem setting when it is not already present.

* **Bug Fixes**
  * Improved reliability when mounting the read-only system partition during normal and rescue boots.
  * Preserved existing filesystem configuration instead of removing EROFS-related settings during updates.

### DIFF
--- a/buildroot-external/board/arm-uefi/generic-aarch64/grub.cfg
+++ b/buildroot-external/board/arm-uefi/generic-aarch64/grub.cfg
@@ -69,17 +69,17 @@ file_env -f ($root)/cmdline.txt cmdline
 regexp --set 1:boothd (.+),.+ ${root}
 
 menuentry "Slot A (OK=$A_OK TRY=$A_TRY)" {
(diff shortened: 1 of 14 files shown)


agners (approved):
LGTM!

PR #4982, opened 2026-08-25
Add service to reset Docker storage when requested

sairon:
Corrupted Docker image layers cannot be always fixed by removing and re-pulling a single image, because layers are shared between images. The only (easy) way to recover is wiping all of Docker's storage which is currently cumbersome and requires OS shell access.

Add service triggered by /mnt/data/docker/.wipe-scheduled flag file which wipes the Docker directory by atomically renaming it and deleting synchronously. If this is interrupted, docker-prepare script removes the leftovers before Docker is started on every boot.

It should be noted that docker-prepare also forces the switch to containerd snapshotter after the wipe.

Refs home-assistant/supervisor#6555

## Summary by CodeRabbit

- **New Features**
  - Added automatic Docker storage reset when a reset is scheduled.
  - Safely moves existing Docker storage aside to support recovery from interrupted resets.
  - Cleans up temporary storage from completed or interrupted resets.

- **Bug Fixes**
  - Prevents Docker storage reset while the Docker data directory is mounted.
  - Limits startup cleanup to Docker reset leftovers, avoiding removal of unrelated files.

### DIFF
--- a/buildroot-external/rootfs-overlay/usr/libexec/haos-docker-reset
+++ b/buildroot-external/rootfs-overlay/usr/libexec/haos-docker-reset
@@ -14,8 +14,6 @@ if [ -e "${DOCKER_DIR}" ]; then
     mv "${DOCKER_DIR}" "${WIPED_DIR}"
 fi
 
-mkdir -m 0710 "${DOCKER_DIR}"
-
 # Leftovers are removed by docker-prepare if this is interrupted
 echo "[INFO] Removing old Docker storage"
 rm -rf "${DOCKER_DIR}".wipe.*
--- a/buildroot-external/rootfs-overlay/usr/libexec/haos-docker-reset
+++ b/buildroot-external/rootfs-overlay/usr/libexec/haos-docker-reset
@@ -18,4 +18,4 @@ mkdir -m 0710 "${DOCKER_DIR}"
 
 # Leftovers are removed by docker-prepare if this is interrupted
 echo "[INFO] Removing old Docker storage"
-rm -rf "${DOCKER_DIR}".wiped*
+rm -rf "${DOCKER_DIR}".wipe.*
--- a/buildroot-external/rootfs-overlay/usr/libexec/haos-docker-reset
+++ b/buildroot-external/rootfs-overlay/usr/libexec/haos-docker-reset
@@ -1,0 +1,21 @@
+#!/bin/sh
+set -e
+
+DOCKER_DIR="/mnt/data/docker"
+
+if findmnt /var/lib/docker > /dev/null; then
+    echo "[ERROR] Unable to reset Docker storage while /var/lib/docker is mounted"
+    exit 1
+fi
+
+if [ -e "${DOCKER_DIR}" ]; then
+    WIPED_DIR="${DOCKER_DIR}.wipe.$(date +%s)"
+    echo "[INFO] Moving Docker storage to ${WIPED_DIR}"
+    mv "${DOCKER_DIR}" "${WIPED_DIR}"
+fi
+
+mkdir -m 0710 "${DOCKER_DIR}"
+
+# Leftovers are removed by docker-prepare if this is interrupted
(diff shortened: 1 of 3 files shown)


buildroot-external/rootfs-overlay/usr/libexec/haos-docker-reset
```diff
@@ -0,0 +1,21 @@
...
+    WIPED_DIR="${DOCKER_DIR}.wipe.$(date +%s)"
+    echo "[INFO] Moving Docker storage to ${WIPED_DIR}"
+    mv "${DOCKER_DIR}" "${WIPED_DIR}"
+fi
+
+mkdir -m 0710 "${DOCKER_DIR}"
+
+# Leftovers are removed by docker-prepare if this is interrupted
+echo "[INFO] Removing old Docker storage"
+rm -rf "${DOCKER_DIR}".wiped*
```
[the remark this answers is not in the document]
agners: `.wiped` doesn't match, so this seems legitimate. 

Also let's use the same suffix here and in `/usr/libexec/docker-prepare`.

agners (changes requested):

buildroot-external/rootfs-overlay/usr/libexec/haos-docker-reset
```diff
@@ -0,0 +1,21 @@
...
+    exit 1
+fi
+
+if [ -e "${DOCKER_DIR}" ]; then
+    WIPED_DIR="${DOCKER_DIR}.wipe.$(date +%s)"
+    echo "[INFO] Moving Docker storage to ${WIPED_DIR}"
+    mv "${DOCKER_DIR}" "${WIPED_DIR}"
+fi
+
+mkdir -m 0710 "${DOCKER_DIR}"
```
agners: If you want to be extra safe, this mkdir would need to go into docker-prepare as well.

Or alternatively, something along these lines:

```
mkdir -m 0710 "${WIPED_DIR}"

# Atomic exchange:
mv --exchange "${DOCKER_DIR}" "${WIPED_DIR}"
```

😎

agners: Hm, `mv --exchange` might not be available in busybox `mv` 😢 . I mean this is being pedantic. Does get the dir created automatically anyways? How do we create the dir after wipe? 🤔

buildroot-external/rootfs-overlay/usr/libexec/haos-docker-reset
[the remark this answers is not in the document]
sairon: > `.wiped` doesn't match, so this seems legitimate.
> 
> Also let's use the same suffix here and in `/usr/libexec/docker-prepare`.

D'oh, last minute changes... :facepalm: It still worked but not in the planned way. Fixed.

buildroot-external/rootfs-overlay/usr/libexec/haos-docker-reset
sairon: > Hm, `mv --exchange` might not be available in busybox `mv` 😢 .

Yep:
```
mv: unrecognized option '--exchange'
BusyBox v1.37.0 (2026-08-24 11:46:47 UTC) multi-call binary.

Usage: mv [-finT] SOURCE DEST
or: mv [-fin] SOURCE... { -t DIRECTORY | DIRECTORY }

Rename SOURCE to DEST, or move SOURCEs to DIRECTORY

        -f      Don't prompt before overwriting
        -i      Interactive, prompt before overwrite
        -n      Don't overwrite an existing file
        -T      Refuse to move if DEST is a directory
        -t DIR  Move all SOURCEs into DIR
```

> Does get the dir created automatically anyways? How do we create the dir after wipe? 🤔

Actually, it seems to be created by systemd and permissions are fixed by Docker itself (as explained in the newly pushed commit). So let's get rid of that.

agners (approved):
LGTM!

PR #4984, opened 2026-08-27
Enable autofs support explicitly

agners:
Supervisor is moving network mounts to systemd `.automount` units (home-assistant/supervisor#7167), which makes the kernel automounter a hard requirement for network storage.

Every board already has it, but only as a side effect: Buildroot's systemd package force-enables `CONFIG_AUTOFS_FS` through `SYSTEMD_LINUX_CONFIG_FIXUPS`. Those fixups run after our board configs and fragments, so the option is even set on images where `board/raspberrypi/kernel.config` explicitly disabled it (that line was there until 11.2 and never had an effect — confirmed on a HAOS 11.1 rpi3-64, whose `/proc/config.gz` has `CONFIG_AUTOFS_FS=y`).

Worth stating explicitly anyway, for two reasons:

- `KCONFIG_ENABLE_OPT` skips symbols already set to `=m`, so the Rockchip boards (Green, ODROID-M1/M1S) build autofs as a module while every other board has it built in. Setting it in the shared fragment makes this uniform, so arming an automount never depends on on-demand module loading.
- Nothing in our configs records that network storage depends on autofs. A kernel bump or a change to Buildroot's systemd fixups could take it away without any signal on our side.

No functional change for the boards that already build it in.

## Summary by CodeRabbit

* **New Features**
  * Enabled automatic filesystem mounting support in the system kernel.


agners:
I've double checked my local builds, and a couple of real-world boards (notably RPi, ova, x86-64, back to HAOS 15.2), and indeed AUTOFS seems to be enabled everywhere.

sairon (approved):
Sounds reasonable, LGTM!

Issue #4986, opened 2026-08-28
Add OS >=18.3 and <19.0 to upgrade path in version file


sairon:
This is a tracking issue for changes added before 19.0 release that perform migrations that we want to enforce and clean up later. Before 19.0 release, latest 18.x should be added to the upgrade path mapping in the version file.
Already implemented:
- #4976
https://github.com/home-assistant/operating-system/blob/d846fdb13b3e12fa413ec88349ef544f8bc99d95/buildroot-external/ota/rauc-hook#L33-L36

PR #4988, opened 2026-08-28
Persist timesyncd config in /etc/systemd/timesyncd.conf.d

sairon:
The persistent timesyncd configuration is a single bind-mounted file over `/etc/systemd/timesyncd.conf`. The NTP configuration through OS Agent (home-assistant/os-agent#207) attempted to do atomic updates, but a temporary file cannot be created next to it in the read-only `/etc/systemd`, and replacing the file in the overlay directly changes the inode, so the bind mount keeps pointing to the old content. Also, a persisted copy of the whole file means changes to the shipped defaults never reach existing installations.

Bind-mount the `/etc/systemd/timesyncd.conf.d` directory from the overlay instead and keep the shipped `timesyncd.conf` read-only. Configuration is layered as drop-ins, from lowest priority to higher:

* `10-ntp.conf` (in `/run`): NTP servers from DHCP
* `20-custom.conf`: `timesyncd.conf` imported from the CONFIG partition
* `50-os-agent.conf`: NTP servers set through the OS Agent D-Bus API

On upgrade, settings from the previously persisted `timesyncd.conf` other than the shipped defaults are moved to `20-custom.conf`. This migration can be removed later (with #4986). `50-os-agent.conf` should be only managed by OS Agent, so we don't need to care about user's edits. Finally, if needed, drop-ins with higher priorities can be added for complex customizations.

Refs home-assistant/supervisor#6278

## Summary by CodeRabbit

- **New Features**
  - Time synchronization settings are now managed through drop-in configuration files, allowing custom settings to coexist with system defaults.
  - Imported time synchronization configuration is stored separately from the main system configuration.

- **Bug Fixes**
  - Existing time synchronization settings are automatically migrated to the new format, helping preserve custom configuration during upgrades.
  - Default system settings are no longer copied over or overwritten unnecessarily.

### DIFF
--- a/buildroot-external/rootfs-overlay/usr/lib/systemd/system/etc-systemd-timesyncd.conf.d.mount
+++ b/buildroot-external/rootfs-overlay/usr/lib/systemd/system/etc-systemd-timesyncd.conf.d.mount
@@ -5,8 +5,8 @@ After=mnt-overlay.mount haos-overlay.service
 Before=systemd-timesyncd.service
 
 [Mount]
-What=/mnt/overlay/etc/systemd/timesyncd.conf
-Where=/etc/systemd/timesyncd.conf
+What=/mnt/overlay/etc/systemd/timesyncd.conf.d
+Where=/etc/systemd/timesyncd.conf.d
 Type=none
 Options=bind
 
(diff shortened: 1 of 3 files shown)


agners (approved):
LGTM

Issue #4913, opened 2026-07-22
RPi4: overlay2 layer extraction writes 0-byte regular files (exec format error), persists after re-pull; repeated extraction hung the host


taroboys:
### Describe the issue you are experiencing

Home Assistant Core update from `2026.6.4` to `2026.7.3` (also reproduced with `2026.7.2`) fails to start and Supervisor auto-rolls back. Root cause traced to **local Docker `overlay2` layer corruption on write/extract**, similar to #6835, but persistent even after removing all affected images and a clean reboot — unlike #6835 where removing the stale shared layer + re-pull fixed it.

### What version of Home Assistant Operating System has the issue?

18.1

### What platform are you using?

Raspberry Pi 4 (rpi4-64, aarch64)

### Steps to reproduce the issue

1. `ha core update` to `2026.7.3` on `2026.6.4` → new container starts, Core crashes within ~5s, Supervisor auto-rollback to `2026.6.4` (log: `HomeAssistant update failed -> rollback!`)
2. Inspected via a separately-installed "Advanced SSH & Web Terminal" add-on with Docker API access (`docker exec`/`docker run` into containers directly, since the update crash happens before Core's own log file is written)
3. `docker run --rm --entrypoint sh ghcr.io/home-assistant/raspberrypi4-64-homeassistant:2026.7.3 -c "echo ok"` → `exec /bin/sh: exec format error`
4. Confirmed this is **not architecture-related**: `docker inspect` reports `Architecture: arm64, Os: linux` correctly (matches host `aarch64`)
5. Extracted files directly without executing them (`docker create` + `docker cp`, no exec involved):
   - Regular files (e.g. `/bin/busybox`, `/etc/passwd`, `/etc/hostname`) come out **0 bytes**
   - Symlinks (e.g. `/bin/sh -> /bin/busybox`) extract correctly
6. Reproduced identically with a totally unrelated public image, `alpine:latest` — same 0-byte regular-file pattern, ruling out a Home Assistant-specific image defect
7. Verified this is not general SD card/storage corruption: 30x direct `dd` writes of 20MB random data to the same partition (`/data`, same physical disk as Docker's storage root `/mnt/data/docker`), synced and re-read, all matched — no bit-level corruption on plain writes
[...]

PR #4994, opened 2026-09-01
Force filesystem sync when unpacking container image layers

agners:
## Summary

Enable the containerd diff-service `sync_fs` option so the daemon issues one `syncfs(2)` after each unpacked image layer. containerd fsyncs its metadata database when committing a snapshot, but never syncs the unpacked layer data itself — after an unclean shutdown shortly after an image pull this leaves layers with zero-byte files (correct names/modes/mtimes, no data) while the snapshot stays durably recorded as valid. Because committed snapshots are never re-verified or re-unpacked, the corruption even survives deleting and re-pulling the image (moby/moby#49784, containerd/containerd#8973), and with shared base layers it is only recoverable by wiping the Docker storage.

This is the failure signature behind a growing number of corruption reports, e.g. the 0-byte coredns in home-assistant/plugin-dns#207, home-assistant/supervisor#6476, home-assistant/supervisor#6835, #4913. Upstream background: containerd/containerd#8895, fixed opt-in via containerd/containerd#10284 (containerd >= 2.0). The daemon-side option applies to all clients of the daemon, including dockerd with the containerd image store (the default since HAOS 17). Installs still on the overlay2 graphdriver unpack inside dockerd and are unaffected either way.

## Verification (QEMU, HAOS 18.3.dev, Docker 29.6.2 / containerd 2.2.6)

- **Stock settings, power cut <1s after `docker pull` of the DNS plugin image**: corrupts 2/2 runs — one run zeroed `/usr/bin/coredns` entirely, the other zeroed the s6 scripts and left coredns truncated (20021248 of 25919650 bytes), matching the two damage profiles reported in plugin-dns#207.
- **Deterministic A/B with kernel background writeback suppressed** (survival can then only come from the stack syncing explicitly): baseline ends with 1215 zero-byte files and `exec format error`; with `sync_fs` the image survives the identical power cut intact (152 kB dirty at the cut vs 76 MB).
- **strace on containerd**: 0 `syncfs` calls per image load before, exactly one per layer after — also confirming dockerd pulls route through the daemon's diff service where this option takes effect.
- **Cost** (load+unpack of the 2 GB Core image): 41.6s → 42.2s on a fast disk; 226s → 230s (+1.9%) on a 30 MB/s / 250 IOPS throttled disk. QEMU throttling does not model SD-card sync latency, so the slowest cards may pay somewhat more, but the cost is bounded at one `syncfs` per layer rather than per file.

Given HAOS commonly runs on SD cards without a UPS, trading a slightly slower pull for image data that is actually on disk when the pull completes seems like the right default.

Related: home-assistant/supervisor#6555 (Docker storage reset, the recovery path for already-corrupted installs).

## Summary by CodeRabbit

* **Reliability**
  * Improved data durability when applying container image layers.
  * Reduced the risk of losing unpacked image data after unexpected power loss.

### DIFF
--- a/buildroot-external/rootfs-overlay/etc/containerd/config.toml
+++ b/buildroot-external/rootfs-overlay/etc/containerd/config.toml
@@ -7,3 +7,12 @@ disabled_plugins = [
   "io.containerd.tracing.processor.v1.otlp",
   "io.containerd.internal.v1.tracing"
 ]
+
+# containerd only fsyncs its metadata database, not unpacked layer content, so
+# a power loss shortly after an image pull leaves layers with zero-byte files
+# while the snapshot is durably recorded as valid. Such corruption survives
+# delete/re-pull since existing snapshots are never re-verified. Force a
+# syncfs(2) after each applied layer to make unpacked data as durable as the
+# metadata that references it.
+[plugins."io.containerd.service.v1.diff-service"]
+  sync_fs = true


agners:
docker load times (image tar on the data partition, per-cycle wiped store)

Small image — DNS plugin, 5 layers, ~65 MB (isolates fixed per-layer sync cost):

```
  ┌─────────────────┬───────────────┬───────────────┬───────────────┐
  │    Platform     │   baseline    │    sync_fs    │     delta     │
  ├─────────────────┼───────────────┼───────────────┼───────────────┤
  │ RPi4 (SD)       │ 5.29 / 5.26 s │ 7.83 / 8.55 s │ +2.9 s (+55%) │
  ├─────────────────┼───────────────┼───────────────┼───────────────┤
  │ RPi3 (SD)       │ 11.30 s       │ 15.06 s       │ +3.8 s (+33%) │
  ├─────────────────┼───────────────┼───────────────┼───────────────┤
  │ OVA (fast disk) │ 1.04 / 0.95 s │ 1.26 / 1.14 s │ +0.2 s (+20%) │
  └─────────────────┴───────────────┴───────────────┴───────────────┘
```

Big image — Core, ~10 layers, ~2 GB (throughput-dominated):

```
  ┌─────────────────┬──────────────────────────────┬──────────────────────────────┬──────────────────────────────┐
  │    Platform     │           baseline           │           sync_fs            │            delta             │
  ├─────────────────┼──────────────────────────────┼──────────────────────────────┼──────────────────────────────┤
  │ RPi4 (SD)       │ 196.0 / 209.4 s (mean 202.7) │ 208.0 / 232.8 s (mean 220.4) │ +17.7 s (+8.7%)              │
  ├─────────────────┼──────────────────────────────┼──────────────────────────────┼──────────────────────────────┤
  │ RPi3 (SD)       │ 417.9 s                      │ 423.6 s                      │ +5.6 s (+1.3%)               │
  ├─────────────────┼──────────────────────────────┼──────────────────────────────┼──────────────────────────────┤
  │ OVA (fast disk) │ 30.1 / 30.6 s                │ 31.7 / 37.0 s                │ +4.0 s (+13%, high variance) │
  └─────────────────┴──────────────────────────────┴──────────────────────────────┴──────────────────────────────┘

```

The measurements appear to have quite some variance. While I've stopped all containers and removed the complete storage, page cache or other circumstances might have played a role.

It is interesting that DNS has this higher increase. It could be just due to its small size. or it could be related to CoreDNS binary which is rather large and in the second to last layer (and rather often corrupts in real world).

In general, the cost appears acceptable to me.

sairon (approved):
Thanks for the tests, it doesn't look that bad, and if it helps with the corruptions we're seeing, the minor performance hit is acceptable for me - still better than nuking the whole Docker storage, which is the only other option we have.

